PR #25, opened 2021-07-05
Add a way to view a user's details

nitnelave:
The new API endpoint "/api/user/<user_id>" returns the user's details, and is accessible by both admins and the user themselves.


nitnelave:
This addresses point number 2 of https://github.com/nitnelave/lldap/issues/11 .

PR #73, opened 2021-10-28
Implement password reset via LDAP

nitnelave:
This adds support for other systems to modify a user's password, for instance Authelia.

In addition, this adds a few missing LDAP functions required for a proper LDAP server, such as rootDSE request handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1644,9 +1644,9 @@ dependencies = [
 
 [[package]]
 name = "ldap3_server"
-version = "0.1.7"
+version = "0.1.9"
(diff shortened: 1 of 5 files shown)


JaneJeon:
Oh, so does it mean that this is now outdated? https://github.com/nitnelave/lldap/blob/main/example_configs/authelia_config.yml#L10

nitnelave:
Yes! Feel free to open a PR to fix that, otherwise I'll do it when I have time.

nitnelave:
Added in #91

PR #78, opened 2021-11-12
(Doc) Keycloak requires a full DN, not just the username

draganczukp:
Something I noticed when setting up Keycloak. Setting just "admin" for the bind DN doesn't work. Instead you have to set the whole DN: `cn=admin,ou=people,dc=example,dc=com`


nitnelave:
You are completely right! I missed that one.

nitnelave:
Thanks for the contribution!

Issue #93, opened 2021-11-27
Serve libraries locally


kaysond:
Right now all of the below libraries are pointing to cdns. For people like me who run a reverse proxy with some security headers, this can get blocked.

https://github.com/nitnelave/lldap/blob/5b5395103ae56ebbea841be76c000b8b243895dc/app/index.html#L8-L24

Specifically, I set `Content-Security-Policy: default-src 'self' 'unsafe-inline'`, which prevents backend services from loading resources from external sites as a security measure.

Would you be open to a PR that adds these libs to the repo and serves them locally? I'm thinking we should add a `app/static` directory and serve all static files from there (we can also then add a `COPY /app/app/static` to the dockerfile so nothing gets forgotten)

nitnelave:
Hmmm... I'd rather avoid having a copy of the libraries in the repository. However, I'm open to a PR that adds a `main_static.html` and sets up the Dockerfile to copy the main_static and download the files locally. Would that work for you?

kaysond:
I think that's smarter actually. We could add `ARG`s for the versions of the libraries.

What would be in `main_static.html`?

nitnelave:
Same as `main.html`, but referencing the static folder. Basically, if you run the project from source, it'll use the dynamic, CDN libraries, and if you run it from docker it'll use the `static/` folder.

JaneJeon:
What’s stopping you from basically webpack/rollup’ing all of the libraries into the bundle?

nitnelave:
I'd like to avoid making the bundle bigger if possible. That way the logic of the spa can start before the fonts and css are loaded, which should give a faster page load.

Also, I don't know what I'm doing with roll-up :)

JaneJeon:
@nitnelave that's the neat thing about rollup or webpack or any of these "bundlers" (I should know this, javascript is my language of choice) - they don't necessarily have to bundle it all into one big file. In fact, the more common thing to do in production (and I believe it is the _default_ for webpack though I am not sure for rollup) is to chunk them (typically, you'll see a whooooole bunch of tiny little JS files in a folder, and you serve that _folder_, not the singular JS file), so that ONLY the needed parts are loaded when they are requested, so that the MINIMUM amount of JS is even served from the backend in the first place.

You need to try it. @kaysond

nitnelave:
Hmm, after reading both of your links, I see the point of chunk splitting, but I don't think it's useful for our project here:
[...]

PR #96, opened 2021-11-28
Serve libraries locally in the docker container

kaysond:
Addresses #93 

Created a `static` directory and the `index_local.html` that gets copied in the container. I put all of the libraries in `static/libraries.txt` and the container automatically downloads all of the files in that list.

I had to manually add/modify the google fonts css file since it includes font imports that also need to be downloaded.

I also cleaned up the tcp_server a little in terms of routing.

Everything works fine, but I am getting some warnings on build:
```
Step 13/26 : RUN cargo build --release -p lldap     && ./app/build.sh 
 ---> Running in ae86bb0e19da
   Compiling lldap_auth v0.2.0 (/app/auth)
   Compiling lldap v0.2.0 (/app/server)
    Finished release [optimized] target(s) in 2m 20s
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
   Compiling lldap_auth v0.2.0 (/app/auth)
   Compiling lldap_app v0.2.0 (/app/app)
    Finished release [optimized] target(s) in 57.40s
[WARN]: :-) origin crate has no README
[INFO]: Installing wasm-bindgen...
[INFO]: Optimizing wasm binaries with `wasm-opt`...
[INFO]: Optional fields missing from Cargo.toml: 'description', 'repository', and 'license'. These are not necessary, but recommended                                                       [INFO]: :-) Done in 2m 07s
[INFO]: :-) Your wasm pkg is ready to publish at /app/app/pkg.

./main.js → ./pkg/bundle.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
bootstrap (imported by pkg/lldap_app.js)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
bootstrap (guessing 'bootstrap')
created ./pkg/bundle.js in 708ms
```
Not sure if this is old or caused by making bootstrap local.


kaysond:
A couple of things occurred to me with this approach:
- source maps don't work because we're not using the cdn
- it in theory introduces an attack vector that didn't exist before because we're not checking hashes (I had to remove them because for some reason they didn't match...)

kaysond:
Also found another issue - the bootstrap icons font is expected to be at `/static/fonts/...` but the way the dockerfile is set up, it ends up at `/static`.

nitnelave:
Do you need help with that or do you want to keep investigating on your side?

kaysond:
I'll keep working on it, just wanted to make sure nothing is way outside your expectation.

I do need your input on the rust error, though. I'm guessing `[WARN]: :-) origin crate has no README` is ok?

kaysond:
alright everything is fixed and tested. this should be good for merging

server/src/infra/tcp_server.rs
```diff
@@ -76,8 +70,16 @@ fn http_config<Backend>(
...
     .service(Files::new("/pkg", "./app/pkg"))
-    // Default to serve index.html for unknown routes, to support routing.
-    .service(web::scope("/").route("/.*", web::get().to(index)));
+    // Serve static files
+    .service(Files::new("/static", "./app/static"))
+    // Serve static fonts
+    .service(Files::new("/static/fonts", "./app/static/fonts"))
+    // Serve the index
+    .service(
+        web::scope("/")
```
nitnelave, line 79: We need to serve every 404 with the index, for routing purposes.
If the user loads the page "http://server.tld/user/bob" we need to serve the index, and the SPA logic will route to the correct page.

app/build.sh
```diff
@@ -24,4 +24,4 @@ then
   exit 1
 fi
 
-$ROLLUP_BIN ./main.js --format iife --file ./pkg/bundle.js
+$ROLLUP_BIN ./main.js --format iife --file ./pkg/bundle.js --globals bootstrap:bootstrap
```
nitnelave, line 27: Thanks for silencing the warning!

nitnelave (changes requested):

server/src/infra/tcp_server.rs
kaysond: I was wondering why that wildcard route was there. I was also wondering why every incorrect `/...` request failed. 🤦🏽‍♂️

nitnelave: Sorry, last nit: can you add back the comment?
// Default to serve index.html for unknown routes, to support routing.

nitnelave (approved):

server/src/infra/tcp_server.rs
kaysond: Done!

nitnelave (approved):

nitnelave:
Thanks a lot! Merged.

PR #100, opened 2021-12-01
jitsi: Add search base

nitnelave:
The config was missing the search base, leading to wrong queries.


bunkerman:
I add it to the example config

PR #101, opened 2021-12-02
Update jitsi_meet.conf

bunkerman:
adjusted the config, that it works with the current lldap release


nitnelave (changes requested):
Can you delete LDAP_SEARCH_BASE at the end too?

nitnelave (approved):

PR #116, opened 2022-02-01
fixed a typo and added verbose for future users to find

abbiewade:
**Issues**

- when getting help debugging on discord, it was mentioned that there was a verbose variable, which I added to the config file so other people can find it. 
- When trying to use a different administrator password, the web UI would use the changed password, but the bindDN password would always be `password`. 

**Fixes**

Updated the config file to have the correct value and added verbose.

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -20,7 +20,7 @@
 ## them to re-login.
 ## You should probably set it through the LLDAP_JWT_SECRET environment
 ## variable from a secret ".env" file.
-## This can also be set from a file's contents by specifying the file path 
+## This can also be set from a file's contents by specifying the file path
 ## in the LLDAP_JWT_SECRET_FILE environment variable
 ## You can generate it with (on linux):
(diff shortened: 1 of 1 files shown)


lldap_config.docker_template.toml
```diff
@@ -95,3 +95,7 @@ key_file = "/data/private_key"
 #from="LLDAP Admin <sender@gmail.com>"
 ## Same for reply-to, optional.
 #reply_to="Do not reply <noreply@localhost>"
+
+## Tune the logging to be more verbose by setting this to be true.
+## You can set it with the VERBOSE environment variable.
```
nitnelave, line 100: Isn't it LLDAP_VERBOSE?

nitnelave (changes requested):

lldap_config.docker_template.toml
abbiewade: yes it is! all fixed now!

nitnelave (approved):

nitnelave:
Thanks for the change!

PR #121, opened 2022-02-20
feat: add simple login

haveachin:


auth/src/lib.rs
```diff
@@ -35,6 +35,27 @@ pub mod login {
...
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ServerLoginResponse {
+        pub token: String,
+        #[serde(skip_serializing_if = "String::is_empty")]
+        #[serde(rename = "refreshToken")]
+        pub refresh_token: String,
+    }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ServerResetResponse {
```
nitnelave, line 54: Server reset? What's that for? Maybe we need a better name.

auth/src/lib.rs
```diff
@@ -66,6 +87,12 @@ pub mod registration {
         pub server_data: String,
         pub registration_upload: opaque::server::registration::RegistrationUpload,
     }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ClientConventionalRegistrationRequest {
```
nitnelave, line 92: I'd like to split this PR into login (that I would more easily approve) and registration/reset, that I would like to keep OPAQUE only if possible. For me, the easy login method is not meant to be opened up to the world, but more like running scripts on the same host more conveniently.

server/src/infra/auth_service.rs
```diff
@@ -48,22 +48,27 @@ fn create_jwt(key: &Hmac<Sha512>, user: String, groups: HashSet<GroupIdAndName>)
...
-            None => Err(HttpResponse::Unauthorized().body("Invalid refresh token")),
-            Some((token, u)) => {
-                let refresh_token_hash = {
-                    let mut s = DefaultHasher::new();
-                    token.hash(&mut s);
-                    s.finish()
-                };
-                Ok((refresh_token_hash, u.to_string()))
-            }
+        None => match request.headers().get("refresh-token") {
```
nitnelave, line 67: Instead of a nested match none, you can use a method on option to combine the 2. I'm not sure exactly which method, but something like `or_else` would probably help.

Of course, you'd need to map them to the same type, some kind of string.
If it gets too complicated, it's fine like this.

auth/src/lib.rs
```diff
@@ -35,6 +35,27 @@ pub mod login {
...
+        pub username: String,
+        pub password: String,
+    }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ServerLoginResponse {
+        pub token: String,
+        #[serde(skip_serializing_if = "String::is_empty")]
+        #[serde(rename = "refreshToken")]
+        pub refresh_token: String,
```
nitnelave, line 50: Maybe `Option<String>`?

auth/src/lib.rs
```diff
@@ -35,6 +35,27 @@ pub mod login {
         pub server_data: String,
         pub credential_finalization: opaque::client::login::CredentialFinalization,
     }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ClientConventionalLoginRequest {
```
nitnelave, line 40: Which convention? :)

Maybe better to rename to simple http login

server/src/infra/auth_service.rs
```diff
@@ -314,6 +330,55 @@ where
...
 }
 
+async fn conventional_login<Backend>(
+    data: web::Data<AppState<Backend>>,
+    request: web::Json<login::ClientConventionalLoginRequest>,
+) -> HttpResponse
+where
+    Backend: TcpBackendHandler + BackendHandler + OpaqueHandler + 'static,
+{
+    let password = request.password.clone();
```
nitnelave, line 340: I don't think you need this clone. You can write `let password = & request.password;`

server/src/infra/auth_service.rs
```diff
@@ -314,6 +330,55 @@ where
     get_login_successful_response(&data, &name).await
 }
 
+async fn conventional_login<Backend>(
```
nitnelave, line 333: Are there no refactoring opportunities for the 2 places that do the entire protocol on the server side?
Maybe returning a result with a custom enum as error for the various cases, cleaning up the code and separating the logic and the error handling.

server/src/infra/auth_service.rs
```diff
@@ -390,6 +455,92 @@ where
     HttpResponse::Ok().finish()
 }
 
+async fn conventional_register<Backend>(
```
nitnelave, line 458: As mentioned above, let's split that into another PR

server/src/infra/auth_service.rs
```diff
@@ -508,6 +659,7 @@ where
             web::resource("/opaque/login/finish")
                 .route(web::post().to(opaque_login_finish::<Backend>)),
         )
+        .service(web::resource("/login").route(web::post().to(conventional_login::<Backend>)))
```
nitnelave, line 662: Maybe `/login/simple`?

nitnelave (changes requested):
Thanks for the PR!
Overall, this looks good, it has all the required bits and bobs and probably works.
It also hopelessly breaks the client since it cannot parse the login response, so I can't merge it until the client changes are there too :)

auth/src/lib.rs
```diff
@@ -35,6 +35,27 @@ pub mod login {
         pub server_data: String,
         pub credential_finalization: opaque::client::login::CredentialFinalization,
     }
+
+    #[derive(Serialize, Deserialize, Clone)]
```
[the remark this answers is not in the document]
haveachin: Not sure what you mean by that. I don't see any message structs with masking secrets. Do you mean the triple-slash comments?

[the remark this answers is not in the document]
nitnelave: Ah, sorry, it's only in the dependencies. See https://github.com/kanidm/ldap3_proto/pull/6/files for example.

auth/src/lib.rs
haveachin: So you would basically require any service that is automatically registering users to use OPAQUE?

server/src/infra/auth_service.rs
haveachin: Sorry, I tried it for like 2 hours and I could figure it out. I'm sure that there is a way, but I'm most certain that it will be more complex that what I have now.

server/src/infra/auth_service.rs
haveachin: I got no clue how Rust works, but I remember something about error handling in one Rust talk I watched. I will try something, but no guarantee that is any good.

server/src/infra/auth_service.rs
haveachin: I changed the routes to `/simple/login` and `/simple/register`

server/src/infra/auth_service.rs
haveachin: fixed

auth/src/lib.rs
haveachin: Renamed it to password_reset::ServerPasswordResetResponse

auth/src/lib.rs
haveachin: Changed it

auth/src/lib.rs
haveachin: Changed it to ClientSimpleLoginRequest

server/src/infra/auth_service.rs
haveachin: That really sucks, because I actually need to register users via API without the opaque protocol

server/src/infra/auth_service.rs
haveachin: I got something working, but I'm not sure if it is better than what I had before

haveachin: nvm, the str pointer were already empty when I was trying to read them

app/src/components/login.rs
```diff
@@ -85,18 +85,16 @@ impl CommonComponent<LoginForm> for LoginForm {
...
                     credential_finalization: login_finish.message,
                 };
-                self.common.call_backend(
-                    HostService::login_finish,
-                    req,
-                    Msg::AuthenticationFinishResponse,
-                )?;
+                self.common
+                    .call_backend(HostService::login_finish, req, move |r| {
+                        Msg::AuthenticationFinishResponse(r)
```
nitnelave, line 90: I think you don't need the lambda here, you can leave it as it was (a reference to the enum constructor)

app/src/components/login.rs
```diff
@@ -85,18 +85,16 @@ impl CommonComponent<LoginForm> for LoginForm {
...
+                    })?;
                 Ok(false)
             }
-            Msg::AuthenticationFinishResponse(user_info) => {
+            Msg::AuthenticationFinishResponse(res) => {
+                let (r, b) = res.context("Could not log in")?;
```
nitnelave, line 95: Can you use longer, more explicit names than r, b? I'm reviewing the PR and have already lost context on what they represent

app/src/infra/api.rs
```diff
@@ -186,9 +186,15 @@ impl HostService {
...
         let parse_token = move |data: String| {
-            get_claims_from_jwt(&data)
+            serde_json::from_str::<login::ServerLoginResponse>(&data)
                 .context("Could not parse response")
-                .and_then(set_cookies)
+                .and_then(|r| {
+                    let success = get_claims_from_jwt(r.token.as_str())
+                        .context("Could not parse response")
+                        .and_then(set_cookies)
+                        .is_ok();
```
nitnelave, line 195: Hmm, sorry I'm reviewing on my phone so it's a bit hard for me to tell exactly what's going on here, but:
- you get a Result
- you add more info to the error with context()
- if it was no error, you set the cookies
- and then you *throw away the error* and just return whether it was successful (is_ok)

I'll have a closer look later, why this function returns a bool, but I'd like to avoid throwing away errors.

server/src/infra/auth_service.rs
```diff
@@ -1,14 +1,8 @@
...
-        error::DomainError,
-        handler::{BackendHandler, BindRequest, GroupIdAndName, LoginHandler},
-        opaque_handler::OpaqueHandler,
-    },
-    infra::{
-        tcp_backend_handler::*,
-        tcp_server::{error_to_http_response, AppState},
-    },
-};
+use std::collections::{hash_map::DefaultHasher, HashSet};
```
nitnelave, line 1: Is that cargo fmt reordering the imports? Maybe we should fuse the various std imports. Small nit, feel free to ignore.

server/src/infra/auth_service.rs
```diff
@@ -314,6 +333,55 @@ where
...
+) -> HttpResponse
+where
+    Backend: TcpBackendHandler + BackendHandler + OpaqueHandler + 'static,
+{
+    let password = &request.password;
+    let mut rng = rand::rngs::OsRng;
+    let opaque::client::login::ClientLoginStartResult { state, message } =
+        match opaque::client::login::start_login(&password, &mut rng) {
+            Ok(n) => n,
+            Err(_) => return HttpResponse::InternalServerError().body(String::new()),
```
nitnelave, line 348: Empty error string?

auth/src/lib.rs
nitnelave: Let's move this conversation to the next PR, since I'd like to anyway split it into one for login and one for registration.

server/src/infra/auth_service.rs
nitnelave: I'm not saying no, I'm saying that we'll talk about it in the next PR :) I can probably easily be convinced.

nitnelave (changes requested):
Alright, that looks much better! Still a couple of nits, and I'll have a last check once I'm on my computer, but it looks good now.

server/src/infra/auth_service.rs
haveachin: Not sure, I just formatted the file using rustfmt

app/src/components/login.rs
haveachin: I refactored it to a lambda

app/src/components/login.rs
haveachin: They should be something like response and success (?), but i'm not 100% sure what the bool is

app/src/infra/api.rs
haveachin: Yeah, the problem is that it expects a result with a Result<(login::ServerLoginResponse, bool) and I'm not sure how I should change. I'm actually happy that I got it working. I find it really hard to read Rust code in general, sorry :(

server/src/infra/auth_service.rs
haveachin: It returns now "Internal Server Error: " appended with the error

server/src/infra/auth_service.rs
haveachin: OK, I moved it to a separate branch

app/src/infra/api.rs
nitnelave: Okay, I checked out your branch to have a better look, and I see the issue. The String returned here is the username, not the token. You were right in parsing the server login response and _then_ calling `get_claims_from_jwt`, but you can just stop after `and_then(set_cookies)` like the original code. The callback still takes a String

haveachin: Oof, I overcomplicated it. Should be fixed now

server/src/infra/auth_service.rs
nitnelave: How about:
```rust
    match (
      request.cookie("refresh_token"),
      request.headers().get("refresh-token")
    ) {
      (Some(c), _) => parse_refresh_token(c.value()),
      (_, Some(t)) => parse_refresh_token(t.to_str().unwrap()),
      (None, None) => Err(HttpResponse::Unauthorized().body("Missing refresh token")),
```

I think it captures well the logic.

nitnelave (changes requested):

nitnelave:
Thanks for your efforts, we're almost there!

server/src/infra/auth_service.rs
```diff
@@ -314,6 +333,58 @@ where
...
+    data: web::Data<AppState<Backend>>,
+    request: web::Json<login::ClientSimpleLoginRequest>,
+) -> HttpResponse
+where
+    Backend: TcpBackendHandler + BackendHandler + OpaqueHandler + 'static,
+{
+    let password = &request.password;
+    let mut rng = rand::rngs::OsRng;
+    let opaque::client::login::ClientLoginStartResult { state, message } =
+        match opaque::client::login::start_login(&password, &mut rng) {
```
nitnelave, line 346: `password`, not `&password`

auth/src/lib.rs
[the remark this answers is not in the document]
nitnelave: I didn't mean to _derive_ Debug, but to implement it by hiding the password. Nevermind, I'll do it myself after this PR.

nitnelave (approved):

nitnelave:
One last thing before I merge: can you squash your commits? Or even better organize them into logical independent units that each pass the tests, but I think for this PR it's enough to squash.

auth/src/lib.rs
```diff
@@ -35,6 +36,27 @@ pub mod login {
...
+    pub struct ClientSimpleLoginRequest {
+        pub username: String,
+        pub password: String,
+    }
+
+    impl fmt::Debug for ClientSimpleLoginRequest {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.debug_struct("ClientSimpleLoginRequest")
+                .field("username", &self.username)
```
nitnelave, line 49: You're just missing `.field("password", "***********")` and you're golden ;)

nitnelave (changes requested):

auth/src/lib.rs
nitnelave: (and re-squash the commits)

haveachin: done

nitnelave (approved):

nitnelave:
Thanks a lot for the PR! And for being patient with the nits :)

PR #141, opened 2022-04-24
Add example configuration for dolibarr

Cyrix126:


example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
+# Configuration pour Dolibarr
+
+## This example will help you to create user in dolibarr from your users in your lldap server from a specific group and to login with the password from the lldap server.
```
nitnelave, line 3: Let's not make this a title (remove the ##).

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+
+## to connect ldap->dolibarr
```
nitnelave, line 5: Nit: Capital letter for "To"

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
+# Configuration pour Dolibarr
+
+## This example will help you to create user in dolibarr from your users in your lldap server from a specific group and to login with the password from the lldap server.
+
+## to connect ldap->dolibarr 
+
+Install module LDAP from Home -> Modules/Applications
```
nitnelave, line 7: Maybe clarify: "In Dolibarr, install the LDAP module..."

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+Go to the configuration of this module and fill it like this:
+
+
+Users and groups synchronization: LDAP -> Dolibarr
```
nitnelave, line 11: Can you put the values (here and below) between ticks `` ? To make them stand out and show that that's what the user should enter.

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+And two new tabs will appear on the top:
+Users and Groups
+
+We will use only Users in this example to get the users we want to import.
+The tab Groups would be to import groups.
+
+Click on the Users tab and fill it like this:
+Users' DN: ou=people,dc=example,dc=com
```
nitnelave, line 38: Same here, let's use ticks ``

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+List of objectClass: person
+Search filter: memberOf=cn=yournamegroup,ou=groups,dc=example,dc=com
```
nitnelave, line 40: Maybe add that if they don't have a group, they can just use the `uid=*` filter

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+
+##To import ldap users into the dolibarr database (needed to login with those users):
+
+Navigate to  Users & Groups -> New Users
+Click on the blank form "Users in LDAP database", you will get the list of the users in the group filled above. With the "GET" button, you will import the selected user.
+
+
+##To enable LDAP login:
+
+Modify your conf.php in your dolibarr folder in htdocs/conf
```
nitnelave, line 61: Can you put `conf.php` and `htdocs/conf` in ticks `` ?

nitnelave (changes requested):
Thanks a lot for the guide, it will be helpful for the next people who want to use Dolibarr :)

I just requested a few changes to make it more readable and consistent with the other guides, but it looks great!

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+Go to the configuration of this module and fill it like this:
+
+
+Users and groups synchronization: LDAP -> Dolibarr
+Contacts' synchronization: No
+Type: OpenLdap
+Version: Version 3
+Primary server: ldap://example.com
+Secondary server: Empty
+Server port: port 389
```
nitnelave, line 17: The LDAP port by default for LLDAP is 3890, let's change it to that.

example_configs/dolibarr.md
```diff
@@ -0,0 +1,88 @@
...
+$dolibarr_main_authentication='dolibarr'; 
+```
+
+with:
+```
+// Authentication settings
+// Only  add "ldap" to only login using the ldap server, or/and "dolibar" to compare with local users. In any case, you need to have the user existing in dolibarr.
+$dolibarr_main_authentication='ldap,dolibarr'; 
+$dolibarr_main_auth_ldap_host='ldap://127.0.0.1:389';
+$dolibarr_main_auth_ldap_port='389';
```
nitnelave, line 74: Same here, the default LLDAP port is 3890.

example_configs/dolibarr.md
Cyrix126: In this case we could just leave the filter empty ?

example_configs/dolibarr.md
Cyrix126: I modified other lines to make titles to separate the three steps. Is that ok ?

example_configs/dolibarr.md
nitnelave: You can give it a try, but I think LDAP expects a valid filter there. If we can leave it empty, that's even better.

example_configs/dolibarr.md
nitnelave: Almost :) You just need to make the first line a level 1 title (with only one #)

example_configs/dolibarr.md
nitnelave: Nit: search filter between ``

Cyrix126: it works without filter

nitnelave (approved):

nitnelave:
Thanks! Could you just squash your changes into a single commit?

nitnelave:
And rebase with master.

Cyrix126:
> Thanks! Could you just squash your changes into a single commit?

hmm how ? first time I do a pull request on github

nitnelave:
If you have a git client, then you can do:
```sh
git checkout main
git remote add upstream https://github.com/nitnelave/lldap.git
git pull upstream main
git rebase --interactive main
```

The last line will open an editor, and you replace the `pick` with `f` on all the lines _except the first one_, then you save and quit. Then you can do `git push -f`.
Or if you want, I can use your commits and create another PR to do it for you.

nitnelave:
Ah, no, not a merge! I prefer to keep the history linear, so rebases instead of merges. If that's too git-fu for you, I don't mind take it over (with your commits).

nitnelave:
Closed in favor of #146 

Thanks!

PR #143, opened 2022-04-24
Create Organizr.md

JaidenW:
Help document for configuring LDAP backend on Organizr

### DIFF
--- a/example_configs/Organizr.md
+++ b/example_configs/Organizr.md
@@ -1,5 +1,5 @@
 # Configuration for Organizr
-####  System Settings > Main > Authentication
+##  System Settings > Main > Authentication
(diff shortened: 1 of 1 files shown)


example_configs/Organizr.md
```diff
@@ -0,0 +1,40 @@
+# Configuration for Organizr
+####  System Settings > Main > Authentication
```
nitnelave, line 2: Let's use only two hash for a level 2 title ##

nitnelave (approved):
Thanks a lot! I just have a tiny formatting nit, but otherwise it's great :)

JaidenW (commented):
Updated formatting

PR #164, opened 2022-05-07
Improve Authentik Support by handlign wildcards + DN

### LINKED ISSUE #158: Authentik Support - attrs: ["*", "+"] not supported
This is more of a tracking issue to track my own process of getting this to work with authentik, however, i do not 

```
2022-05-04T05:56:29.541807Z DEBUG lldap::infra::ldap_handler: Received search request: LdapSearchRequest { base: "ou=people,dc=example,dc=com", scope: Subtree, aliases: Always, sizelimit: 0, timelimit: 0, typesonly: false, filter: Equality("objectClass", "person"), attrs: ["*", "+"] }    
2022-05-04T05:56:29.542227Z  INFO sqlx::query: SELECT `users`.`user_id`, `email`, `users`.`display_name`, …; rows: 12, elapsed: 256.336µs
```
...
```
2022-05-04T05:56:29.543631Z DEBUG lldap::infra::ldap_server: Replying with LDAP op: SearchResultDone(LdapResult { code: NoSuchAttribute, matcheddn: "", message: "Unsupported group attribute: objectSid", referral: [] })   
```

Although I'm unsure  if i've correctly config'd this on the authentik side ( maybe 85% sure), I get the impression that lldap just can't handle the `attrs: ["*", "+"]` part of the query. 

On authentik's end, that part of the query can't be customized, I assume it's some part of the lldap protocol that just isn't implemented by lldap

matthewstrasiotto:
Resolves #158

Review very welcome, I'm new to rust and found this challenging to write

Doesnt introduce regressions, but test cases are needed (I just don't want to write them 🙃 )

I haven't even run this against an authentik instance yet, I'll probably make a docker-compose for testing against one

### DIFF
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -120,11 +120,10 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
         // We ignore the operational attribute wildcard
         "+" => return Ok(None),
         "*" => {
-            warn!(
+            bail!(
                 "Matched {}, * should have been expanded into attribute list and * removed",
                 attribute
-            );
-            return Ok(None);
+            )
         }
         _ => {
             warn!("Ignoring unrecognized group attribute: {}", attribute);
@@ -209,11 +208,10 @@ fn get_group_attribute(
         // We ignore the operational attribute wildcard
         "+" => return Ok(None),
         "*" => {
-            warn!(
+            bail!(
                 "Matched {}, * should have been expanded into attribute list and * removed",
                 attribute
-            );
-            return Ok(None);
+            )
         }
         _ => {
             warn!("Ignoring unrecognized group attribute: {}", attribute);
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -108,7 +108,7 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
             "mailAccount".to_string(),
             "person".to_string(),
         ],
-        "dn" => vec![dn.to_string()],
+        "dn" | "distinguishedname" => vec![dn.to_string()],
         "uid" => vec![user.user_id.to_string()],
         "mail" => vec![user.email.clone()],
         "givenname" => vec![user.first_name.clone()],
@@ -152,7 +152,7 @@ fn get_group_attribute(
 ) -> Result<Option<Vec<String>>> {
     Ok(Some(match attribute.to_lowercase().as_str() {
         "objectclass" => vec!["groupOfUniqueNames".to_string()],
-        "dn" => vec![format!(
+        "dn" | "distinguishedname" => vec![format!(
             "cn={},ou=groups,{}",
             group.display_name, base_dn_str
         )],
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -6,6 +6,7 @@ use crate::domain::{
     opaque_handler::OpaqueHandler,
 };
 use anyhow::{bail, Context, Result};
+use itertools::Itertools;
 use ldap3_server::proto::{
     LdapBindCred, LdapBindRequest, LdapBindResponse, LdapExtendedRequest, LdapExtendedResponse,
     LdapFilter, LdapOp, LdapPartialAttribute, LdapPasswordModifyRequest, LdapResult,
@@ -116,9 +117,48 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
         "cn" | "displayname" => vec![user.display_name.clone()],
         "createtimestamp" | "modifytimestamp" => vec![user.creation_date.to_rfc3339()],
         "1.1" => return Ok(None),
-        _ => bail!("Unsupported user attribute: {}", attribute),
+        // We ignore the operational attribute wildcard
+        "+" => return Ok(None),
+        "*" => {
+            warn!(
+                "Matched {}, * should have been expanded into attribute list and * removed",
+                attribute
+            );
+            return Ok(None);
+        }
+        _ => {
+            warn!("Ignoring unrecognized group attribute: {}", attribute);
+            return Ok(None);
+        }
     }))
 }
+
+fn expand_attribute_wildcards(attributes: &[String], all_attribute_keys: &[&str]) -> Vec<String> {
+    let mut attributes_out = attributes.to_owned();
+
+    if attributes_out.iter().any(|x| x == "*") || attributes_out.is_empty() {
+        debug!(r#"Expanding * / empty attrs:"#);
+        // Remove occurrences of '*'
+        attributes_out.retain(|x| x != "*");
+        // Splice in all non-operational attributes
+        attributes_out.extend(all_attribute_keys.iter().map(|s| s.to_string()));
+    }
+
+    debug!(r#"Expanded: "{:?}""#, &attributes_out);
+
+    // Deduplicate, preserving order
+    attributes_out.into_iter().unique().collect_vec()
+}
+const ALL_USER_ATTRIBUTE_KEYS: &[&str] = &[
+    "objectclass",
+    "dn",
+    "uid",
+    "mail",
+    "givenname",
+    "sn",
+    "cn",
+    "createtimestamp",
+];
 
 fn make_ldap_search_user_result_entry(
     user: User,
@@ -126,9 +166,11 @@ fn make_ldap_search_user_result_entry(
     attributes: &[String],
 ) -> Result<LdapSearchResultEntry> {
     let dn = format!("uid={},ou=people,{}", user.user_id.as_str(), base_dn_str);
+
+    let expanded_attributes = expand_attribute_wildcards(attributes, ALL_USER_ATTRIBUTE_KEYS);
     Ok(LdapSearchResultEntry {
         dn: dn.clone(),
-        attributes: attributes
+        attributes: expanded_attributes
             .iter()
             .filter_map(|a| {
                 let values = match get_user_attribute(&user, a, &dn) {
@@ -164,9 +206,24 @@ fn get_group_attribute(
             .map(|u| format!("uid={},ou=people,{}", u, base_dn_str))
             .collect(),
         "1.1" => return Ok(None),
-        _ => bail!("Unsupported group attribute: {}", attribute),
+        // We ignore the operational attribute wildcard
+        "+" => return Ok(None),
+        "*" => {
+            warn!(
+                "Matched {}, * should have been expanded into attribute list and * removed",
+                attribute
+            );
+            return Ok(None);
+        }
+        _ => {
+            warn!("Ignoring unrecognized group attribute: {}", attribute);
+            return Ok(None);
+        }
     }))
 }
+
+const ALL_GROUP_ATTRIBUTE_KEYS: &[&str] =
+    &["objectclass", "dn", "uid", "cn", "member", "uniquemember"];
 
 fn make_ldap_search_group_result_entry(
     group: Group,
@@ -174,9 +231,11 @@ fn make_ldap_search_group_result_entry(
     attributes: &[String],
     user_filter: &Option<&UserId>,
 ) -> Result<LdapSearchResultEntry> {
+    let expanded_attributes = expand_attribute_wildcards(attributes, ALL_GROUP_ATTRIBUTE_KEYS);
+
     Ok(LdapSearchResultEntry {
         dn: format!("cn={},ou=groups,{}", group.display_name, base_dn_str),
-        attributes: attributes
+        attributes: expanded_attributes
             .iter()
             .filter_map(|a| {
                 let values = match get_group_attribute(&group, base_dn_str, a, user_filter) {
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -1636,6 +1636,151 @@ mod tests {
             ]
         );
     }
+    #[tokio::test]
+    async fn test_search_wildcards() {
+        use chrono::TimeZone;
+        let mut mock = MockTestBackendHandler::new();
+
+        mock.expect_list_users().returning(|_| {
+            Ok(vec![User {
+                user_id: UserId::new("bob_1"),
+                email: "bob@bobmail.bob".to_string(),
+                display_name: "Bôb Böbberson".to_string(),
+                first_name: "Bôb".to_string(),
+                last_name: "Böbberson".to_string(),
+                ..Default::default()
+            }])
+        });
+        mock.expect_list_groups()
+            .with(eq(Some(GroupRequestFilter::And(vec![]))))
+            .returning(|_| {
+                Ok(vec![Group {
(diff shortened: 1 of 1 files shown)


matthewstrasiotto:
@nitnelave this is tested against authentik, and should be ready for review, although it does need test cases.
I was wondering if you could help me with test cases? If not, that's okay

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +120,54 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+    attributes: &[String],
+    all_attribute_keys: &[&str],
+    operational_attribute_keys: &[&str],
+) -> Vec<String> {
+    let mut attributes_out = attributes.to_owned();
+
+    if let Some(all_idx) = attributes_out.iter().position(|x| x == "*") {
+        debug!(r#"Expanding *"#);
+        // Splice in all non-operational attributes
+        attributes_out.splice(
```
nitnelave, line 133: `splice` is not what you're looking for here. Maybe `extend_from_slice`?
I'm trying to think of ways to avoid copying too much, though, so how about this:
`make_ldap_search_user_result_entry` creates a `Vec<&str>` from the `&[String]`, and here we extend it with the 'static str since they're hardcoded.

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +120,54 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+            all_idx..all_idx + 1,
+            all_attribute_keys
+                .into_iter()
+                .cloned()
```
nitnelave, line 137: I'm not sure you need this clone. and `into_iter` can probably be just `iter` since it's readonly.

nitnelave (changes requested):

nitnelave:
For the tests, just create a test method at the end of the file, and call the functions you want to test with various inputs, using `assert` to check that the values are as expected.

server/src/infra/ldap_handler.rs
matthewstrasiotto: i think splice, or at least something that removes the original element, because we certainly don't want to keep the '*' and '+'

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +120,54 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+    return attributes_out;
+}
+
 fn make_ldap_search_user_result_entry(
     user: User,
     base_dn_str: &str,
     attributes: &[String],
 ) -> Result<LdapSearchResultEntry> {
     let dn = format!("uid={},ou=people,{}", user.user_id.as_str(), base_dn_str);
+    let all_attribute_keys = &["objectclass", "dn", "uid", "mail", "givenname", "sn", "cn"];
```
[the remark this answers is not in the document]
matthewstrasiotto: sounds good, where should that live?

server/src/infra/ldap_handler.rs
matthewstrasiotto: you're probably right about the `into_iter` but i definitely chained some of these for a reason (make compiler happy)

matthewstrasiotto: oh, actually, i think the `.map` call is probably what made compiler happy

server/src/infra/ldap_handler.rs
matthewstrasiotto: Consider replacing vecs withs sets 

> nitnelave — Today at 23:19
> Right, splice might be what we need indeed
> The constant can live next to the function
> Make sure that you test with multiple *
> strazto — Today at 23:20
> oh, true...
> nitnelave — Today at 23:20
> Actually, maybe having a set rather than vec would make sense
> strazto — Today at 23:20
> also true
> that's a bigger restructure though
> is supplying multiple * valid user input?
> / supplying multiple of any attribute key
> nitnelave — Today at 23:23
> There's valid input, and what people will try :) we should support it
> strazto — Today at 23:24
> as in, clean it up, or at least give a better error message
> nitnelave — Today at 23:24
> I think with the Set<&str> idea it shouldn't be too much effort
> strazto — Today at 23:25
> my experience is that when it comes to configuring ldap is that everyone ive ever seen try it just does random sh*t until it works
> nitnelave — Today at 23:25
> Now that I think about it, having string rather than str is probably okay, with the small string optimization
> And it'll probably be simpler to write
> strazto — Today at 23:25
> i think you've lost me
> nitnelave — Today at 23:26
> :D
> In your expand function, replace the input from &[String] to &mut Set<String>
> strazto — Today at 23:27
> gotcha, so just don't worry about changing String -> &str
> nitnelave — Today at 23:27
> Hmm, though apparently rust doesn't have that optimization
> I'll have to read more about it, but write with String to start with
> Then we'll see if we want to replace it
> strazto — Today at 23:28
> sounds good, can be optimzied if it's actually causing a problem later
> something something premature optimization
> nitnelave — Today at 23:28
> It's most likely premature optimization, but I don't care, it's my hobby project :D
> strazto — Today at 23:28
> that FASTEST hobby ever
> nitnelave — Today at 23:29
> Haha

matthewstrasiotto:
Current tasks:

- [ ] The constant can live next to the function
- [ ] Write test cases
   - [ ] Make sure that you test with multiple *
   - [ ] Change to sets to handle duplicates
 - [ ] Splice is okay, but maybe change vecs to sets anyway

server/src/infra/ldap_handler.rs
matthewstrasiotto: I think HashSets are premature optimization.

A valid query is `attrs: ['*', '+']` or `attrs: [ 'lldap schema attributes']` .

These are both bounded by O(1)

except in the edge case a user gives a tonne of duplicate attributes (  `attrs: ['*', '+', '*', '+', '*', '+', '*', '+', '*', '+', '*', '+', '*', '+', '*', '+', ]`
This behaviour is handled, but performance will be linearly bounded by number of duplicate attrs, which doesn't matter because they don't deserve good performance.

matthewstrasiotto:
@nitnelave ready for review. I didn't use sets for attribute uniqueness, i just dedup at the end of the wildcard expansion.

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +121,51 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+        debug!(r#"Removing +"#);
+        attributes_out.retain(|x| x != "+");
+    }
+
+    debug!(r#"Expanded: "{:?}""#, &attributes_out);
+
+    // Deduplicate, preserving order
+    attributes_out = attributes_out.into_iter().unique().collect_vec();
+
+    return attributes_out;
```
nitnelave, line 145: Just `attributes_out`, no return, no semicolon. Or same on the line above: 
`attributes_out.into_iter().unique().collect_vec()`

server/src/infra/ldap_handler.rs
```diff
@@ -1523,6 +1565,144 @@ mod tests {
             ]
         );
     }
+    #[tokio::test]
+    async fn test_search_wildcards() {
+        use chrono::TimeZone;
+        let mut mock = MockTestBackendHandler::new();
+
+        mock.expect_list_users().times(..).returning(|_| {
```
nitnelave, line 1573: Can't you just not write `.times`?

server/src/infra/ldap_handler.rs
```diff
@@ -1523,6 +1565,144 @@ mod tests {
...
+        let request3 = make_search_request(
+            "dc=example,dc=com",
+            LdapFilter::And(vec![]),
+            vec!["*", "+", "+"],
+        );
+
+        assert_eq!(ldap_handler.do_search(&request3).await, expected_result);
+
+        let request4 =
+            make_search_request("dc=example,dc=com", LdapFilter::And(vec![]), vec![""; 0]);
```
nitnelave, line 1702: I think you can just write `vec![]` here, no? Or does it complain about type inference?

server/src/infra/ldap_handler.rs
```diff
@@ -1523,6 +1565,144 @@ mod tests {
...
+
+        assert_eq!(ldap_handler.do_search(&request4).await, expected_result);
```
nitnelave, line 1704: Maybe add a test with `vec!["*", "uid", "distinguishedname"]`

Which brings the question: should we dedupe `dn` vs `distinguishedname`? I think it's okay not to.

nitnelave (changes requested):

server/src/infra/ldap_handler.rs
matthewstrasiotto: I'm not sure about this, I'll try it.
Might also be possible to omit times.

server/src/infra/ldap_handler.rs
matthewstrasiotto: It does complain about type inference

server/src/infra/ldap_handler.rs
matthewstrasiotto: probably better to not dedup those two, even though they're the same, in the same sense that `uid` and `cn` return the same value.

Although I imagine that the dn <-> distinguishedname alias is probably actually in the RFC somewhere, in the rare case someone requests both, they're likely going to be looking for whatever they ask for by key.

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +121,51 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+
+    if attributes_out.iter().any(|x| x == "+") {
+        debug!(r#"Removing +"#);
+        attributes_out.retain(|x| x != "+");
+    }
+
+    debug!(r#"Expanded: "{:?}""#, &attributes_out);
+
+    // Deduplicate, preserving order
+    attributes_out = attributes_out.into_iter().unique().collect_vec();
```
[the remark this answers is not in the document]
matthewstrasiotto: > Order requirement


```rust
# No order requirement, no need for elements to be adjacent for dedup
vec.iter().unique() 
``` 

```rust
# Order Requirement, elements must be adjacent
vec.dedup()

# Need to use
vec.sort().dedup()

```

I'll read the docs but I'm actually concerned that `HashSets` don't have a sense of order in their keys, and converting vecs to sets will probably break a bunch of test cases. 

You're probably right that it's not good that this function has the side effect of dedup.

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +121,51 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
+        debug!(r#"Expanding *"#);
+        // Splice in all non-operational attributes
+        attributes_out.extend(all_attribute_keys.iter().map(|s| s.to_string()));
+        // Remove occurrences of '*'
+        attributes_out.retain(|x| x != "*");
+    }
+
+    if attributes_out.iter().any(|x| x == "+") {
+        debug!(r#"Removing +"#);
+        attributes_out.retain(|x| x != "+");
```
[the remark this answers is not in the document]
matthewstrasiotto: Does the same go for `*`, or since we intend to handle `*` by first expanding & removing it, should we actually throw an error in `get_user_attribute` if somehow a `*` makes it that far?

I'll rephrase:

Should we
1. remove `*` in expansion, but retain `+`, 
   - ignore `+` in resolution
   - throw an error when resolving `*`, but give a useful error message
2. retain both in expansion, ignore both in resolution

server/src/infra/ldap_handler.rs
matthewstrasiotto: Oh - is this lint given by `cargo clippy` ?

server/src/infra/ldap_handler.rs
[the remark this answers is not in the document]
nitnelave: Yes, 1 is better. Actually, in the attribute translation we should ignore rather than error on unknown attributes: it was useful in the beginning, but now we have enough support to get away with ignoring, I think. Maybe we can still log ignored attributes with debug!

server/src/infra/ldap_handler.rs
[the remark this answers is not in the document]
nitnelave: I don't think keeping the order is important. As for the test cases, we can implement a comparison method to ignore the order.

server/src/infra/ldap_handler.rs
nitnelave: Indeed! And it's also nice Rust style. In general, keep return for exiting early from the function.

server/src/infra/ldap_handler.rs
```diff
@@ -120,15 +121,51 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
     }))
 }
 
+fn expand_attribute_wildcards(attributes: &[String], all_attribute_keys: &[&str]) -> Vec<String> {
```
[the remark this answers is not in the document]
matthewstrasiotto: can I? the attributes are different for the different object types

server/src/infra/ldap_handler.rs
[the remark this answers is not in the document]
matthewstrasiotto: is it okay if we save that for a different PR?

server/src/infra/ldap_handler.rs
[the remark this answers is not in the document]
nitnelave: Oh, I see, groups or users attributes. Nevermind then :)

server/src/infra/ldap_handler.rs
```diff
@@ -108,27 +109,68 @@ fn get_user_attribute(user: &User, attribute: &str, dn: &str) -> Result<Option<V
...
         "givenname" => vec![user.first_name.clone()],
         "sn" => vec![user.last_name.clone()],
         "cn" | "displayname" => vec![user.display_name.clone()],
         "createtimestamp" | "modifytimestamp" => vec![user.creation_date.to_rfc3339()],
         "1.1" => return Ok(None),
-        _ => bail!("Unsupported user attribute: {}", attribute),
+        // We ignore the operational attribute wildcard
+        "+" => return Ok(None),
+        "*" => {
+            warn!(
```
nitnelave, line 123: error!

server/src/infra/ldap_handler.rs
```diff
@@ -164,19 +206,36 @@ fn get_group_attribute(
             .map(|u| format!("uid={},ou=people,{}", u, base_dn_str))
             .collect(),
         "1.1" => return Ok(None),
-        _ => bail!("Unsupported group attribute: {}", attribute),
+        // We ignore the operational attribute wildcard
+        "+" => return Ok(None),
+        "*" => {
+            warn!(
```
nitnelave, line 212: error!

nitnelave (approved):
Thanks, it looks good now! Just a couple of nits about log levels, but it's pretty much ready.

server/src/infra/ldap_handler.rs
matthewstrasiotto: `error!` or `bail!`?

nitnelave: `bail!` is a good idea, actually.

PR #184, opened 2022-06-10
ldap: generate uuid attribute

nitnelave:


nitnelave:
Outdated, better implemented in master.

PR #193, opened 2022-06-20
Add filter by group

martadinata666:
Add User filter search by groups, as some safety practice to group the user and filter user available in that group, rather than querying all user in ldap.


nitnelave (changes requested):
I think it would be nice to have both: for those who don't really care, have the option without group filter, and for those who do, document how to add it.

PR #199, opened 2022-06-27
Added Docker option as per martadinata

MickMorley:


nitnelave:
Is that redundant with https://github.com/nitnelave/lldap/pull/198 ?

PR #200, opened 2022-06-28
Multiarch Docker Image

martadinata666:
Created a PR so we can talk about how we can provide docker image for easier deploy. Thanks.


nitnelave:
Nice work! I thought about making a github workflow for that, but since it's painful to test and there wasn't that much demand at the time, I didn't. Thanks for doing it :)

Quick question: how does it work with the labeling for docker images?
Also, it seems like you need to guard the uploads with "only on master", see https://github.com/nitnelave/lldap/blob/main/.github/workflows/docker.yml#L42

.github/workflows/Dockerfile.ci
```diff
@@ -0,0 +1,68 @@
...
+ARG DEBIAN_FRONTEND=noninteractive
+ARG TARGETPLATFORM
+RUN apt update && apt install -y wget
+WORKDIR /dim
+COPY bin/ bin/
+COPY web/ web/
+
+RUN mkdir -p target/
+RUN mkdir -p /lldap/app
+
```
nitnelave, line 11: Aren't you missing a build step here?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,258 @@
...
+        run: rustup default nightly
+      - name: smoke test
+        run: rustc --version
+      #- name: install cross
+      #  run: cargo install cross
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: compile aarch64
+        run: cargo build --target=aarch64-unknown-linux-gnu --release -p lldap -p migration-tool
+        #run: cross build --target=aarch64-unknown-linux-musl --release -p lldap -p migration-tool
```
nitnelave, line 125: If we don't need cross, let's remove it here.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,258 @@
...
+    branches: [ '*' ] #rolling build on main only
+    tags: 'v*' #tag for specific release v1.0.2,or anytag, start with v
+  pull_request:
+    branches: [ 'main' ] #build test on main, not pushing
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  
+jobs:
```
nitnelave, line 14: Can you add some comments about how sections are organized and what they do? It's not self-evident.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,258 @@
...
+        with:
+          name: ui
+          path: web
+      - name: setup qemu
+        uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/metadata-action@v3
+        id: meta
+        with:
+          #images: ghcr.io/${{ github.repository }}
```
nitnelave, line 230: Let's remove this.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,258 @@
...
+        
+      - name: Download llap ui artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: ui
+          path: web
+      - name: setup qemu
+        uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/metadata-action@v3
```
nitnelave, line 227: The login and push should be guarded to only happen on main.

nitnelave (changes requested):
Does this mean that the docker.yml is not needed anymore? If so, can you make sure that all the functionality is still there and then remove it?

.github/workflows/Dockerfile.ci
martadinata666: it for preparing copy file, as i COPY all the binary and the static first into container.
the web -> /lldap/app
the bin/ -> target/ 
and both copy out the last stage.

.github/workflows/docker-build.yml
martadinata666: Added some comments for more description

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,276 @@
...
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+# In total 5 jobs, all of the jobs are containerized
+# ---
+
+# build-ui , create/compile the web
+## Use rustlang/rust:nighlty image
```
nitnelave, line 18: Nit: nightly.

But actually, bigger nit: can we use stable? Or do we need nightly to cross-compile?

.github/workflows/docker-build.yml
martadinata666: > Nice work! I thought about making a github workflow for that, but since it's painful to test and there wasn't that much demand at the time, I didn't. Thanks for doing it :)
> 
> Quick question: how does it work with the labeling for docker images? Also, it seems like you need to guard the uploads with "only on master", see https://github.com/nitnelave/lldap/blob/main/.github/workflows/docker.yml#L42

The labeling, i usually use actions secrets like
`${{ secrets.DOCKERHUB_USERNAME }}/${{ secrets.DOCKERHUB_IMAGE_NAME }}`
But tecnically you can just override to your name `nitnelave/lldap`

I prevent to push on `docker/build-push-action`
`
          push: ${{ github.event_name != 'pull_request' }}
`
As your is preventing the login.

And `docker.yml` will obsolete, but just modify it to manual dispatch, backup is always welcome. As rust itself still pretty unstable we dont know when it breaks.

nitnelave: Right, I meant tagging. Read through your PR and suggested a couple of missing tags.

If the login doesn't work, the action fails, so that's unnecessary noise in PRs, can you add a guard to the login?

martadinata666: Sure 👍🏼

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,276 @@
+name: Build LLDAP Bin and Dockerize
+
+on:
+  push:
+    branches: [ '*' ] #branch to build
+    tags: 'v*' #tag for specific release v1.0.2,or anytag, start with v
```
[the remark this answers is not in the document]
martadinata666: I will match it your docker.yml

.github/workflows/docker-build.yml
nitnelave: If it passes, you'll just have to update the comments (here and below).

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,300 @@
...
+# build-docker-image job will fetch artifacts and run Dockerfile.ci then push the image.
+
+# On current https://hub.docker.com/_/rust
+# 1-bullseye, 1.61-bullseye, 1.61.0-bullseye, bullseye, 1, 1.61, 1.61.0, latest
+
+jobs:
+  build-ui:
+    runs-on: ubuntu-latest
+    container: 
+      image: rust:1.61
```
nitnelave, line 57: Actually, can we get the latest stable? (and update that in the comments above)

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,300 @@
+name: Build LLDAP Bin and Dockerize
```
nitnelave, line 1: Maybe just "Docker" for the name? That's clear enough and lets you see the arch more easily in the checks.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,300 @@
...
+        uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/metadata-action@v3
+        id: meta
+        with:
+          images: ${{ secrets.DOCKERHUB_USERNAME }}/${{ secrets.DOCKERHUB_IMAGE_NAME }}
+          tags: |
+            type=sha
+            type=ref,event=branch
+            type=ref,event=pr
```
nitnelave, line 270: I think this tag is failing, but we don't need it (since we don't push on a PR). And we don't need the branch either.

Maybe just get the tags I have already set up ? https://github.com/nitnelave/lldap/blob/main/.github/workflows/docker.yml#L73 with the slug action above.

.github/workflows/docker-build.yml
martadinata666: Well technically pass, except the desc thing 👍🏼 
https://github.com/martadinata666/lldap/runs/7095765397

.github/workflows/docker-build.yml
nitnelave: Regardless, can you use the tags that I already had? It handles cases like "1.2.3-beta1" and so on.

martadinata666: Well it should be ref https://github.com/docker/metadata-action , but let it same as docker.yml

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,295 @@
...
+          path: web
+          
+      - name: setup qemu
+        uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/metadata-action@v3
+        id: meta
+        with:
+          images: ${{ secrets.DOCKERHUB_USERNAME }}/${{ secrets.DOCKERHUB_IMAGE_NAME }}
+          tags: nitnelave/lldap:latest, nitnelave/lldap:stable, nitnelave/lldap:v${{ steps.slug.outputs.version-semantic }}, nitnelave/lldap:v${{ steps.slug.outputs.version-major }}, nitnelave/lldap:v${{ steps.slug.outputs.version-major }}.${{ steps.slug.outputs.version-minor }}, nitnelave/lldap:v${{ steps.slug.outputs.version-major }}.${{ steps.slug.outputs.version-minor }}.${{ steps.slug.outputs.version-patch }}
```
nitnelave, line 266: This requires the `slug` step, see the original docker.yml

nitnelave:
Can you update/remove the original docker.yml?

.github/workflows/docker-build.yml
martadinata666: done.

nitnelave (approved):

nitnelave:
Thanks a lot!

martadinata666:
Im not really sure about the tag, already follow the docker.yml , should be fine.

nitnelave:
The docker builds are taking a long time, and I realized that it's mostly due to package installations. I took great care to make the docker builds themselves fast, can we try to use cache better ? With https://github.com/actions/cache for instance.

martadinata666:
ah great suggestions, let see what i can add.

PR #204, opened 2022-06-30
hotfix, use Dockerfile.ci

martadinata666:
Ah sorry, accidentally remove the part it will use Dockerfile.ci

https://github.com/nitnelave/lldap/pull/200#issuecomment-1171044356


nitnelave:
I think we need 2 `build_and_push` steps: one for just `latest` that runs on every commit on master, and one for releases that pushes all the tags.

.github/workflows/docker-build.yml
```diff
@@ -281,21 +281,34 @@ jobs:
         id: slug
       
       - name: Login to Docker Hub
-        if: github.event_name != 'pull_request'
+        #if: github.event_name != 'pull_request'
```
nitnelave, line 284: We still need the guard. The push action will succeed without logging in if push is disabled.

nitnelave:
Check that you haven't broken it, I don't see any job as a result of the last push, so it's blocked expecting the two manually-defined required ones.

martadinata666:
Last results : 
https://github.com/nitnelave/lldap/actions/runs/2591225566

I think it ready.

PR #209, opened 2022-07-01
Update Authelia config for Authelia 4.36.1

grehund:
Authelia 4.36.1 changed the key for password resets. Fixed an error where the user: field was incorrectly set to uid=admin, resulting in a fatal error starting Authelia.

### DIFF
--- a/example_configs/authelia_config.yml
+++ b/example_configs/authelia_config.yml
@@ -43,6 +43,6 @@ authentication_backend:
     display_name_attribute: displayName
     # The username and password of the admin user.
     # "admin" should be the admin username you set in the LLDAP configuration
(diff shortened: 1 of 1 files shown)


nitnelave:
Thanks for the contribution! I'd like to keep the "uid=" since it's the most correct format (I shouldn't have started with cn= in the first place). If it doesn't work for you, it's probably because you're using the "stable" version instead of "latest"

grehund:
ok, thanks for letting me know. Yes, I'm using the "stable" version.

nitnelave:
Can you update your PR to revert the `cn=` change? I'd still like to merge the password config update.

nitnelave:
Thanks!

PR #213, opened 2022-07-01
Add cache and sccache

martadinata666:
From my test it just shaved about 2~3 mnites. Not so impressive.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -18,6 +18,10 @@ on:
         
 env:
   CARGO_TERM_COLOR: always
+  RUSTC_WRAPPER: sccache
+  SCCACHE_DIR: $GITHUB_WORKSPACE/.sccache
+  SCCACHE_VERSION: v0.3.0
+  LINK: https://github.com/mozilla/sccache/releases/download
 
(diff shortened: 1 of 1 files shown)


nitnelave:
Hmm, that's interesting, but I was more thinking about caching the workflow steps: can we use docker-like caching of steps for installing npm, cargo, wasm-pack and so on? If they jump directly into the docker step with exactly the same context, the cache is going to apply and we'll save a lot, especially with the dependencies. And we can use cargo chef to precompile the dependencies and cache that

martadinata666:
Im not really familiar with directory structure of rust, according to this we can also cache `target` dir, https://doc.rust-lang.org/cargo/guide/build-cache.html.

About chef, like your maik dockerfile? Cargo chef cook? Well we can just try what works or not. Sccache told reduce compiling seems not right 😂

martadinata666:
This should be sufficient, caching sccache, target, cargo. Minimal download. on first build will pretty slow due first cache creation, pretty big, almost 1GB 😱

nitnelave:
Normally, it should pick up the changes in cargo.toml and recompile. Hopefully it doesn't always recompile just because the context is different. We'll have to see, but let's start with your change, it's already much better, thanks!

PR #215, opened 2022-07-03
Added a dark theme to the style.css

JaidenW:


JaidenW:
<footer class="text-center text-muted fixed-bottom bg-light"> Should remain the same, i didn't realize that it was fixed and overlaps some elements, the bg-light tag is required, & handled via CSS override :)

app/static/style.css
```diff
@@ -10,3 +10,55 @@ header h1 {
   font-weight: 700;
   text-decoration: none;
 }
+
+/* Super simple dark theme - By Jaiden W */
+
+body.dark-theme{
+  background:#202020;
+  color:#fff!;
```
nitnelave, line 18: I think we don't want pure white on pure black. If we take Material design's recommendations (not that it's gospel, but it's a good starting guideline) we'd have a background of #121212 for the very back, up to #242424 for the most accentuated background, and a text color of at most #dedede

nitnelave (changes requested):

nitnelave:
What would you think of something like https://vinorodrigues.github.io/bootstrap-dark/ ? Rather than creating/maintaining our own dark theme.

JaidenW:
That might be better, especially if you intend to add anything else to it in the future, that would likely also likely make it easier for you to style things on your own

nitnelave:
Closing in favor of https://github.com/nitnelave/lldap/issues/250

PR #218, opened 2022-07-05
Docker build static

martadinata666:
Trying enabling staticc build, currently let keep the build separate to see the behaviour.
1. Add and using Dockerfile.ci.alpine
2. Modify `:latest` to `:latest-alpine`


nitnelave:
Is this ready for review?

martadinata666:
No, i accidentally rename current `docker-build.yml` instead add new `docker-build-static.yml`, will fix later.

.github/workflows/docker-build-static.yml
```diff
@@ -0,0 +1,385 @@
...
+            /usr/local/cargo/registry/cache
+            /usr/local/cargo/git/db
+            target
+          key: lldap-ui-${{ github.sha }}
+          restore-keys: |
+            lldap-ui-
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: install cargo wasm
+        run: cargo install wasm-pack
```
nitnelave, line 88: maybe add an ` || true` here: if the binary is restored from cache, cargo install will "fail" because the binary already exists.

nitnelave:
Is this ready for review? (assuming that the build passes)

.github/workflows/docker-build-static.yml
```diff
@@ -0,0 +1,386 @@
...
+      - name: Build and push latest
+        if: github.event_name != 'release'
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: ${{ github.event_name != 'pull_request' }}
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          file: ./.github/workflows/Dockerfile.ci.alpine
+          tags: nitnelave/lldap:latest-alpine
+          #cache-from: type=gha
```
nitnelave, line 354: Remove these comments

.github/workflows/docker-build-static.yml
```diff
@@ -0,0 +1,386 @@
...
+          path: |
+            .cargo/bin
+            .cargo/registry/index
+            .cargo/registry/cache
+            .cargo/git/db
+            target
+          key: lldap-bin-amd64-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            lldap-bin-amd64-
+      #- name: add cargo chef
```
nitnelave, line 232: This part didn't work? Or it's not needed because we cache ./target? In that case can you remove it?

.github/workflows/docker-build-static.yml
martadinata666: This just weird behaviour, when the other one can skip installation properly but this one cant. 😮‍💨

PR #220, opened 2022-07-05
Remove sccache, cache key using cargo lock hash

martadinata666:
add cargo update, check sccache show stats to see if it caching correctly

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -18,10 +18,8 @@ on:
         
 env:
   CARGO_TERM_COLOR: always
-  RUSTC_WRAPPER: sccache
-  SCCACHE_DIR: $GITHUB_WORKSPACE/.sccache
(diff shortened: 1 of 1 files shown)


martadinata666:
Just removed sccahe, and reset cache if cargo.lock changes.

nitnelave:
This is out of date, no?

martadinata666:
No, this still stands for non-static build. As currently still using github sha instead cargo.lock.

nitnelave:
Yeah, let's simplify, drop the debian build, and keep just the alpine one (let's remove `-alpine` from the tags).

Also, could you partly restore the docker build using the top-level Dockerfile, to make sure that people can still build their own docker image if they want? Without pushing to DockerHub

martadinata666:
I see, i will drop this PR, and do new PR will default to alpine. And for the top level dockerfile let see what i can improve.

Issue #230, opened 2022-07-09
[Enhancement] Support custom UID:GID


martadinata666:
Add support for custom UID/GID.
1. Use gosu
2. Modified entrypoint
3. Will use UID 1000 if not defined by user.

Sample of usage after supported
```
volumes:
  lldap_data:
    driver: local

services:
  lldap:
    image: nitnelave/lldap:stable
    ports:
      # For LDAP
      - "3890:3890"
      # For the web front-end
      - "17170:17170"
    volumes:
      - "lldap_data:/data"
      # Alternatively, you can mount a local folder
      # - "./lldap_data:/data"
    environment:
      - UID=2000 # desired UID
      - GID=3000 # desired GID
      - LLDAP_JWT_SECRET=REPLACE_WITH_RANDOM
      - LLDAP_LDAP_USER_PASS=REPLACE_WITH_PASSWORD
      - LLDAP_LDAP_BASE_DN=dc=example,dc=com
```

Any suggestions?

nitnelave:
I think that sounds right. We might want to do in the entry point a chown of the DB (& private key?) to the given UID to solve most permission issues.
+1 to `gosu`, it looks like a good way to go.

PR #235, opened 2022-07-11
Enabling UID:GID

martadinata666:
Following #230 
1. Add gosu package
2. Modify entrypoint to chown `/app` and `/data` to UID:GID


nitnelave:
Thanks!

PR #244, opened 2022-07-13
More info about `:latest` tag and permissions

martadinata666:
Currently there is significant difference `:stable` tag and `:latest` tag, as people behaviour of using `:latest`, give notice that latest prone to break, changes.
1. Add compose sample for `:latest`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@
  - [Client configuration](#Client-configuration)
    - [Compatible services](#compatible-services)
    - [General configuration guide](#general-configuration-guide)
-   - [Sample cient configurations](#Sample-client-configurations)
+   - [Sample client configurations](#Sample-client-configurations)
  - [Comparisons with other services](#Comparisons-with-other-services)
    - [vs OpenLDAP](#vs-openldap)
    - [vs FreeIPA](#vs-freeipa)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,12 +90,16 @@ Configure the server by copying the `lldap_config.docker_template.toml` to
 Environment variables should be prefixed with `LLDAP_` to override the
 configuration.
 
+If the `lldap_config.toml` not existed or created beforehand LLDAP will use default one, with `password` as the default password, after login you can update admin password via web interface.
```
nitnelave, line 93: "If the `lldap_config.toml` doesn't exist when starting up, LLDAP will use the default one. The default admin password is `password`, you can change it using the web interface.

README.md
```diff
@@ -122,6 +126,33 @@ services:
       - LLDAP_LDAP_BASE_DN=dc=example,dc=com
 ```
 
+Example for docker compose for `:latest` tag:
```
nitnelave, line 129: Can you add a couple of words on the difference?

nitnelave (changes requested):

README.md
martadinata666: This is because our `:latest` using UID:GID  when people it also expected it working at stable. Can we create a release like `0.4.1` so the UID:GID consistently implemented in our image. Then we only need to change the current compose using UID:GID?

PR #246, opened 2022-07-13
Set admin email from config

frankmoskal:
Allows folks to set `ldap_user_email` alongside `ldap_user_dn` and `ldap_user_pass` via the configuration file / environment variables. I need to have my admin account provisioned with an email at creation, so this ends up being pretty useful :)

I'm having issues with building a Docker image / the linker (on Debian). but I don't think due to my changes? Please let me know if I've forgotten anything!

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -45,6 +45,11 @@
 ## For the administration interface, this is the username.
 #ldap_user_dn = "admin"
 
+## Admin email.
+## Email for the admin account. It is only used when initially creating
+## the admin user, and can safely be omitted.
+#ldap_user_email = "admin@example.com"
+
 ## Admin password.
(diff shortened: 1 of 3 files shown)


lldap_config.docker_template.toml
```diff
@@ -45,6 +45,11 @@
 ## For the administration interface, this is the username.
 #ldap_user_dn = "admin"
 
+## Admin email.
+## Email for the admin account. It is only used when initially creating
+## the admin user, and can safely be omitted.
+#ldap_user_email = "admin"
```
nitnelave, line 51: Can you give a better example value? This is not a valid email.

nitnelave (approved):
Thanks for the PR!

lldap_config.docker_template.toml
frankmoskal: 🤦🏻 oops, fixed!

PR #251, opened 2022-07-13
Dockerfile rewrite

martadinata666:
1. Update Dockerfile as close as possible to CI build.
2. Tested fine on amd64
3. Test on arm64 in progress.
4. Need some docs to non native build.


Dockerfile
```diff
@@ -1,66 +1,117 @@
...
-    && adduser --disabled-password \
-        --gecos '' \
-        --ingroup app \
-        --home /app \
-        --uid 10001 \
-        app \
-    # Install required packages
-    && apk add npm openssl-dev musl-dev make perl curl
+# Set env for our builder
+ENV CARGO_TERM_COLOR=always
```
nitnelave, line 5: You can merge all the ENV lines into a single one:
```Dockerfile
ENV X=Y \
    X2=Y2
```

Dockerfile
```diff
@@ -1,66 +1,117 @@
...
+#    apt install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-arm64-cross libc6-dev-arm64-cross libssl-dev:arm64 && \
+#    apt clean && \
+#    rm -rf /var/lib/apt/lists/*
+### add arm64 target
+#RUN rustup target add aarch64-unknown-linux-gnu
 
-# Prepare the dependency list.
-FROM chef AS planner
-COPY . .
-RUN cargo chef prepare --recipe-path /tmp/recipe.json
```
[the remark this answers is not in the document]
martadinata666: Is it make some local cache? Im thinking as intermedia container usually cleaned-up and we restart build from scratch 🤔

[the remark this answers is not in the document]
nitnelave: It makes a cache that is saved in the intermediate docker image. The step can be entirely cached and restored by the docker daemon. That may not be entirely applicable for CI, but for local dev it's essential.

Dockerfile
```diff
@@ -49,12 +45,32 @@ RUN npm install -g wasm-pack && \
 ### add armhf target
 #RUN rustup target add rustup target add armv7-unknown-linux-gnueabihf
 
+
+# Install cargo-chef and wasm-pack via cargo [npm dont have arm64 bin]
```
nitnelave, line 49: Nit: `npm doesn't have`

Dockerfile
```diff
@@ -1,66 +1,134 @@
...
+COPY . .
+RUN cargo chef prepare --recipe-path /tmp/recipe.json
+
+# Build depedencies
+FROM builder-base AS builder
+COPY --from=planner /tmp/recipe.json recipe.json
+RUN RUSTFLAGS="-Ctarget-feature=-crt-static" cargo chef cook --release -p lldap_app --target wasm32-unknown-unknown
+RUN RUSTFLAGS="-Ctarget-feature=-crt-static" cargo chef cook --release -p lldap
+RUN RUSTFLAGS="-Ctarget-feature=-crt-static" cargo chef cook --release -p migration-tool
+COPY . /lldap-src
```
nitnelave, line 68: why do you move to another directory?
I think in that case the `target/` directory with all the dependencies is not going to follow and you'll be throwing away all the work that you just did.

Dockerfile
```diff
@@ -1,66 +1,134 @@
...
+# Prepare dependencies
+FROM builder-base AS planner
 WORKDIR /app
+COPY . .
+RUN cargo chef prepare --recipe-path /tmp/recipe.json
+
+# Build depedencies
+FROM builder-base AS builder
+COPY --from=planner /tmp/recipe.json recipe.json
+RUN RUSTFLAGS="-Ctarget-feature=-crt-static" cargo chef cook --release -p lldap_app --target wasm32-unknown-unknown
```
nitnelave, line 65: It makes sense to separate them, but I'd change the order to migration-tool, lldap_app and then lldap, given which are most likely to change deps.

Dockerfile
```diff
@@ -1,66 +1,134 @@
...
-ENV LDAP_PORT=3890
-ENV HTTP_PORT=17170
 
-EXPOSE ${LDAP_PORT} ${HTTP_PORT}
 
-ENTRYPOINT ["/app/docker-entrypoint.sh"]
+### Final image
+FROM alpine:3.16
+WORKDIR /app
+ENV UID=1000
```
nitnelave, line 112: Same, you can merge the ENV steps

Dockerfile
```diff
@@ -1,66 +1,134 @@
...
+    --uid "$UID" \
+    "$USER" && \
+    mkdir -p /data && \
+    chown $USER:$USER /data
+### Copy out the binary and web from builder
+COPY --from=builder --chown=$USER:$USER /lldap /app
+COPY --from=builder --chown=$USER:$USER /docker-entrypoint.sh /docker-entrypoint.sh
+VOLUME ["/data"]
+WORKDIR /app
+ENTRYPOINT ["tini", "--", "/docker-entrypoint.sh"]
```
nitnelave, line 133: I think we don't need both the entrypoint and the CMD, you can merge them into an entrypoint.

Dockerfile
```diff
@@ -1,66 +1,134 @@
...
+COPY lldap_config.docker_template.toml /lldap/
+# The applications
+RUN cp target/*/release/lldap /lldap/lldap && \
+    cp target/*/release/migration-tool /lldap/migration-tool && \
+    cp -R app/index.html \
+          app/pkg \
+          app/static \
+          /lldap/app/
+# Just checking
+RUN ls -al /lldap && \
```
nitnelave, line 98: Umm, why? We have a cp just above that copies to /lldap/app, so it'll always exist.

PR #302, opened 2022-09-22
Update grafana_ldap_config.toml

dl1ch:
fix typos and attributes

### DIFF
--- a/example_configs/grafana_ldap_config.toml
+++ b/example_configs/grafana_ldap_config.toml
@@ -20,7 +20,7 @@ ssl_skip_verify = false
 # client_key = "/path/to/client.key"
 
 # Search user bind dn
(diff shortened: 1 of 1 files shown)


martadinata666:
Looking from https://grafana.com/docs/grafana/v9.0/setup-grafana/configure-security/configure-authentication/ldap/#enable-ldap, seems `cn` is right?

nitnelave:
> Looking from https://grafana.com/docs/grafana/v9.0/setup-grafana/configure-security/configure-authentication/ldap/#enable-ldap, seems `cn` is right?

No, it seems the PR is right: although `cn` is a common attribute for LDAP, LLDAP uses the `uid` as non-modifiable identifier.

nitnelave (approved):

nitnelave:
Thanks for the contribution, @dl1ch !

PR #311, opened 2022-10-02
Fix download assets js and fonts

martadinata666:
Seems we oversight the download path of js and fonts.

https://github.com/nitnelave/lldap/issues/310

### DIFF
--- a/.github/workflows/Dockerfile.ci.alpine
+++ b/.github/workflows/Dockerfile.ci.alpine
@@ -54,7 +54,7 @@ RUN set -x \
     && chmod a+r -R .; \
     ls -al .; \
     ls -al app/; \
-    cat app/static/index.html; \
+    cat app/index.html; \
     ls -al app/static; \
     ls -al app/static/fonts;
     
--- a/.github/workflows/Dockerfile.ci.alpine
+++ b/.github/workflows/Dockerfile.ci.alpine
@@ -46,10 +46,13 @@ RUN cp target/lldap /lldap/ && \
           web/static \
           /lldap/app/
 
+WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
     && chmod a+r -R .
+    && ls -al app/static
+    && ls -al app/static/fonts
     
 FROM alpine:3.16
 WORKDIR /app
--- a/.github/workflows/Dockerfile.ci.alpine
+++ b/.github/workflows/Dockerfile.ci.alpine
@@ -39,6 +39,7 @@ RUN if [ "${TARGETPLATFORM}" = "linux/arm/v7" ]; then \
 # Web and App dir
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 COPY lldap_config.docker_template.toml /lldap/
+COPY web/index_local.html web/index.html
 RUN cp target/lldap /lldap/ && \
     cp target/migration-tool /lldap/ && \
     cp -R web/index.html \
@@ -50,9 +51,11 @@ WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
-    && chmod a+r -R .
-RUN ls -al app/static && \
-    ls -al app/static/fonts
+    && chmod a+r -R .; \
+    && ls -al .; \
+    && ls -al app/static; \
+    && cat app/static/index.html; \
+    && ls -al app/static/fonts;
     
 FROM alpine:3.16
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


.github/workflows/Dockerfile.ci.alpine
```diff
@@ -46,10 +46,13 @@ RUN cp target/lldap /lldap/ && \
           web/static \
           /lldap/app/
 
+WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
     && chmod a+r -R .
+RUN ls -al app/static && \
```
nitnelave, line 54: What's the purpose of this `ls` ?

.github/workflows/docker-build-static.yml
```diff
@@ -91,6 +91,8 @@ jobs:
         run: ./app/build.sh
       - name: check path
         run: ls -al app/
+      - name: overwrite index.html with index_local.html
```
nitnelave, line 94: Shouldn't that be part of the Dockerfile?

nitnelave (changes requested):

.github/workflows/Dockerfile.ci.alpine
martadinata666: it just to make sure, the file downloaded correctly.

.github/workflows/docker-build-static.yml
martadinata666: This can both ways, it just when doing via CI it affect both container directly instead modifying two Dockerfiles. 🤔

.github/workflows/Dockerfile.ci.alpine
nitnelave: Then merge it with the `RUN` before (one fewer layer in the image).

.github/workflows/docker-build-static.yml
nitnelave: I'd rather keep it in the Dockerfiles: this is part of building the image, not part of checking it.

.github/workflows/Dockerfile.ci.alpine
```diff
@@ -39,17 +39,23 @@ RUN if [ "${TARGETPLATFORM}" = "linux/arm/v7" ]; then \
...
           web/pkg \
           web/static \
           /lldap/app/
 
+WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
-    && chmod a+r -R .
+    && chmod a+r -R .; \
```
nitnelave, line 54: I think you want `&&`: You want the command to fail if any of these fail.

.github/workflows/Dockerfile.ci.alpine
```diff
@@ -39,17 +39,23 @@ RUN if [ "${TARGETPLATFORM}" = "linux/arm/v7" ]; then \
...
+    ls -al .; \
+    ls -al app/static; \
+    cat app/static/index.html; \
```
nitnelave, line 57: why `cat`? I think `ls` is better to check the presence of the file.

.github/workflows/Dockerfile.ci.debian
```diff
@@ -46,10 +46,13 @@ RUN cp target/lldap /lldap/ && \
           web/static \
           /lldap/app/
 
+WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
     && chmod a+r -R .
+RUN ls -al app/static && \
```
nitnelave, line 54: Same here, we want to merge the stages.
And we need to use index_local as well.

.github/workflows/Dockerfile.ci.alpine
nitnelave: You still want `&&` instead of `;` to separate the commands.

.github/workflows/Dockerfile.ci.alpine
```diff
@@ -39,17 +39,24 @@ RUN if [ "${TARGETPLATFORM}" = "linux/arm/v7" ]; then \
...
 
+WORKDIR /lldap
 RUN set -x \
     && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
     && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
-    && chmod a+r -R .
+    && chmod a+r -R . \
+    && ls -al . \
+    && ls -al app/ \
+    && cat app/index.html \
```
nitnelave, line 57: Instead of `cat` maybe `[ -f app/index.html ]`? We just want to test whether the file exists.
Similarly, the `ls` can be replaced with existence tests like `[ -d app/ ]`. And we don't need to test whether `.` exists.

martadinata666: Do we need test every file? Because practically we need to see is everything downloaded correctly, but also we can remove it. Or we can just remove it, as after few runs it downloaded and copied fine.

nitnelave: yeah, I think you can do without testing the files.

nitnelave (approved):
Thanks!

Issue #242, opened 2022-07-12
Attach artifacts to release for popular architectures 


onedr0p:
It would be nice if you could attach the binary for lldap to the releases page for popular architectures and platforms like gnu, musl, amd64 and arm64.

This could be done with https://github.com/rust-build/rust-build.action perhaps?

You also might be able to look at vector for github workflows that compile rust across architectures:

https://github.com/vectordotdev/vector/releases/tag/v0.23.0

nitnelave:
Updated release v0.4.1 to include some artifacts: https://github.com/nitnelave/lldap/releases/tag/v0.4.1

If someone wants to contribute a github workflow that does it automatically for releases, it'd be welcome. Maybe whenever a tag is pushed, it creates a release draft with the artifacts (and then I can come and fix the changelog).

PR #325, opened 2022-10-12
add create release artifacts

martadinata666:
Add artifacts to release.

Sample result : https://github.com/martadinata666/lldap/releases/tag/v0.4.2

#242


martadinata666:
> So in that workflow, I create a release with no artifact then this runs and populates the artifacts automatically?

im testing with create tag -> create release. Let see if we can publish first, then will it override the release, 🤔

nitnelave:
Great, thanks for the improvement!

PR #334, opened 2022-10-15
Added dell idrac sample

Roemer:


example_configs/dell_idrac.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+### Search Filter
+```
+(&(objectClass=person)(memberof=cn=idrac_users,ou=groups,dc=example,dc=com))
+```
+
+For the Group Role Mappings, you define groups by their full `Group DN`, eg.
+```
+cn=idrac_users,ou=groups,dc=int,dc=flau-intra,dc=net
```
nitnelave, line 56: Example.com

nitnelave (approved):

nitnelave (approved):
Thanks!

PR #350, opened 2022-10-20
Ensure generated JWT doesn't include quotation mark

### LINKED ISSUE #349: JWT instructions in sample config file can cause an error
Ran into the following error after generating a new JWT key using the instructions in the [lldap_config.docker_template.toml](https://github.com/nitnelave/lldap/blob/main/lldap_config.docker_template.toml) file:

```
$ ./lldap run
Loading configuration from lldap_config.toml
Error: expected newline, found an identifier at line 31 column 40 in lldap_config.toml TOML file
```

Turns out the command generated a key with a quotation mark in it, and I didn't notice it when copy&pasting it into the config file.

This might become moot with #262, but I'll submit a quick PR nonetheless.

IndrekHaav:
Closes #349

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -27,7 +27,7 @@
 ## This can also be set from a file's contents by specifying the file path
 ## in the LLDAP_JWT_SECRET_FILE environment variable
 ## You can generate it with (on linux):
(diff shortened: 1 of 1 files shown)


nitnelave:
Thanks!

PR #355, opened 2022-10-28
docker: add tzdata package, to support TZ env variable

ajgon:
As PR title says, it will add `tzdata` package to the final docker image. Thanks to that, it will be easy to set timezone inside of the container, via `TZ=` env variable.


nitnelave:
You'll need to add it to the debian and alpine containers in https://github.com/nitnelave/lldap/tree/main/.github/workflows : those are the ones actually used for distribution. The top-level one is more for development.

ajgon:
Done. I think Debian has it out of the box, but I added it, just to be sure.

PR #363, opened 2022-11-02
Update readme to use nodejs 16

martadinata666:
as many variant of system, and different nodejs out there, define nodejs version to match the CI build.


nitnelave:
Thanks!

PR #377, opened 2022-11-24
server: make `host` configurable to enable IPv6 support

### LINKED ISSUE #359: ipv6 bind support
hello!

I've been testing this and it works great for systems on our internal network, which uses ipv4.

When we try connect from outside via ipv6 though, it's refused. It's not a firewall issue, IPs are already whitelisted.
It looks to only be binding on ipv4 addresses:

```
netstat -tulpn | grep lldap
tcp        0      0 0.0.0.0:3890            0.0.0.0:*               LISTEN      107866/lldap        
tcp        0      0 0.0.0.0:17170           0.0.0.0:*               LISTEN      107866/lldap        
```

I'm not seeing a config option for which interface to bind to either, so presumably it's not implemented?

Would it be possible to get ipv6 support? We can continue without for the affected machines but it'd be really sweet to get everything working with LLDAP :)

Thanks so much for your work, it's an awesome project!

wheinze:
Introduce a new configuration variable called `host` to enable IPv6 support.

If by any means, the naming of the variable should be adjusted, or a bool is preferred to toggle this behavior, I am for sure willing to adjust this ;)

Resolves #359

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -8,11 +8,15 @@
 # verbose=false
 
 ## The host address that the LDAP server will be bound to.
-## To enable IPv6 support, simply switch "host" to "::1":
-#host = "0.0.0.0"
+## To enable IPv6 support, simply switch "ldap_host" to "::1":
+#ldap_host = "0.0.0.0"
 
 ## The port on which to have the LDAP server.
 #ldap_port = 3890
+
+## The host address that the HTTP server will be bound to.
+## To enable IPv6 support, simply switch "api_host" to "::1":
+#api_host = "0.0.0.0"
 
 ## The port on which to have the HTTP server, for user login and
 ## administration.
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -7,6 +7,10 @@
 ## You can set it with the LLDAP_VERBOSE environment variable.
 # verbose=false
 
+## The host address that the LDAP server will be bound to.
+## To enable IPv6 support, simply switch "host" to "::1":
+#host = "0.0.0.0"
+
 ## The port on which to have the LDAP server.
 #ldap_port = 3890
 
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -15,8 +15,8 @@
 #ldap_port = 3890
 
 ## The host address that the HTTP server will be bound to.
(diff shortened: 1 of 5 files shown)


nitnelave:
I think we should split the host into ldap and http parts (`api_host` and `ldap_host`): Some people will want to have LDAP only available on localhost for safety, whereas they will expose the API to the web.

wheinze:
@nitnelave Thanks for the quick feedback and your hints on how to fully integrate into CLI as well ;)

I provided `ldap_host` and `api_host` as suggested by you.

I'm just wondering if it would make sense to align on other HTTP API vars and rather call it `http_host` :thinking: 
What do you think?

nitnelave:
Oh yeah, definitely. I didn't have a look at the code when suggesting it, keeping it consistent is better

wheinze:
> Oh yeah, definitely. I didn't have a look at the code when suggesting it, keeping it consistent is better

Adjusted to use `http_host` ;)

lldap_config.docker_template.toml
```diff
@@ -7,9 +7,17 @@
 ## You can set it with the LLDAP_VERBOSE environment variable.
 # verbose=false
 
+## The host address that the LDAP server will be bound to.
+## To enable IPv6 support, simply switch "ldap_host" to "::1":
```
nitnelave, line 11: Can you add something like:

"To only allow connections from localhost (if you want to restrict to local self-hosted services), change it to "127.0.0.1" ("::1" in IPv6)"

And yes, btw, "::1" is localhost, you want to suggest "::" for the general case for IPv6.

nitnelave (commented):
I take it you've tested it with IPv6? :)

Looks good, just a small comment on the config template.

lldap_config.docker_template.toml
wheinze: > Can you add something like:
> 
> "To only allow connections from localhost (if you want to restrict to local self-hosted services), change it to "127.0.0.1" ("::1" in IPv6)"

Done: https://github.com/nitnelave/lldap/pull/377/commits/578d5e1fdb5c407f33c2d46f04519f20dd74c22d

> And yes, btw, "::1" is localhost, you want to suggest "::" for the general case for IPv6.

yep, sure thx. did correct it.

wheinze:
> I take it you've tested it with IPv6? :)

yup, I tested all combinations with IPv4 and IPv6 for `ldap_host` and  `http_host` on my local dev setup.

PR #379, opened 2022-11-24
fix: use async api for email sending

### LINKED ISSUE #378: Async SMTP
The current implementation utilizes synchronous SMTP API which can lead to hangs of the whole server during password reset if SMTP server is not responding (timeout). It would be nice to use asynchronous version to make sure it won't block the main thread.

Michsior14:
Resolves #378

### DIFF
--- a/server/src/infra/mail.rs
+++ b/server/src/infra/mail.rs
@@ -54,15 +54,15 @@ To reset your password please visit the following URL: {}/reset-password/step2/{
 Please contact an administrator if you did not initiate the process.",
         username, domain, token
     );
-    return send_email(to, "[LLDAP] Password reset requested", body, options).await;
+    send_email(to, "[LLDAP] Password reset requested", body, options).await
 }
 
 pub async fn send_test_email(to: Mailbox, options: &MailOptions) -> Result<()> {
-    return send_email(
+    send_email(
         to,
         "LLDAP test email",
         "The test is successful! You can send emails from LLDAP".to_string(),
         options,
     )
-    .await;
+    .await
 }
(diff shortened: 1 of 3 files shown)


server/src/infra/mail.rs
```diff
@@ -54,14 +54,15 @@ To reset your password please visit the following URL: {}/reset-password/step2/{
 Please contact an administrator if you did not initiate the process.",
         username, domain, token
     );
-    send_email(to, "[LLDAP] Password reset requested", body, options)
+    return send_email(to, "[LLDAP] Password reset requested", body, options).await;
```
nitnelave, line 57: You don't need the `return`:
```rust
send_email(...).await
```

server/src/infra/mail.rs
```diff
@@ -54,14 +54,15 @@ To reset your password please visit the following URL: {}/reset-password/step2/{
...
 }
 
-pub fn send_test_email(to: Mailbox, options: &MailOptions) -> Result<()> {
-    send_email(
+pub async fn send_test_email(to: Mailbox, options: &MailOptions) -> Result<()> {
+    return send_email(
```
nitnelave, line 61: Same here, no return.

nitnelave (changes requested):

nitnelave:
I'm letting the checks run, but I think you'll need to run `cargo fmt`.

Michsior14:
> I'm letting the checks run, but I think you'll need to run `cargo fmt`.

It should be already formatted correctly.

Michsior14:
@nitnelave What about the clippy? Should I fix it in this PR or you prefer another one?

nitnelave (approved):

nitnelave:
Thanks for the PR! Merged.

You can open another PR for the clippy fixes, if you want.

PR #381, opened 2022-11-24
server: prefer immutable path

wheinze:
Align path building on advice in [docs](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#examples)


nitnelave (approved):
Nice!

PR #383, opened 2022-11-25
docker: update Rust to v1.65.0

wheinze:
Get ready for #382 and update Rust in docker to fix (https://github.com/nitnelave/lldap/actions/runs/3544340544/jobs/5951515771)

### DIFF
--- a/.github/workflows/Dockerfile.dev
+++ b/.github/workflows/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM rust:1.62-slim-bullseye
+FROM rust:1.65-slim-bullseye
 
 # Set needed env path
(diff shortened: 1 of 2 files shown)


nitnelave (approved):

nitnelave:
It seems you may have missed aarch-64? It's still complaining about 1.62.

Thanks for the PR, btw, I hadn't even noticed yet that the build wasn't passing :D

wheinze:
> It seems you may have missed aarch-64? It's still complaining about 1.62.
> 
> Thanks for the PR, btw, I hadn't even noticed yet that the build wasn't passing :D

@nitnelave: Guess its caused by [`nitnelave/rust-dev:latest`](https://github.com/nitnelave/lldap/blob/main/.github/workflows/docker-build-static.yml#L209) not yet being updated to v1.65.0 ;) 

Is the according Dockerfile available publicly so I could adjust it?

nitnelave:
I think that's just https://github.com/nitnelave/lldap/blob/main/.github/workflows/Dockerfile.dev

I'll have to manually rebuild and push it, though.

PR #388, opened 2022-12-03
typo corrected

Hobbabobba:


nitnelave:
The attribute name is case insensitive, and here they're all written in lowercase. I don't think we should update that, or then we'd need to make everything camelcase

PR #390, opened 2022-12-06
Update vaultwarden.md

lordratner:
I think this is a typo, but I could be wrong. It's throwing an error for me.


nitnelave:
This setting should be fine, it's better to use `uid`. What kind of error are you getting?

lordratner:
Ok it must have been something else on my side, sorry!

PR #391, opened 2022-12-06
[WIP] displayName as struct to faciliate optional display name field. 

Moovlin:
Opening this as a WIP PR to get a review on the code I've written to make sure it meets the standards of the project as well as get some feedback on the implementation. At this point, i created a new struct which is near identical to the userId struct and then converted all `Optional(String)`s to `DisplayName(String)`. 

Testing:
LLDAP launches and I've run `cargo test` with one test failing. This may be a test that is failing simply because of the change made or it may require more work.

### DIFF
--- a/app/src/components/user_details_form.rs
+++ b/app/src/components/user_details_form.rs
@@ -355,7 +355,7 @@ impl UserDetailsForm {
         let mut user_input = update_user::UpdateUserInput {
             id: self.common.user.id.clone(),
             email: None,
-            displayName: None,
+            displayName: self.common.user.id.clone(),
             firstName: None,
             lastName: None,
             avatar: None,
@@ -367,7 +367,7 @@ impl UserDetailsForm {
             user_input.email = Some(email);
         }
         if base_user.display_name != model.display_name {
-            user_input.displayName = Some(model.display_name);
+            user_input.displayName = model.display_name;
         }
         if base_user.first_name != model.first_name {
             user_input.firstName = Some(model.first_name);
--- a/schema.graphql
+++ b/schema.graphql
@@ -57,7 +57,7 @@ type Query {
 input CreateUserInput {
   id: String!
   email: String!
-  displayName: String
+  displayName: String!
   firstName: String
   lastName: String
   avatar: String
@@ -84,7 +84,7 @@ type Success {
 input UpdateUserInput {
   id: String!
   email: String
-  displayName: String
+  displayName: String!
   firstName: String
   lastName: String
   avatar: String
--- a/server/src/domain/handler.rs
+++ b/server/src/domain/handler.rs
@@ -1,7 +1,8 @@
 use super::{
     error::Result,
     types::{
-        Group, GroupDetails, GroupId, JpegPhoto, User, UserAndGroups, UserColumn, UserId, Uuid,
+        DisplayName, Group, GroupDetails, GroupId, JpegPhoto, User, UserAndGroups, UserColumn,
+        UserId, Uuid,
     },
 };
 use async_trait::async_trait;
@@ -44,7 +45,7 @@ pub struct CreateUserRequest {
     // Same fields as User, but no creation_date, and with password.
     pub user_id: UserId,
     pub email: String,
-    pub display_name: Option<String>,
+    pub display_name: DisplayName,
     pub first_name: Option<String>,
     pub last_name: Option<String>,
     pub avatar: Option<JpegPhoto>,
@@ -55,7 +56,7 @@ pub struct UpdateUserRequest {
     // Same fields as CreateUserRequest, but no with an extra layer of Option.
     pub user_id: UserId,
     pub email: Option<String>,
-    pub display_name: Option<String>,
+    pub display_name: DisplayName,
     pub first_name: Option<String>,
     pub last_name: Option<String>,
     pub avatar: Option<JpegPhoto>,
--- a/server/src/domain/types.rs
+++ b/server/src/domain/types.rs
@@ -167,6 +167,82 @@ impl ValueType for UserId {
     }
 }
 
+#[derive(PartialEq, Eq, Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(from = "String")]
+pub struct DisplayName(String);
+
+impl DisplayName {
+    pub fn new(display_name: &str) -> Self {
+        Self(display_name.to_string())
+    }
+
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl std::fmt::Display for DisplayName {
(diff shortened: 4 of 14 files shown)


app/src/components/user_details_form.rs
```diff
@@ -355,7 +355,7 @@ impl UserDetailsForm {
         let mut user_input = update_user::UpdateUserInput {
             id: self.common.user.id.clone(),
             email: None,
-            displayName: None,
+            displayName: self.common.user.id.clone(),
```
nitnelave, line 358: No, that stays a `None`

schema.graphql
```diff
@@ -57,7 +57,7 @@ type Query {
 input CreateUserInput {
   id: String!
   email: String!
-  displayName: String
+  displayName: String!
```
nitnelave, line 60: This shouldn't change. I'll comment in the Rust code where I disagree on the change.

server/src/domain/types.rs
```diff
@@ -167,6 +167,82 @@ impl ValueType for UserId {
...
+
+impl std::fmt::Display for DisplayName {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl From<String> for DisplayName {
+    fn from(s: String) -> Self {
+        Self::new(&s)
```
nitnelave, line 196: You're making an extra copy here: since you have a `String`, you can just write `Self(s)`

server/src/domain/types.rs
```diff
@@ -167,6 +167,82 @@ impl ValueType for UserId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl std::fmt::Display for DisplayName {
```
nitnelave, line 188: Do we need display?

server/src/domain/types.rs
```diff
@@ -167,6 +167,82 @@ impl ValueType for UserId {
...
+
+impl From<&DisplayName> for Value {
+    fn from(display_name: &DisplayName) -> Self {
+        display_name.as_str().into()
+    }
+}
+
+impl TryGetable for DisplayName {
+    fn try_get(res: &QueryResult, pre: &str, col: &str) -> Result<Self, TryGetError> {
+        Ok(DisplayName::new(&String::try_get(res, pre, col)?))
```
nitnelave, line 214: Same here, you're making an extra copy. Since you have a `From<String>`, you can write
`Ok(DisplayName::from(String::try_get(res, pre, col)?))`

server/src/domain/types.rs
```diff
@@ -167,6 +167,82 @@ impl ValueType for UserId {
...
+    fn try_from_u64(_n: u64) -> Result<Self, DbErr> {
+        Err(DbErr::ConvertFromU64(
+            "DisplayName cannot be constructed from u64",
+        ))
+    }
+}
+
+impl ValueType for DisplayName {
+    fn try_from(v: Value) -> Result<Self, ValueTypeErr> {
+        Ok(DisplayName::new(
```
nitnelave, line 228: Same thing here, use `::from` with the `String`

server/src/domain/handler.rs
```diff
@@ -44,7 +45,7 @@ pub struct CreateUserRequest {
     // Same fields as User, but no creation_date, and with password.
     pub user_id: UserId,
     pub email: String,
-    pub display_name: Option<String>,
```
nitnelave, line 47: We want to keep the change inside this interface: it should only be the communication with the DB that changes. We still get a `Option<String>`, but then we convert it into a `DisplayName`

server/src/domain/types.rs
```diff
@@ -167,6 +167,82 @@ impl ValueType for UserId {
...
+#[derive(PartialEq, Eq, Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(from = "String")]
+pub struct DisplayName(String);
+
+impl DisplayName {
+    pub fn new(display_name: &str) -> Self {
+        Self(display_name.to_string())
+    }
+
+    pub fn as_str(&self) -> &str {
```
nitnelave, line 179: I think instead of those two, we want conversion functions from `DisplayName` to `Option<&str>` and `Option<String>`. We want to translate from `""` to `None`

nitnelave: Similarly, you might need `From<Option<String>>`

server/src/domain/types.rs
```diff
@@ -304,7 +380,7 @@ impl IntoActiveValue<JpegPhoto> for JpegPhoto {
 pub struct User {
     pub user_id: UserId,
     pub email: String,
-    pub display_name: Option<String>,
```
nitnelave, line 307: Let's keep it as `Option<String>`

server/src/infra/auth_service.rs
```diff
@@ -171,9 +171,7 @@ where
         Some(token) => token,
     };
     if let Err(e) = super::mail::send_password_reset_email(
-        user.display_name
```
nitnelave, line 174: If we limit the change to the DB interface, this shouldn't have to change.

server/src/infra/graphql/mutation.rs
```diff
@@ -27,7 +27,7 @@ impl<Handler: BackendHandler> Mutation<Handler> {
 pub struct CreateUserInput {
     id: String,
     email: String,
-    display_name: Option<String>,
```
nitnelave, line 30: Same here, this file shouldn't change.

server/src/infra/ldap_handler.rs
```diff
@@ -460,6 +460,8 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
             &self.ldap_info.base_dn,
             &self.ldap_info.base_dn_str,
         )?;
+
+        //
```
nitnelave, line 464: Remove comment.

nitnelave (changes requested):
Good wrapper implementation for `DisplayName`, but it needs conversions back and forth from `Option<String>`: we want to have the code interface be `Option<String>`, but all communication with the DB should be via `DisplayName`, to turn a `None` into an empty string.

What's missing in this change is also the update to sql_migrations.rs: we need a bit of SQL to alter (if possible) the table to change the type of the column, make it not accept null, after changing all the nulls to empty strings.

nitnelave:
Sorry, it took me a while to get to review this code, life is busy these days :)

nitnelave:
Hey @Moovlin , what's the status here? Do you want to keep working on this, or should I take over?

Moovlin:
@nitnelave Life has unfortunately really gotten in the way. In the interest of making progress, you should take over.

nitnelave:
No worries, it happens :)

I'll take this PR over, but feel free to come back and grab an issue if you ever feel like contributing again!

nitnelave:
Alright, this PR can be closed, I implemented it on master.

PR #396, opened 2022-12-14
Small fix to enable mysql schema creation

naturalethic:
These two small changes will fix this to work with mysql.  I do not know if it breaks sqlite.


nitnelave (approved):
The changes look good, but I'd like to test them with all of sqlite, mysql and postgresql before merging.

nitnelave:
Closing in favor of https://github.com/nitnelave/lldap/pull/418

PR #402, opened 2022-12-31
Add devcontainer

### LINKED ISSUE #331: Add vscode dev-container
It would be much easier to contribute to the project if there is a VSCode Dev-Container definition in this project. This allows working on this project without installing any tools (beside vscode and docker).

pixelrazor:
Fixes #331 
The dockerfile could probably be optimized - i just tinkered until i got something that wants to work out the box. the cargo install step is pretty hefty, so i COULD try to tinker with making that a separate layer as early as possible to help cache it, but i don't think it's too important given this shouldn't change often at all.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,10 @@
+FROM rust:1.66
+
+RUN apt update && \
+    curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - && \
+    apt update && \
+    apt install -y --no-install-recommends nodejs libssl-dev musl-dev make perl curl 
+
+RUN RUSTFLAGS=-Ctarget-feature=-crt-static cargo install wasm-pack cargo-chef \
+    && npm install -g rollup \
+    && rustup target add wasm32-unknown-unknown
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,10 +1,26 @@
 FROM rust:1.66
 
+ARG USERNAME=lldapdev
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Create the user
+RUN groupadd --gid $USER_GID $USERNAME \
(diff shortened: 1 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM rust:1.66
+
+RUN apt update && \
+    curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - && \
+    apt update && \
+    apt install -y --no-install-recommends nodejs libssl-dev musl-dev make perl curl 
+
+RUN RUSTFLAGS=-Ctarget-feature=-crt-static cargo install wasm-pack cargo-chef \
```
nitnelave, line 8: Either you don't need cargo-chef, or (better) you can steal the dependency pre-compilation step from this dockerfile: https://github.com/nitnelave/lldap/blob/main/Dockerfile#L25-L41

pixelrazor: I'll opt to remove chef. The container itself doesn't contain the code - vscode bind mounts it after creation

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+    && npm install -g rollup \
+    && rustup target add wasm32-unknown-unknown
```
pixelrazor, line 10: I'm going to add auser in the container - vscode will amp that user's UID and GID to the current user and ffix the file permission issues

nitnelave:
Thanks, that looks better! I'm no expert on VS code containers, so I'll trust you that it works.

Last nit: can you add `.devcontainer` to the `.dockerignore` file?

pixelrazor:
> Thanks, that looks better! I'm no expert on VS code containers, so I'll trust you that it works.
> 
> Last nit: can you add `.devcontainer` to the `.dockerignore` file?

Done. I added `.vscode` to gitignore as well - that wolds workspace specific configuration for vscode

PR #407, opened 2023-01-04
Initial implementation of avatar component

### LINKED ISSUE #356: Show avatar in navigation if set
If the logged-in user has an avatar it would be nice if it could show up on the web interface across all pages. Makes the experience slightly more personalized! :)

pixelrazor:
Fixes #356


app/src/components/avatar.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "../schema.graphql",
+    query_path = "queries/get_user_avatar.graphql",
+    response_derives = "Debug, Hash, PartialEq, Eq, Clone",
+    custom_scalars_module = "crate::infra::graphql"
+)]
+pub struct GetUserAvatar;
+
+pub type User = get_user_avatar::GetUserAvatarUser;
```
nitnelave, line 15: I don't think you use this type? You can probably remove it.

app/src/components/avatar.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    fn mut_common(&mut self) -> &mut CommonComponentParts<Self> {
+        &mut self.common
+    }
+}
+
+impl Avatar {
+    fn get_user_avatar(&mut self) {
+        if self.common.username.len() > 0 {
```
nitnelave, line 61: Don't use empty strings to signify "no user", but instead use an `Option<String>`

nitnelave: Or just rely on the fact that you'll always be called with a valid username by construction.

app/src/components/app.rs
```diff
@@ -247,15 +249,13 @@ impl App {
...
-                      fill="currentColor"
-                      class="bi bi-person-circle"
-                      viewBox="0 0 16 16">
-                      <path d="M11 6a3 3 0 1 1-6 0 3 3 0 0 1 6 0z"/>
-                      <path fill-rule="evenodd" d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8zm8-7a7 7 0 0 0-5.468 11.37C3.242 11.226 4.805 10 8 10s4.757 1.225 5.468 2.37A7 7 0 0 0 8 1z"/>
-                    </svg>
+                    {
+                      if let Some((user_id, _)) = &self.user_info {
+                        html! {
+                          <Avatar username=user_id.clone() width=32 height=32 />
```
nitnelave, line 255: ~You want to be able to send a message to the `Avatar` component to reload when the avatar changes. As it turns out, this is rather tricky. One solution would be with: https://dev.to/davidcks/how-to-send-a-message-to-a-child-component-in-yew-503~

~Then once you have the component link of the avatar child, you can send a message to the child. Not ideal, but that's the best we can have, I think.~

Actually, now that I think about it, we can just change the properties of Avatar as a message, since it just needs to reload. By setting a property like "last avatar changed time", the Avatar child can re-render (and re-fetch) when it receives this. Much simpler.

The other part of the story is to pass a callback that will send a message to the App with the last avatar changed time. You pass that callback down into UserDetails/UserDetailsForm components, and in the form, if the avatar changes, you call the callback.

nitnelave (changes requested):

app/src/components/avatar.rs
pixelrazor: Option didn't seem to want to play nicely with Props; I think i'll just opt for the latter

app/src/components/app.rs
nitnelave: Actually, the proper solution for this is to use agents: https://yew.rs/docs/0.18.0/concepts/agents

pixelrazor: Thanks for this! I'll try to make the time soon for using that

nitnelave: I'm working on an update of the frontend to yew 0.19 or 0.20, so you might as well wait until that's done, since that'll change a bunch of things.

pixelrazor: Should this still become an agent, or is this something that would work well with yew 0.19 context?

nitnelave: I think it should be an agent. But I've never used them :)

nitnelave: Actually, we can also have a look at the contexts: https://yew.rs/docs/next/concepts/contexts

pixelrazor: Used agents after all, a little smoother when i went down both rabbit holes

app/src/components/avatar.rs
```diff
@@ -0,0 +1,114 @@
...
+    response_derives = "Debug",
+    custom_scalars_module = "crate::infra::graphql"
+)]
+pub struct GetUserAvatar;
+
+pub struct Avatar {
+    common: CommonComponentParts<Self>,
+    /// The user info. If none, the error is in `error`. If `error` is None, then we haven't
+    /// received the server response yet.
+    avatar: Option<String>,
```
nitnelave, line 22: Can we wrap that in a new type? `Option<AvatarData>` with `pub struct AvatarData(String);`

app/src/components/avatar.rs
```diff
@@ -0,0 +1,114 @@
...
+    schema_path = "../schema.graphql",
+    query_path = "queries/get_user_avatar.graphql",
+    response_derives = "Debug",
+    custom_scalars_module = "crate::infra::graphql"
+)]
+pub struct GetUserAvatar;
+
+pub struct Avatar {
+    common: CommonComponentParts<Self>,
+    /// The user info. If none, the error is in `error`. If `error` is None, then we haven't
```
nitnelave, line 20: What if the user doesn't have an avatar?

app/src/components/avatar.rs
```diff
@@ -0,0 +1,114 @@
...
+            Msg::UserAvatarResponse(response) => match response {
+                Ok(user) => self.avatar = user.user.avatar,
+                Err(e) => {
+                    self.avatar = None;
+                    bail!("Error getting user avatar: {}", e);
+                }
+            },
+            Msg::Update => self.get_user_avatar(ctx),
+        }
+        Ok(true)
```
nitnelave, line 53: Technically this will cause an unnecessary redraw even if the avatar was None (not received yet) and we got None (user doesn't have an avatar).

app/src/components/user_details_form.rs
```diff
@@ -402,6 +406,7 @@ impl UserDetailsForm {
             self.user.avatar = Some(avatar);
         }
         self.just_updated = true;
+        self.avatar_event_bus.send(Request::Update);
```
nitnelave, line 409: Why not send the avatar directly in the bus? And we check whether the avatar has changed (before setting `self.user.avatar`).

app/src/components/app.rs
```diff
@@ -306,15 +307,7 @@ impl App {
...
-                  <svg xmlns="http://www.w3.org/2000/svg"
-                    width="32"
-                    height="32"
-                    fill="currentColor"
-                    class="bi bi-person-circle"
-                    viewBox="0 0 16 16">
-                    <path d="M11 6a3 3 0 1 1-6 0 3 3 0 0 1 6 0z"/>
-                    <path fill-rule="evenodd" d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8zm8-7a7 7 0 0 0-5.468 11.37C3.242 11.226 4.805 10 8 10s4.757 1.225 5.468 2.37A7 7 0 0 0 8 1z"/>
-                  </svg>
+                  <Avatar username={user_id.clone()} width=32 height=32 />
```
nitnelave, line 310: Should we call it `ShowAvatar` maybe?

nitnelave (changes requested):
Nice! A couple of changes

app/src/components/avatar.rs
pixelrazor: this is a copy that i think was copied from another component, or is so old i forrgot i even wrote it

app/src/components/avatar.rs
pixelrazor: I'll see about adding logic to check if the avatar has changed or not

pixelrazor: So i want to be cautious of futuure work for allowing avatars to be deleted. I did some smrter decisions around redrawing though

app/src/components/avatar.rs
```diff
@@ -0,0 +1,133 @@
...
+    avatar: Option<AvatarData>,
+    _producer: Box<dyn Bridge<AvatarEventBus>>,
+}
+
+/// State machine describing the possible transitions of the component state.
+/// It starts out by fetching the user's details from the backend when loading.
+pub enum Msg {
+    /// Received the user details response, either the user data or an error.
+    UserAvatarResponse(Result<get_user_avatar::ResponseData>),
+    Update(Response),
```
nitnelave, line 40: Maybe AvatarChanged?

app/src/components/avatar.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub width: i32,
+    pub height: i32,
+}
+
+impl CommonComponent<ShowAvatar> for ShowAvatar {
+    fn handle_msg(
+        &mut self,
+        ctx: &Context<Self>,
+        msg: <Self as Component>::Message,
+    ) -> Result<bool> {
```
nitnelave, line 55: How about `mut avatar_changed = false;` and when you update the avatar to a different value than is already stored, you set it to true?

That way you don't have multiple happy path returns, you can keep a single return at the end.

app/src/components/avatar.rs
```diff
@@ -0,0 +1,133 @@
...
+        match msg {
+            Msg::UserAvatarResponse(response) => match response {
```
nitnelave, line 57: Style nit to reduce nesting:
`let avatar = response.context("while fetching the avatar")?.user.avatar;`

app/src/components/avatar.rs
```diff
@@ -0,0 +1,133 @@
...
+        };
+        avatar.get_user_avatar(ctx);
+        avatar
+    }
+
+    fn update(&mut self, ctx: &Context<Self>, msg: Self::Message) -> bool {
+        CommonComponentParts::<Self>::update(self, ctx, msg)
+    }
+
+    fn view(&self, ctx: &Context<Self>) -> Html {
```
nitnelave, line 111: This looks like a good candidate for the suspense API by yew: show a placeholder until the data is loaded. Maybe it would simplify things?

app/src/components/avatar.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(true)
+    }
+
+    fn mut_common(&mut self) -> &mut CommonComponentParts<Self> {
+        &mut self.common
+    }
+}
+
+impl ShowAvatar {
+    fn get_user_avatar(&mut self, ctx: &Context<Self>) {
```
nitnelave, line 81: We could move that to the avatar agent.

app/src/components/avatar_event_bus.rs
```diff
@@ -0,0 +1,55 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use yew_agent::{Agent, AgentLink, Context, HandlerId};
+
+use super::avatar::AvatarData;
```
nitnelave, line 5: Can you make that an absolute import? `crate::infra::avatar...`
Although the type AvatarData should probably be defined here rather than in the component.

app/src/components/avatar_event_bus.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Request {
```
nitnelave, line 8: No need to make them enums, no?

app/src/components/avatar_event_bus.rs
```diff
@@ -0,0 +1,55 @@
...
+pub enum Response {
+    Update((String, Option<AvatarData>)),
+}
+
+pub struct AvatarEventBus {
+    link: AgentLink<AvatarEventBus>,
+    subscribers: HashSet<HandlerId>,
+}
+
+impl Agent for AvatarEventBus {
```
nitnelave, line 22: Maybe we can keep a map of subscribers per username? So we only send the update to the right component, based on what they requested.

nitnelave: It would have to be a double map:
UserId to Vec<HandlerId>, and HandlerId to Vec<UserId> for cleanup.
Maybe it's overkill? WDYT ?

app/src/components/avatar_event_bus.rs
```diff
@@ -0,0 +1,55 @@
+use serde::{Deserialize, Serialize};
```
nitnelave, line 1: Sorry, just noticed: the avatar event bus should be in the infra folder, not component

app/src/components/avatar_event_bus.rs
nitnelave: Actually the request can be an enum: one to request async fetching, and one to just notify of an update.

nitnelave (changes requested):

nitnelave:
Sorry for the many comments, it's my first time playing with agents, so the final design wasn't clear in my head. I think moving most of the fetching and state logic in the agent makes sense, and will help for the next change (table of avatars)

pixelrazor:
> Sorry for the many comments, it's my first time playing with agents, so the final design wasn't clear in my head. I think moving most of the fetching and state logic in the agent makes sense, and will help for the next change (table of avatars)

I think i'll just migrate this to agent 0.2.0 Workers  with that and uuse suspense. Great thinking!

One gripe is a lot of newer yew stuff feels a lot more geared toward functional components, and maybe it's just me, but i'm struggling to find a good way to work them in (especially given the CommonComponent functionality on struct components). I'll still work on my approach to them, but i'm just sharing out loud in case it sparks and ideas or discussion

nitnelave:
I'm willing to move off of common component. It was much more important before 0.19 since a lot of functionality was baked in there, but now it's mostly just a glorified way to handle errors and whether a fetch is running.
Suspense is a better way to handle the latter, and the former can be done is a more elegant way with functional components, I think.

But yeah, migrating from struct components to a hooks-based functional components approach will require a bit of gymnastics. I haven't yet tried to rewrite a more complex component with hooks, I think there are some useful patterns to discover.

nitnelave (changes requested):

PR #409, opened 2023-01-05
fetch missing web components

martadinata666:
add command to fetch the libraries and fonts that missing from on zipped web release.


nitnelave:
Thanks!

PR #411, opened 2023-01-06
simplify artifacts download and some other things

martadinata666:
* Use nitnelave/rust-dev:latest for all job
* simplify artifacts download, no more per arch downloading
* add `rustup target add wasm32-unknown-unknown` on ui build, will this affect anything?
* update actions/cache version
* modify release to all in one compressed

### DIFF
--- a/.github/workflows/docker-build-static.yml
+++ b/.github/workflows/docker-build-static.yml
@@ -240,8 +240,8 @@ jobs:
         uses: actions/checkout@v3.3.0
       - name: Download All Artifacts
         uses: actions/download-artifact@v3
-          with:
-            path: bin
+        with:
+          path: bin
 
       - name: Download llap ui artifacts
         uses: actions/download-artifact@v3
--- a/.github/workflows/docker-build-static.yml
+++ b/.github/workflows/docker-build-static.yml
@@ -392,8 +392,32 @@ jobs:
              for file in $(cat app/static/libraries.txt); do wget -P app/static "$file"; done
              for file in $(cat app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done
              chmod a+r -R .
-      - name: compress web
-        run: sudo apt update && sudo apt install -y zip && zip -r web.zip app/
+
+      - name: Setup LLDAP dir for packing
+        run: |
+             mkdir aarch64-lldap
+             mkdir amd64-lldap
+             mkdir armhf-lldap
+             mv bin/aarch64-lldap aarch64-lldap/lldap
+             mv bin/amd64-lldap amd64-lldap/lldap
+             mv bin/armhf-lldap armhf-lldap/lldap
+             mv bin/aarch64-migration-tool aarch64-lldap/migration-tool
+             mv bin/amd64-migration-tool amd64-lldap/migration-tool
+             mv bin/armhf-migration-tool armhf-lldap/migration-tool
+             cp -r app aarch64-lldap/
+             cp -r app amd64-lldap/
+             cp -r app armhf-lldap/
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-build-static.yml
```diff
@@ -100,7 +89,8 @@ jobs:
   build-armhf:
     runs-on: ubuntu-latest
     container:
-      image: rust:1.65
+#      image: rust:1.65
```
nitnelave, line 92: Remove the comment.

.github/workflows/docker-build-static.yml
```diff
@@ -481,12 +401,12 @@ jobs:
         id: create_release
         with:
           allowUpdates: true
-          artifacts:  bin/armhf-bin/lldap-armhf,
-                      bin/aarch64-bin/lldap-aarch64,
-                      bin/amd64-bin/lldap-amd64,
-                      bin/armhf-bin/migration-tool-armhf,
-                      bin/aarch64-bin/migration-tool-aarch64,
-                      bin/amd64-bin/migration-tool-amd64,
+          artifacts:  bin/aarch64-lldap,
```
nitnelave, line 404: Should we maybe wrap them in a tar with the web assets as well?

Anyone who downloads the binary needs the web assets as well, and people who need just the web assets can then replace the binary with their own.

.github/workflows/docker-build-static.yml
```diff
@@ -392,21 +392,41 @@ jobs:
...
+             cp -r app armhf-lldap/
+             ls -alR aarch64-lldap/
+             ls -alR amd64-lldap/
+             ls -alR armhf-lldap/
+
+      - name: Compress
+        run: |
+             sudo apt update
+             sudo apt install -y zip
+             zip -r aarch64-lldap.zip aarch64-lldap/
```
nitnelave, line 418: tar?

nitnelave: It's more standard for linux packages. zip for windows why not, but we don't have a windows binary.

nitnelave: https://xkcd.com/1168/

.github/workflows/docker-build-static.yml
```diff
@@ -413,20 +414,18 @@ jobs:
 
       - name: Compress
         run: |
-             sudo apt update
-             sudo apt install -y zip
-             zip -r aarch64-lldap.zip aarch64-lldap/
-             zip -r amd64-lldap.zip amd64-lldap/
-             zip -r armhf-lldap.zip armhf-lldap/
+             tar -czvf -r aarch64-lldap.tar.gz aarch64-lldap/
+             tar -czvf amd64-lldap.tar.gz amd64-lldap/
```
nitnelave, line 418: I'm confused: do we need the `-r` or not?

nitnelave: Yes, I think we want it for all.

nitnelave: ah no you're creating a file that doesn't exist. Carry on! :)

martadinata666:
The last part is verify docker image, it should be work, the artifacts path sometimes tricky 🤞🏼

nitnelave:
> Release should be like this now https://github.com/martadinata666/lldap/releases/tag/v0.4.2-alpha8

The artifacts structure looks good!

Thanks a lot for your help with docker, this is not something that I enjoy :D

nitnelave:
Just checking, are you checking the docker image?

martadinata666:
No, not really, from the actions log, the structure should be fine. as far as I can tell. it had everything in `/app`

PR #414, opened 2023-01-09
Gitea: add additional attributes and group sync

poVoq:
Not extensively tested, but group/team sync seems to work in Forgejo.

### DIFF
--- a/example_configs/gitea.md
+++ b/example_configs/gitea.md
@@ -1,4 +1,4 @@
-# Configuration for Gitea
+# Configuration for Gitea (& Forgejo)
 In Gitea, go to `Site Administration > Authentication Sources` and click `Add Authentication Source`
 Select `LDAP (via BindDN)`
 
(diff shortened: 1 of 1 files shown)


example_configs/gitea.md
```diff
@@ -14,9 +14,30 @@ To log in they can either use their email address or user name. If you only want
...
+* Group Search Base DN: `ou=groups,dc=example,dc=com`
+* Group Attribute Containing List Of Users: `member`
+* User Attribute Listed In Group: `dn`
+* Map LDAP groups to Organization teams: `{"cn=Groupname1,ou=groups,dc=example,dc=com":{"Organization1": ["Teamname"]},"cn=Groupname2,ou=groups,dc=example,dc=com": {"Organization2": ["Teamname1", "Teamname2"]}}`
+
+Check `Remove Users from syncronised teams...`
+
+The `Map LDAP groups to Organization teams` config is JSON formatted and can be extended to as many groups as needed.
+
+Replace every instance of `dc=example,dc=com` with your configured domain.
```
nitnelave, line 43: That's still there on line 23, I think you don't need to repeat it.

nitnelave (approved):

example_configs/gitea.md
poVoq: Yes I just thought to make double sure people would not forget and I didn't want to remove it further up.

PR #416, opened 2023-01-12
Add dn filter support

lucat1:
Adds support to filter results by the dn name of the resource.

You can try it on the default configuration with:
```
$ ldapsearch -b "dc=example,dc=com" -D "cn=admin,ou=people,dc=example,dc=com" -W -H ldap://localhost:3890 "(dn=uid=admin,ou=people,dc=example,dc=com)"
$ ldapsearch -b "dc=example,dc=com" -D "cn=admin,ou=people,dc=example,dc=com" -W -H ldap://localhost:3890 "(dn=uid=lldap_strict_readonly,ou=groups,dc=example,dc=com)" 
```


server/src/domain/ldap/group.rs
```diff
@@ -219,3 +233,48 @@ pub async fn get_groups_list<Backend: BackendHandler>(
...
+
+    use crate::domain::ldap::group::GroupRequestFilter;
+    use crate::domain::ldap::utils::{parse_distinguished_name, LdapInfo};
+
+    use super::convert_group_filter;
+
+    static BASE_DN_STR: &str = "dc=example,dc=com";
+
+    #[test]
+    fn test_convert_group_filter() {
```
nitnelave, line 249: Oh, you put the tests there (which is where I told you to put the tests, actually...)

I think it'd better to add a test case to ldap_handler.rs where we test at a higher level the entire LDAP layer. Also all the setup is already done there.

Sorry for that!

nitnelave (changes requested):

nitnelave:
There's also a small nit from cargo clippy.

nitnelave (approved):

server/src/domain/ldap/group.rs
```diff
@@ -126,6 +127,19 @@ fn convert_group_filter(
                         vec![],
                     )))),
                 },
+                "dn" => Ok(
+                    match get_group_id_from_distinguished_name(
```
nitnelave, line 131: We _could_ write it as
```rust
Ok(
  get_group_id(...)
  .map(GroupRequestFilter::DisplayName)
  .unwrap_or_else(|_| {
    warn!(...);
    GroupRequestFilter::Not(...)
  }))
```
And it might even make sense to add a `false()` and `true()` static functions to `GroupRequestFilter`, but that's for a next PR. Let's keep it as is, I'll make a second pass later.

nitnelave (approved):

PR #417, opened 2023-01-12
Support LDAP compare

lucat1:
Added support for the ldap compare command. This is a requirement for all services based on django.

You can try the functionality on the default configuration with the following commands:
```
$ ldapcompare -D "cn=admin,ou=people,dc=example,dc=com" -W -H ldap://localhost:3890 "uid=admin,ou=people,dc=example,dc=com" "uid:admin"
$ ldapcompare -D "cn=admin,ou=people,dc=example,dc=com" -W -H ldap://localhost:3890 "uid=lldap_admin,ou=groups,dc=example,dc=com" "uid:lldap_admin"
```

Also comparing other fields besides the uid has been tested and appears to be working.


server/Cargo.toml
```diff
@@ -28,7 +28,7 @@ itertools = "0.10.1"
 juniper = "0.15.10"
 juniper_actix = "0.4.0"
 jwt = "0.13"
-ldap3_proto = "*"
+ldap3_proto = { git = "https://github.com/kanidm/ldap3", package = "ldap3_proto" }
```
nitnelave, line 31: I'd rather pin it to a specific commit hash than depend on master.

Also, you can do that with an override in the top-level Cargo.toml.

server/src/infra/ldap_handler.rs
```diff
@@ -527,6 +544,47 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
...
+        let res = self.do_search_or_dse(&req).await?;
+        let entry = res.first().ok_or(LdapError {
+            code: LdapResultCode::OperationsError,
+            message: "Search request returned nothing".to_string(),
+        })?;
+        match entry {
+            LdapOp::SearchResultEntry(entry) => {
+                let mut available = false;
+                for attr in entry.attributes.iter() {
+                    if attr.atype == request.atype && attr.vals.contains(&request.val) {
```
nitnelave, line 565: I'm afraid that comparing the `atype` will not work if the comparison attribute is not normalized (i.e. requesting `distinguishedName` or even `DN` when we return `dn`).

It's also missing the case of "attribute is not part of the schema". I think we should start by checking whether the entry is a user or a group, then map both the compare attribute and the entry attributes using `map_user_field` (or `map_group_field`) before checking for equality. That'll also give us the case of "attribute not supported".

server/src/infra/ldap_handler.rs
```diff
@@ -527,6 +544,47 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
...
+            &self.ldap_info.base_dn_str,
+            LdapFilter::And(vec![LdapFilter::Equality(
+                "dn".to_string(),
+                request.dn.to_string(),
+            )]),
+            vec![],
+        );
+        let res = self.do_search_or_dse(&req).await?;
+        let entry = res.first().ok_or(LdapError {
+            code: LdapResultCode::OperationsError,
```
nitnelave, line 558: Isn't that a NoSuchObject error?

server/src/infra/ldap_handler.rs
```diff
@@ -527,6 +544,47 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
...
+            message: "Search request returned nothing".to_string(),
+        })?;
```
nitnelave, line 560: It might also be nice to check whether there's more than 1 result.

server/src/infra/ldap_handler.rs
```diff
@@ -2060,6 +2105,79 @@ mod tests {
...
+                    display_name: Some("bôb böbberson".to_string()),
+                    first_name: Some("bôb".to_string()),
+                    last_name: Some("böbberson".to_string()),
+                    uuid: uuid!("698e1d5f-7a40-3151-8745-b9b8a37839da"),
+                    ..Default::default()
+                },
+                groups: None,
+            }])
+        });
+        mock.expect_list_groups().returning(|_| {
```
nitnelave, line 2125: I would do a different test for the group.

server/src/infra/ldap_handler.rs
```diff
@@ -2060,6 +2105,79 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_compare() {
+        let mut mock = MockTestBackendHandler::new();
+        mock.expect_list_users().returning(|_, _| {
+            Ok(vec![UserAndGroups {
+                user: User {
+                    user_id: UserId::new("bob"),
+                    email: "bob@bobmail.bob".to_string(),
+                    display_name: Some("bôb böbberson".to_string()),
```
nitnelave, line 2116: you can remove the display_name, first_name, last_name and uuid.

server/src/infra/ldap_handler.rs
```diff
@@ -2060,6 +2105,79 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_compare() {
```
nitnelave, line 2109: Add tests to cover `CompareFalse`, an invalid dn, an invalid attribute, and no such object?

nitnelave (changes requested):

server/src/infra/ldap_handler.rs
lucat1: We always expect to have at least one element of type `SearchResultDone` because of [this](https://github.com/nitnelave/lldap/blob/main/server/src/infra/ldap_handler.rs#L439), right?

server/src/infra/ldap_handler.rs
lucat1: I can do this. What shall be the error then?

lucat1: I also implemented this one with `OperationsError`. let me know what kind of error you prefer here.

server/src/infra/ldap_handler.rs
lucat1: You suggest also mapping the `atype`s on the entries, but aren't these already "normalized" as they're generated internally? I'm asking as this would save us some computation and make the code clearer. But I can totally do that if it's necessary.

server/src/infra/ldap_handler.rs
nitnelave: Yes, there is always a `SearchResultDone`, you're right.

server/src/infra/ldap_handler.rs
nitnelave: The `compare`'s `atype` is not normalized yet.

nitnelave:
The `DateTime` fields inside the server have all changed to `NaiveDateTime` (you probably noticed that).

Cargo.toml
```diff
@@ -10,8 +10,8 @@ default-members = ["server"]
 
 # Remove once https://github.com/kanidm/ldap3_proto/pull/8 is merged.
```
nitnelave, line 11: Update the comment to wait for a release.

server/src/infra/ldap_handler.rs
nitnelave: Maybe we should update the message to say "Missing SearchDone message in search response"?

server/src/infra/ldap_handler.rs
nitnelave: What you can do is run the request's atype through a "LDAP attribute type -> SQL column -> LDAP attribute type" cycle to normalize it.

server/src/infra/ldap_handler.rs
```diff
@@ -527,6 +544,54 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
...
+        if res.len() > 1 {
+            return Err(LdapError {
+                code: LdapResultCode::OperationsError,
+                message: "Search request returned nothing".to_string(),
+            });
+        }
+
+        match entry {
+            LdapOp::SearchResultEntry(entry) => {
+                let mut available = false;
```
nitnelave, line 570:
```rust
let available = entry.attributes.iter().any(|attr| attr.atype == ...);
```

nitnelave:
Hi! What's the status of this PR? Are you still working on it?

nitnelave:
I took it over in #445

lucat1:
Thank you!

PR #418, opened 2023-01-13
Improve support for MySQL/Postgres

### LINKED ISSUE #87: Support for MySQL/PostgreSQL
Hi, first of all, thanks for the project!

And I just want to clarify - this isn't an _immediate_ feature request, but rather me asking would this be something you'd be okay putting on the roadmap?

Right now, LLDAP apparently stores everything in `/data` directory. But using SQL databases to back all of LLDAP's data could be _so_ versatile:

- If you want to stick with `/data` directory, you can! Just use SQLite as the backing store
- You can directly use any number of MySQL/Postgres databases that users _already have_ in order to store your, well, everything!

And I believe most of the SQL queries should be portable as well (with the only exception being DDLs). Thoughts?

nitnelave:
With these 2 commits, the integration with mysql/postgres is working.

It was only tested by hand, I didn't set up an automated test.

It's completely backwards-compatible with SQlite.

Fixes #87

### DIFF
--- a/server/src/domain/handler.rs
+++ b/server/src/domain/handler.rs
@@ -140,8 +140,14 @@ mod tests {
     fn test_uuid_time() {
         use chrono::prelude::*;
         let user_id = "bob";
-        let date1 = Utc.with_ymd_and_hms(2014, 7, 8, 9, 10, 11).unwrap();
-        let date2 = Utc.with_ymd_and_hms(2014, 7, 8, 9, 10, 12).unwrap();
+        let date1 = Utc
+            .with_ymd_and_hms(2014, 7, 8, 9, 10, 11)
+            .unwrap()
+            .naive_utc();
+        let date2 = Utc
+            .with_ymd_and_hms(2014, 7, 8, 9, 10, 12)
+            .unwrap()
+            .naive_utc();
         assert_ne!(
             Uuid::from_name_and_date(user_id, &date1),
             Uuid::from_name_and_date(user_id, &date2)
--- a/server/src/domain/ldap/user.rs
+++ b/server/src/domain/ldap/user.rs
@@ -1,3 +1,4 @@
+use chrono::TimeZone;
 use ldap3_proto::{
     proto::LdapOp, LdapFilter, LdapPartialAttribute, LdapResultCode, LdapSearchResultEntry,
 };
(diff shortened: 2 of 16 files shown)


ajgon:
Still no luck :/ (Using latest image from docker).

Screenshot 2023-01-13 at 17 35 03

```
2023-01-13T16:32:41.611143741+00:00 INFO     ┝━ ｉ [info]:  | uri: /api/graphql
2023-01-13T16:32:41.611173448+00:00 DEBUG    ┝━ check_if_token_is_valid [ 41.9µs | 2.73% ]
2023-01-13T16:32:41.611214471+00:00 DEBUG    │  ┕━ 🐛 [debug]:  | return: ValidationResults { user: UserId("admin"), permission: Admin }
2023-01-13T16:32:41.611542321+00:00 DEBUG    ┝━ [GraphQL mutation] create_user [ 1.02ms | 1.87% / 66.65% ]
2023-01-13T16:32:41.611546413+00:00 DEBUG    │  ┝━ 🐛 [debug]:  | user.id: "ajgon"
2023-01-13T16:32:41.611550408+00:00 DEBUG    │  ┝━ create_user [ 468µs | 30.50% ]
2023-01-13T16:32:41.611553946+00:00 DEBUG    │  │  ┕━ 🐛 [debug]:  | user_id: UserId("ajgon")
2023-01-13T16:32:41.616227317+00:00 DEBUG    │  ┕━ get_user_details [ 526µs | 34.28% ]
2023-01-13T16:32:41.616238656+00:00 DEBUG    │     ┝━ 🐛 [debug]:  | user_id: UserId("ajgon")
2023-01-13T16:32:41.624743773+00:00 DEBUG    │     ┕━ 🐛 [debug]:  | return: Err(EntityNotFound("ajgon"))
2023-01-13T16:32:41.624844275+00:00 INFO     ┕━ ｉ [info]:  | status_code: 200
```

I see it creates user in database, but without password...

nitnelave:
What database are you using?

ajgon:
Sorry, that's the most important part after all :D - PostgreSQL.

I think I found the issue. Originally I created user without display name, and got error, that display name is null - which means, it's not validated on UI/backend side, but on database side. Yet user was created in DB. And since then I couldn't recereate him (despite deleting from DB). After DB wipe and starting from scratch (with display name provided), everything worked.

So the solution will be either removing `NOT NULL` from `display_name` property, or implement proper validation.

ajgon:
Perfect, aside from this minor issue, I can confirm that Postgres is working. Thank you! This would make k8s deployments way easier :)

atropos112:
I keep getting `u8 unsupported by sqlx-postgres` when trying to connect to a fresh postgresql bitnami imaged container is this something up with my setup or something being looked into?

martadinata666:
> I keep getting `u8 unsupported by sqlx-postgres` when trying to connect to a fresh postgresql bitnami imaged container is this something up with my setup or something being looked into?

did you pull the image again? as the latest build should got this fixed.

atropos112:
turns out I had the default .Values.image.pullPolicy on the helm chart and same the image was set under same version number so my cluster didn't pull new one on restart and used the cached one :face_exhaling: 

On a positive side we are all HA baby woop woop

PR #420, opened 2023-01-14
Just another CI update for easier maintain

martadinata666:
As the building binary pretty much stable, we dont need separate 3 job target. This PR merge 3 job into one by using matrix
* https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs
* only binary building affected
* add minimal postgres and mysql integration (up and run), selfcheck
* Selfcheck will throw err if something bad happen

### DIFF
--- a/.github/workflows/Dockerfile.dev
+++ b/.github/workflows/Dockerfile.dev
@@ -1,4 +1,5 @@
-FROM rust:1.65-slim-bullseye
+# Keep tracking base image
+FROM rust:1.66-slim-bullseye
 
 # Set needed env path
 ENV PATH="/opt/aarch64-linux-musl-cross/:/opt/aarch64-linux-musl-cross/bin/:/opt/x86_64-linux-musl-cross/:/opt/x86_64-linux-musl-cross/bin/:$PATH"
@@ -23,6 +24,14 @@ RUN dpkg --add-architecture arm64 && \
     rm -rf /var/lib/apt/lists/* && \
     rustup target add aarch64-unknown-linux-gnu
 
+### armhf deps
+RUN dpkg --add-architecture armhf && \
+    apt update && \
+    apt install -y gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf libc6-armhf-cross libc6-dev-armhf-cross && \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    rustup target add armv7-unknown-linux-gnueabihf
+
 ### Add musl-gcc aarch64 and x86_64
 RUN wget -c https://musl.cc/x86_64-linux-musl-cross.tgz && \
     tar zxf ./x86_64-linux-musl-cross.tgz -C /opt && \
@@ -30,5 +39,10 @@ RUN wget -c https://musl.cc/x86_64-linux-musl-cross.tgz && \
     tar zxf ./aarch64-linux-musl-cross.tgz -C /opt && \
     rm ./x86_64-linux-musl-cross.tgz && \
     rm ./aarch64-linux-musl-cross.tgz
+
+### Add musl target
+RUN rustup target add x86_64-unknown-linux-musl && \
+    rustup target add aarch64-unknown-linux-musl
+    
 
 CMD ["bash"]
--- a/.github/workflows/docker-build-static.yml
+++ b/.github/workflows/docker-build-static.yml
@@ -19,14 +19,8 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
-# In total 5 jobs, all the jobs are containerized
-# ---
 
-#######################################################################################
-# GitHub actions randomly timeout when downloading musl-gcc                           #
-# Using lldap dev image based on https://hub.docker.com/_/rust and musl-gcc bundled   #
-# Look into .github/workflows/Dockerfile.dev for development image details            #
-#######################################################################################
+### CI Docs
 
 # build-ui , create/compile the web
 ### install wasm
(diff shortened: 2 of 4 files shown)


.github/workflows/docker-build-static.yml
```diff
@@ -151,82 +112,90 @@ jobs:
...
+       - name: Download All Artifacts
+         uses: actions/download-artifact@v3
+         with:
+           path: bin/
+       - name: Set executables
+         run: chmod +x bin/*/lldap
+       - name: run lldap with postgres DB
+         run: |
+              bin/x86_64-unknown-linux-musl-lldap-bin/lldap run &
+              sleep 10s
```
nitnelave, line 177: Ha! That's one way to do it :D 
Does that actually work?

nitnelave: If so, we should also do one with sqlite.

.github/workflows/docker-build-static.yml
```diff
@@ -86,58 +87,18 @@ jobs:
           name: ui
           path: app/
 
-  build-armhf:
```
nitnelave, line 89: Nice cleanup!

.github/workflows/docker-build-static.yml
```diff
@@ -34,7 +34,7 @@ env:
 ### run app/build.sh
 ### upload artifacts
 
-# builds-armhf, build-aarch64, build-amd64 create binary for respective arch
+# builds-armhf, build-aarch64, build-amd64 create binary for respective arch look for .github/workflows/Dockerfile.dev for deps list
```
nitnelave, line 37: Move the added comment to a new line, new sentence?

.github/workflows/Dockerfile.dev
```diff
@@ -31,4 +31,12 @@ RUN wget -c https://musl.cc/x86_64-linux-musl-cross.tgz && \
     rm ./x86_64-linux-musl-cross.tgz && \
     rm ./aarch64-linux-musl-cross.tgz
 
+### armhf deps
```
nitnelave, line 34: I'll need to rebuild/republish the build image, right?

.github/workflows/docker-build-static.yml
martadinata666: Well it works, and `lldap selfcheck` pretty much success. Idk if it enough, it just ensure it run fine first. ☕

nitnelave: LLDAP Healthcheck doesn't actually return a non-0 code if there are errors, though. You'd need to grep for ERROR in the output (or I can make a change).

.github/workflows/Dockerfile.dev
martadinata666: right, im thinking keep the deps in one development image. 👍🏼

.github/workflows/docker-build-static.yml
nitnelave: Fixing it in https://github.com/nitnelave/lldap/pull/421

.github/workflows/Dockerfile.dev
```diff
@@ -31,4 +32,12 @@ RUN wget -c https://musl.cc/x86_64-linux-musl-cross.tgz && \
...
+RUN sudo dpkg --add-architecture armhf && \
```
nitnelave, line 36: Remove all the `sudo`, we're `root` and `sudo` is not installed.

.github/workflows/docker-build-static.yml
```diff
@@ -19,35 +19,52 @@ on:
...
-### Upload artifacts
+### aarch64 and amd64 is musl based
+### armv7 is glibc based, musl had issue with time_t when cross compile https://github.com/rust-lang/libc/issues/1848
 
 ## the CARGO_ env
-#CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER: arm-linux-gnueabihf-gcc
-# This will determine which architecture lib will be used.
+### This will determine which architecture linker
+### CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER: arm-linux-gnueabihf-gcc
+### CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER: aarch64-linux-musl-gcc
```
nitnelave, line 45: No need to re-list the variables, they're also in the job definition.

.github/workflows/docker-build-static.yml
```diff
@@ -68,76 +85,36 @@ jobs:
...
-          name: armhf-migration-tool-bin
-          path: target/armv7-unknown-linux-gnueabihf/release/migration-tool
-
 
-  build-aarch64:
+  build-bin:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [armv7-unknown-linux-gnueabihf,aarch64-unknown-linux-musl,x86_64-unknown-linux-musl]
```
nitnelave, line 111: Nit: spaces between the values in the list.

.github/workflows/docker-build-static.yml
```diff
@@ -151,93 +128,109 @@ jobs:
             .cargo/registry/index
             .cargo/registry/cache
             .cargo/git/db
-            target
-          key: lldap-bin-aarch64-${{ hashFiles('**/Cargo.lock') }}
+            target/
```
nitnelave, line 131: Do we need that trailing slash? If so, can you add it above?

.github/workflows/docker-build-static.yml
```diff
@@ -151,93 +128,109 @@ jobs:
...
-#             echo "/opt/aarch64-linux-musl-cross:/opt/aarch64-linux-musl-cross/bin" >> $GITHUB_PATH
-      - name: add musl aarch64 target
-        run: rustup target add aarch64-unknown-linux-musl
-      - name: build lldap aarch4
-        run: cargo build --target=aarch64-unknown-linux-musl --release -p lldap -p migration-tool
-      - name: check path
-        run: ls -al target/aarch64-unknown-linux-musl/release/
-      - name: upload aarch64 lldap artifacts
+            lldap-bin-${{ matrix.target }}-
+      # armhf not existed in dev image, so install here. can be removed after update dev image
```
nitnelave, line 135: The dev image was already updated, you can remove this.

.github/workflows/docker-build-static.yml
```diff
@@ -151,93 +128,109 @@ jobs:
...
+      - name: Add armhf architecture
+        run: dpkg --add-architecture armhf
+      - name: Install build deps for armhf
+        run: apt update && apt install -y gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf libc6-armhf-cross libc6-dev-armhf-cross tar ca-certificates
+      - name: Install ${{ matrix.target }} target
+        run: rustup target add ${{ matrix.target }}
```
nitnelave, line 141: I think we can remove this rustup?

.github/workflows/docker-build-static.yml
```diff
@@ -151,93 +128,109 @@ jobs:
...
-          path: target/x86_64-unknown-linux-musl/release/lldap
-      - name: upload amd64 migration-tool artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: amd64-migration-tool-bin
-          path: target/x86_64-unknown-linux-musl/release/migration-tool
+       - name: Download all artifacts
+         uses: actions/download-artifact@v3
+         with:
+           path: bin/
```
nitnelave, line 185: We only need one artifact, no?

.github/workflows/docker-build-static.yml
```diff
@@ -357,25 +340,25 @@ jobs:
 ### Download artifacts, clean up ui, upload to release page ###
 ###############################################################
   create-release-artifacts:
-     needs: [build-ui,build-armhf,build-aarch64,build-amd64]
+     needs: [lldap-test]
```
nitnelave, line 343: Why `lldap-test`? I'd rather just depend on the builds.

nitnelave (changes requested):

.github/workflows/docker-build-static.yml
martadinata666: Oh the base image updated, should be removed then.

.github/workflows/docker-build-static.yml
martadinata666: Im thinking that that it should pass test first?

.github/workflows/Dockerfile.dev
martadinata666: im updating the dockerfile dev to include rust musl target. ☕

.github/workflows/docker-build-static.yml
nitnelave: oh, right, for release it makes sense. Although it's already checked when submitting the PR.

martadinata666: gonna revert to old behavior then

nitnelave:
FYI, I pushed the rust-dev image with aarch musl

nitnelave:
Btw, you can just ping me when it's ready for review.

martadinata666:
It should be ready, maybe there is some nit you want? ☕

nitnelave (approved):

PR #424, opened 2023-01-21
Wikijs Example

arcoast:
Appreciate the help in Discord, hopefully this will help the next person who struggles like I did.  

As always @nitnelave thanks for all your work on this, sorry about posting in the wrong channel and bothering you with what ended up being a PEBKAC error!

### DIFF
--- a/example_configs/wikijs.md
+++ b/example_configs/wikijs.md
@@ -6,11 +6,11 @@ ldap://lldap:3890
 ```
 ### Admin Bind DN
 ```
-cn=admin,ou=people,dc=example,dc=com
+uid=admin,ou=people,dc=example,dc=com
 ```
 or 
 ```
-cn=readonlyuser,ou=people,dc=example,dc=com
+uid=readonlyuser,ou=people,dc=example,dc=com
 ```
 ### Admin Bind Credentials
 ```
--- a/example_configs/wikijs.md
+++ b/example_configs/wikijs.md
@@ -1,0 +1,64 @@
+# Configuration for WikiJS
+Replace `dc=example,dc=com` with your LLDAP configured domain.
+### LDAP URL
+```
+ldap://lldap:3890
+```
+### Admin Bind DN
+```
+cn=admin,ou=people,dc=example,dc=com
+```
+or 
+```
+cn=readonlyuser,ou=people,dc=example,dc=com
+```
+### Admin Bind Credentials
+```
+ADMINPASSWORD
+```
+or
+```
+READONLYUSERPASSWORD
+```
+### Search Base
+```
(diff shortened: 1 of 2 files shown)


example_configs/wikijs.md
```diff
@@ -0,0 +1,64 @@
+# Configuration for WikiJS
+Replace `dc=example,dc=com` with your LLDAP configured domain.
+### LDAP URL
+```
+ldap://lldap:3890
+```
+### Admin Bind DN
+```
+cn=admin,ou=people,dc=example,dc=com
```
nitnelave, line 9: Small nit : it's better to use uid here instead of cn

example_configs/wikijs.md
```diff
@@ -0,0 +1,64 @@
...
+```
+uid
+```
+### Email Field Mapping
+```
+mail
+```
+### Display Name Field Mapping
+```
+givenname
```
nitnelave, line 52: That's just the first name. The display name should be displayName

example_configs/wikijs.md
```diff
@@ -0,0 +1,64 @@
...
+```
+### Avatar Picture Field Mapping
+```
+jpegPhoto
+```
+### Allow self-registration
+Toggled on
```
nitnelave, line 59: Self registration? That means create a wiki user if there is an LDAP user?

nitnelave (changes requested):
Thanks for the PR! I'm glad you managed to get it working. A couple of notes but it looks good.

example_configs/wikijs.md
arcoast: OK, will change that.

example_configs/wikijs.md
arcoast: I've tried `displayName` and it's probably a WikiJS thing but it gives `???` as the name.  Using givenname at least returns,  as you've pointed out, the users first name.
Screenshot from 2023-01-21 11-56-39

example_configs/wikijs.md
arcoast: Yep

arcoast: One thought I've had is I could just remove a lot of the example config and just pare it down to bare bones "getting LDAP Auth working" and omit the other stuff.

nitnelave (approved):

example_configs/wikijs.md
NotNite: Sorry to necro, but was this ever investigated? My users don't have the first name set, so I have to use displayName, but it ends up resolving to ???.

nitnelave: You can try "cn" maybe

PR #430, opened 2023-01-31
Dex configuration

### LINKED ISSUE #428: Dex client configuration
It would be great if there was a client configuration for dex. More information about the LDAP integration [here](https://dexidp.io/docs/connectors/ldap/).

IgnisDa:
Fixes #428.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,20 +28,20 @@
   </a>
 </p>
 
- - [About](#About)
- - [Installation](#Installation)
-   - [With Docker](#With-Docker)
-   - [From source](#From-source)
-   - [Cross-compilation](#Cross-compilation)
- - [Client configuration](#Client-configuration)
-   - [Compatible services](#compatible-services)
-   - [General configuration guide](#general-configuration-guide)
-   - [Sample client configurations](#Sample-client-configurations)
- - [Comparisons with other services](#Comparisons-with-other-services)
-   - [vs OpenLDAP](#vs-openldap)
-   - [vs FreeIPA](#vs-freeipa)
- - [I can't log in!](#i-cant-log-in)
- - [Contributions](#Contributions)
+- [About](#about)
+- [Installation](#installation)
+  - [With Docker](#with-docker)
+  - [From source](#from-source)
+  - [Cross-compilation](#cross-compilation)
+- [Client configuration](#client-configuration)
+  - [Compatible services](#compatible-services)
+  - [General configuration guide](#general-configuration-guide)
+  - [Sample client configurations](#sample-client-configurations)
+- [Comparisons with other services](#comparisons-with-other-services)
+  - [vs OpenLDAP](#vs-openldap)
+  - [vs FreeIPA](#vs-freeipa)
+- [I can't log in](#i-cant-log-in)
+- [Contributions](#contributions)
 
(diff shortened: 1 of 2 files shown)


IgnisDa:
I also ran a markdown formatter on the README

README.md
```diff
@@ -286,25 +290,25 @@ LLDAP is much lighter to run (<10 MB RAM including the DB), easier to
 configure (no messing around with DNS or security policies) and simpler to
 use. It also comes conveniently packed in a docker container.
 
-## I can't log in!
```
nitnelave, line 289: I'd like to keep the bang here, it better expresses the user frustration :)

example_configs/dex_config.yml
```diff
@@ -0,0 +1,25 @@
...
+# ##############################
+
+connectors:
+  - type: ldap
+    id: ldap
+    name: LDAP
+    config:
+      host: lldap-host # make sure it does not start with `ldap://`
+      port: 3890
+      insecureNoSSL: true
```
nitnelave, line 15: Maybe mention that this should be false if using LDAPS ?

nitnelave: Same with the skipVerify below.

example_configs/dex_config.yml
```diff
@@ -0,0 +1,25 @@
...
+      insecureSkipVerify: true
+      bindDN: uid=admin,ou=people,dc=example,dc=com # replace admin with your admin user
+      bindPW: very-secure-password # replace with your admin password
+      userSearch:
+        baseDN: ou=people,dc=example,dc=com
+        username: uid
+        idAttr: uid
+        emailAttr: mail
+        nameAttr: displayName
+        preferredUsernameAttr: uid
```
nitnelave, line 25: You don't want to add the group config? It should probably look like that:

```yaml
    groupSearch:
      baseDN: ou=groups,dc=example,dc=com
      filter: "(objectClass=groupOfUniqueNames)"
      userMatchers:
      - userAttr: uid
        groupAttr: member
      nameAttr: displayName
```

nitnelave:
Thanks for the PR!

example_configs/dex_config.yml
IgnisDa: I did not have to add this to my config for dex to function properly. Is this required?

nitnelave: It's only required if you want to expose the groups to other apps, but it would be useful (e.g. only allow access to service X if the user is in group `x_users`).
I think it would be valuable to add to the example config, if you can confirm that it works.

I don't know much about Dex, maybe they have a way in their UI to see the user's groups that you could use to confirm that it works?

example_configs/dex_config.yml
IgnisDa: How do we enable ldaps for lldap?

nitnelave: https://github.com/nitnelave/lldap/blob/main/lldap_config.docker_template.toml#L128

example_configs/dex_config.yml
IgnisDa: Yes it works will update.

nitnelave (approved):

PR #434, opened 2023-02-04
Add example for Gitea's simple auth mode

rexzhang:
test on:
- LLDAP 0.4.1
- Gitea 1.18.3

### DIFF
--- a/example_configs/gitea.md
+++ b/example_configs/gitea.md
@@ -41,3 +41,25 @@ Check `Remove Users from syncronised teams...`
 The `Map LDAP groups to Organization teams` config is JSON formatted and can be extended to as many groups as needed.
 
 Replace every instance of `dc=example,dc=com` with your configured domain.
+
+# Configuration for Gitea in `simple auth` mode
+
+* Host: Your lldap server's ip/hostname
+* Port: Your lldap server's port (389 by default)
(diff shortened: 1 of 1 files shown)


nitnelave:
I'm confused. What's the difference between the added configuration and the existing one?

rexzhang:
Avoid `admin` account password leakage

nitnelave:
Sorry, I'm not familiar with gitea and its config. How does the previous configuration leak the admin password, and how does this one prevent it?

rexzhang:
In the BindDN mode, we need to fill the admin account's bind password of LLDAP into Gitea's config. In the simple auth mode, this isn't necessary.

nitnelave:
Okay, I see the point. I had a closer look at the proposed config, but it's exactly the same as the one before, only without the bind DN and password, correct? In that case we can just make a note about simple auth without repeating all the settings.

rexzhang:
Ok, It's updated

nitnelave:
Thanks!

benisai:
Here ya go. Select ldap-simple in Gitea.

Screenshot 2024-09-03 at 5 34 57 PM

Screenshot 2024-09-03 at 5 35 09 PM

Screenshot 2024-09-03 at 5 35 18 PM

nitnelave (approved):
Thanks!

PR #443, opened 2023-02-12
Fix: Change input field to password type in change_password ui

DarkSpir:
Change input field type for field old_password from its default "text" to "password"


nitnelave (approved):
Thanks!

PR #453, opened 2023-02-15
Improve reset email errors

nitnelave:
This makes the configuration more robust to a trailing slash in the domain, and improves the error messages when clicking on a non-existing or expired token.


nitnelave:
Fixes #365

Issue #265, opened 2022-07-17
Improve page loading experence


jim-kirisame:
Hi, currently the page is blank while loading the huge wasm, which is annoying when in slow network connection.

Add some loading animation will be better, or could you optimize the size of the wasm?

nitnelave:
Reference: https://rustwasm.github.io/docs/book/reference/code-size.html#optimizing-builds-for-code-size

nitnelave:
It would also be nice to get rid of rollup and replace it with a non-npm dependency, ideally in Rust.

nitnelave:
I also investigated wasm-opt, but it doesn't help at all (actually makes the gzipped version slightly bigger).

PR #469, opened 2023-03-05
app: get rid of rollup, gzip the wasm

nitnelave:
This way we get rid of the npm dependency (it's not needed anyway), and we reduce the file size served (2.6MB -> 700kB)


nitnelave:
Fixed #265

PR #470, opened 2023-03-09
feat(ghcr) add ghcr image workflow

xstar97:
this a workflow to build a image for github whenever a tag release is committed.

it will only build when a semVer tag is made for ex: `v.0.0.1`.

to use this image would be the same as a docker image. 

`docker pull ghcr.io/nitnelave/lldap:v0.4.1`


nitnelave:
What's the reason for adding support for ghcr? Why is it needed?

I think we should rather incorporate this into the other docker workflow, so we don't have to repeat the work. In addition, it'll support the various architectures that we do.

@martadinata666 can help if you have trouble adding to the workflow.

PR #475, opened 2023-03-17
docker: Allow chown commands to fail

jfroy:
chown may not be allowed in containers with restricted security contexts. See https://github.com/truecharts/charts/pull/7779.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -22,8 +22,8 @@ if [[ ! -r "$CONFIG_FILE" ]]; then
 fi
 
 echo "> Setup permissions.."
-find /app \! -user "$UID" -exec chown "$UID:$GID" '{}' +
-find /data \! -user "$UID" -exec chown "$UID:$GID" '{}' +
+find /app \! -user "$UID" -exec chown "$UID:$GID" '{}' + || true
+find /data \! -user "$UID" -exec chown "$UID:$GID" '{}' +  || true
 
 
 echo "> Starting lldap.."


martadinata666:
I saw the truecharts threads about adding lldap. It dropped capabilities to `chown`. The script run as root then ensures that `/data` and `/app` are readable and writeable by the `UID:GID` number `1000` by default, lastly dropping to `UID:GID` by using `gosu` and truecharts doesn't like this method, it looking on predefined permissions that doesn't involve chown. Technically speaking, can truecharts just override the entrypoint? It just need to remove `chown` part, as `/app` and `/data` already got `1000:1000` permission by default.

```
#!/usr/bin/env bash
set -euo pipefail

CONFIG_FILE=/data/lldap_config.toml

if [[ ( ! -w "/data" ) ]] || [[ ( ! -d "/data" ) ]]; then
  echo "[entrypoint] The /data folder doesn't exist or cannot be written to. Make sure to mount
  a volume or folder to /data to persist data across restarts, and that the current user can
  write to it."
  exit 1
fi

if [[ ! -f "$CONFIG_FILE" ]]; then
  echo "[entrypoint] Copying the default config to $CONFIG_FILE"
  echo "[entrypoint] Edit this file to configure LLDAP."
  cp /app/lldap_config.docker_template.toml $CONFIG_FILE
fi

if [[ ! -r "$CONFIG_FILE" ]]; then
  echo "[entrypoint] Config file is not readable. Check the permissions"
  exit 1;
fi

echo "> Starting lldap.."
echo ""
exec gosu "$UID:$GID" /app/lldap "$@"

exec "$@"

```

MaverickD650:
The config file isn't directly editable with Truecharts. This a limitation of the TrueNAS implementation of Kubernetes. Options are exposed through ENV vars I can expose in the UI or I believe the config file can be generated with a config map (not too familiar with this). Had the app run as 0 this doesn't work unless I'm missing something. I also believe Truecharts doesn't use docker as such just the docker image.

martadinata666:
> The config file isn't directly editable with Truecharts. This a limitation of the TrueNAS implementation of Kubernetes. Options are exposed through ENV vars I can expose in the UI or I believe the config file can be generated with a config map (not too familiar with this). Had the app run as 0 this doesn't work unless I'm missing something. I also believe Truecharts doesn't use docker as such just the docker image.

Yes, I saw the PR, also it cant do `chown` as this ensure the user can R/W app and data.
My question is,
* can it mount some file? 
* can it override `entrypoint`?

Also can you try using `nitnelave/lldap:latest-debian` as alpine build got some perm issue that oversighted.

Thanks.

MaverickD650:
I can create a config map that generates lldap_config.toml that I think is what you're getting at and it would then read that file. It would be an alternative to exposing using the ENVs. I can look at rewriting the app later and see how far I get but I don't typically write apps that way so my knowledge is more limited. A blocker in doing this though was generating jwt_secret it complicated matters considerably.

martadinata666:
I'm curious, why not let users fill the jwt_secret by themself? 🤔 and send notes to generate themself, is it possible? `LC_ALL=C tr -dc 'A-Za-z0-9!#%&'\''()*+,-./:;<=>?@[\]^_{|}~' </dev/urandom | head -c 32; echo ''` Like portainer template had description, to make notes for user need to do before deploy.

MaverickD650:
The project attempts to provide sane secure defaults where possible which is why I leaned more towards auto generating it and expose options where it adds value

MaverickD650:
See https://github.com/truecharts/charts/pull/7802 for another go happy to take suggestions on getting this working.

MaverickD650:
Still trying to chown on the debian image.

martadinata666:
> Still trying to chown on the debian image.

changing image doesn't mean it will work, it just had correct predefined permission, it still run `docker-entrypoint.sh` that got `chown`.

I'm creating simple charts based on your previous PR https://github.com/truecharts/charts/pull/7779 , and overriding the entrypoints, and it work [action run](https://github.com/martadinata666/charts/actions/runs/4446958823) , you can adapt it to your new charts PR

MaverickD650:
> > Still trying to chown on the debian image.
> 
> changing image doesn't mean it will work, it just had correct predefined permission, it still run `docker-entrypoint.sh` that got `chown`.
> 
> I'm creating simple charts based on your previous PR [truecharts/charts#7779](https://github.com/truecharts/charts/pull/7779) , and overriding the entrypoints, and it work [action run](https://github.com/martadinata666/charts/actions/runs/4446958823) , you can adapt it to your new charts PR

Found it and committed to the PR now thanks for the help.

martadinata666:
For behavior to allow `chown` to fail, pretty skeptical that it will work. As it comes from the k3s itself and kills the pod, but who knows, I don't really sure about k3s behaviors.

jfroy:
> For behavior to allow `chown` to fail, pretty skeptical that it will work. As it comes from the k3s itself and kills the pod, but who knows, I don't really sure about k3s behaviors. rofl

k3s is just a k8s distribution, it doesn't behave differently from standard k8s wrt to security. The TrueChart project is trying to provide charts with good container/pod security, and this is where the chown issue came from. In general containers should be resilient to operations that will fail when run with a restricted security policy.

TrueChart uses init containers to do some privileged setup before running the main container, as one pattern used to run containers with the least privileged required.

In any case, closing since we worked around the issue.

PR #479, opened 2023-03-17
Update devcontainer and build instructions

pixelrazor:
remove references to node/rollup since we use gzip now


nitnelave:
For rust-analyzer, it's possible to configure it to use a different folder for the build directory, so it doesn't conflict with cargo.

pixelrazor:
> For rust-analyzer, it's possible to configure it to use a different folder for the build directory, so it doesn't conflict with cargo.

eh, i'm not too determined to mess with the devcontainer file to get this working. it worked in the past, and i suspect there's a recent bug:
https://github.com/rust-lang/rust-analyzer/issues/14351

.devcontainer/Dockerfile
```diff
@@ -13,12 +13,9 @@ RUN groupadd --gid $USER_GID $USERNAME \
     && chmod 0440 /etc/sudoers.d/$USERNAME
 
 RUN apt update && \
-    curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - && \
-    apt update && \
-    apt install -y --no-install-recommends nodejs libssl-dev musl-dev make perl curl 
+    apt install -y --no-install-recommends libssl-dev musl-dev make perl curl gzip
```
nitnelave, line 16: We can throw in a `&& rm -rf /var/lib/apt/lists/*` for good measure, to reduce the image size. Probably not very important, but still :)

nitnelave (approved):

.devcontainer/Dockerfile
pixelrazor: done (though it didn't make a visible different in the 1.94gb image size)

nitnelave:
You've got some conflicts, but once it's solved, you're good to go.

pixelrazor:
> You've got some conflicts, but once it's solved, you're good to go.

i think i fixed them a minute before you left the comment

nitnelave:
Can you rebase on top of master? There are still some conflicts.

martadinata666:
Looking on this, isn't it better to use prebuild lldap rust development image `nitnelave/rust-dev:latest`, then we just need install the `wasm-pack`.
```
RUN RUSTFLAGS=-Ctarget-feature=-crt-static cargo install wasm-pack \
    && rustup target add wasm32-unknown-unknown
```

Does it need a user creation? 🤔

pixelrazor:
> Looking on this, isn't it better to use prebuild lldap rust development image `nitnelave/rust-dev:latest`, then we just need install the `wasm-pack`.
> 
> ```
> RUN RUSTFLAGS=-Ctarget-feature=-crt-static cargo install wasm-pack \
>     && rustup target add wasm32-unknown-unknown
> ```
> 
> Does it need a user creation? thinking

I could use that container as a base. I'm not picky on the container itself as long as it works. I chose the base used currently because it should perform a bit better than the smaller/lightweight versions of the rust container, but that difference probably isn't SUPER important - that being said, i don't know what rust-dev is based off of.

For the user - i think i had stronger reasons that i've since forgotten, but it's in line with best practice: https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user

martadinata666:
I see, i don't know how far people will use devcontainer it will only build `x86_64-unknown-linux-gnu` target, on other development container already prepared to cross build with their deps. Look here for reference https://github.com/nitnelave/lldap/blob/main/.github/workflows/Dockerfile.dev

pixelrazor:
The way i view it, the devcontainer is just what's needed to develop lldap. anything beyond that is up to the user to figure out, whether they install the tools in the devcontainer themself or use once of the dockerfiles from this repo

nitnelave:
I don't remember so I'm not sure, but isn't nitnelave/rust-dev this dev
container? :D

PR #483, opened 2023-03-20
Dark theme

### LINKED ISSUE #250: Add support for dark mode
Revamp https://github.com/nitnelave/lldap/pull/215 but based on something like https://vinorodrigues.github.io/bootstrap-dark/

Things that we want to have:
 - Toggle button
 - auto-detect based on browser preference

Maybe JS-based, with a class on the body.

pixelrazor:
Resolves #250 

Uses [bootstrap nighshade](https://github.com/vinorodrigues/bootstrap-dark-5/blob/main/docs/bootstrap-nightshade.md). Initial state is set by browser preference, but can be overridden with user choice.

Some of the `-dark` and `-muted` classes did not have any contrast in dark mode, so they were changed

### DIFF
--- a/app/index_local.html
+++ b/app/index_local.html
@@ -31,6 +31,11 @@
     <link
       rel="stylesheet"
       href="/static/style.css" />
+    <script>
+      function inDarkMode(){
+        return darkmode.inDarkMode;
+      }
+    </script>
     <meta name="viewport" content="width=device-width, initial-scale=1">
 </head>
 
--- a/app/index_local.html
+++ b/app/index_local.html
@@ -6,13 +6,16 @@
     <title>LLDAP Administration</title>
     <script src="/static/main.js" type="module" defer></script>
     <link
-      href="/static/bootstrap.min.css"
+      href="/static/bootstrap-nightshade.min.css"
       rel="preload stylesheet"
-      integrity="sha384-+0n0xVW2eSR5OomGNYDnhzAbDsOXxcvSN1TPprVMTNDbiYZCxYbOOl7+AMvyTG2x"
+      integrity="sha384-CvItGYrXmque42UjYhp+bjRR8tgQz78Nlwk42gYsNzBc6y0DuXNtdUaRzr1cl2uK"
       as="style" />
     <script
       src="/static/bootstrap.bundle.min.js"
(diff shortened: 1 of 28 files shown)


app/src/components/app.rs
```diff
@@ -29,6 +30,25 @@ use yew_router::{
...
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = darkmode)]
+    fn toggleDarkMode(doSave: bool);
+
+    #[wasm_bindgen]
+    fn inDarkMode() -> bool;
+}
+
+#[function_component(DarkModeToggle)]
```
nitnelave, line 42: Nice!

app/index_local.html
```diff
@@ -28,6 +31,11 @@
     <link
       rel="stylesheet"
       href="/static/style.css" />
+    <script>
+      function inDarkMode(){
+        return darkmode.inDarkMode;
```
nitnelave, line 36: does that load the browser's default settings?

app/src/components/app.rs
```diff
@@ -29,6 +30,25 @@ use yew_router::{
     BrowserRouter, Switch,
 };
 
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = darkmode)]
+    fn toggleDarkMode(doSave: bool);
```
nitnelave, line 36: Save is done through what? Cookies, local storage?

app/index_local.html
nitnelave: I'm thinking of `window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'`

app/src/components/app.rs
```diff
@@ -328,7 +349,7 @@ impl App {
 
     fn view_footer(&self) -> Html {
         html! {
-          <footer class="text-center text-muted fixed-bottom bg-light py-2">
+          <footer class="text-center fixed-bottom bg-light py-2">
```
nitnelave, line 352: Isn't that going to be too bold/visible? I don't want it to be intrusive.

nitnelave:
Overall, it looks good!

app/src/components/app.rs
nitnelave: We'll need to change the `bg-light` as well, it's awful in dark mode :D

nitnelave:
The general links (to users, groups, etc) are hard to read in dark mode, the contrast is not high enough.

nitnelave:
Didn't mean to close this, you can re-open against master.

PR #484, opened 2023-03-20
Create schema command

pixelrazor:
New command to test a dataabase url and ensure the schema exists and is up to date. Update migration docs to include usage of this command

### DIFF
--- a/docs/database_migration.md
+++ b/docs/database_migration.md
@@ -83,7 +83,7 @@ or
 
 `mysql -u <username> -p <database> < /path/to/dump.sql`
 
-NOTE: MySQL will limit avatar size to 64Kb. You may need to remove avatars for some users. (Untested: you could probably change the column type to LONGBLOB )
+NOTE: MySQL will limit avatar size to 64Kb. You may need to remove avatars for some users. See #486
 
 ## Switch to new database
 
--- a/docs/database_migration.md
+++ b/docs/database_migration.md
@@ -6,43 +6,53 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily 
 
 The process is as follows:
 
-1. Create a dump of existing data.
-2. Change all `CREATE TABLE ...` lines to `DELETE FROM tablename;`. We will later have LLDAP create the schema for us, so we want to clear out existing data to replace it with the original data.
-3. Do any syntax fixes for the target db syntax
-4. Change your LLDAP config database_url to point to the new target and restart.
-5. After LLDAP has started, stop it.
-6. Execute the manicured dump file against the new database.
+1. Create empty schema on target database
+2. Dump existing values
+3. Sanitize for target DB (not always required)
+4. Insert data into target
+5. Change LLDAP config to new target
 
 The steps below assume you already have PostgreSQL or MySQL set up with an empty database for LLDAP to use.
 
-## Create a dump
+## Create schema on target
 
-First, we must dump the existing data to a file. The dump must be tweaked slightly according to your target db. See below for commands
-
-### PostgreSQL
-
-PostgreSQL uses a different hex string format and doesn't support `PRAGMA`.
+LLDAP has a command that will connect to a target database and initialize the
+schema. If running with docker, run the following command to use your active
+instance (this has the benefit of ensuring your container has access):
 
 ```
-sqlite3 /path/to/lldap/config/users.db .dump | \
-sed -r -e "s/X'([[:xdigit:]]+'[^'])/'\\\x\\1/g" \
--e 's/^CREATE TABLE IF NOT EXISTS "([^"]*)".*/DELETE FROM \1;/' \
--e '/^PRAGMA.*/d' > /path/to/dump.sql
+docker exec -it <LLDAP container name> /app/lldap create_schema -d <Target database url>
 ```
 
-### MySQL
+If it succeeds, you can proceed to the next step.
 
-MySQL doesn't support `PRAGMA`.
+## Create a dump of existing data
+
+We want to dump all existing values to some file. The dump should consist just INSERT
+statements. There are various ways to do this, but a simple enough way is filtering a
+whole database dump. For example:
 
 ```
-sqlite3 /path/to/lldap/config/users.db .dump | \
--e 's/^CREATE TABLE IF NOT EXISTS "([^"]*)".*/DELETE FROM \1;/' \
--e '/^PRAGMA.*/d' > /path/to/dump.sql
+sqlite3 /path/to/lldap/config/users.db .dump | grep "^INSERT" > /path/to/dump.sql
 ```
 
-## Generate New Schema
+## Sanitize data (if needed)
 
-Modify your `database_url` in `lldap_config.toml` (or `LLDAP_DATABASE_URL` in the env) to point to your new database. Restart LLDAP and check the logs to ensure there were no errors connecting and creating the tables. After that, stop LLDAP. Now we can import our original data!
+Some databases might use different formats for some data - for example, PostgreSQL uses
+a different syntax for hex strings than SQLite.
+
+### To PostgreSQL
+
+PostgreSQL uses a different hex string format. The command below should switch SQLite
+format to PostgreSQL format.
+
+```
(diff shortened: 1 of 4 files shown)


docs/database_migration.md
```diff
@@ -6,43 +6,53 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
 
 ```
-sqlite3 /path/to/lldap/config/users.db .dump | \
-sed -r -e "s/X'([[:xdigit:]]+'[^'])/'\\\x\\1/g" \
--e 's/^CREATE TABLE IF NOT EXISTS "([^"]*)".*/DELETE FROM \1;/' \
--e '/^PRAGMA.*/d' > /path/to/dump.sql
+sqlite3 /path/to/lldap/config/users.db .dump | grep "^INSERT" > /path/to/dump.sql
 ```
 
-### MySQL
```
nitnelave, line 33: Maybe let's keep a not about MySQL to tell users that we didn't forget about them, but there's nothing to do.

docs/database_migration.md
```diff
@@ -6,43 +6,53 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
-2. Change all `CREATE TABLE ...` lines to `DELETE FROM tablename;`. We will later have LLDAP create the schema for us, so we want to clear out existing data to replace it with the original data.
-3. Do any syntax fixes for the target db syntax
-4. Change your LLDAP config database_url to point to the new target and restart.
-5. After LLDAP has started, stop it.
-6. Execute the manicured dump file against the new database.
+1. Create empty schema on target database
+2. Dump existing values
+3. Sanitize for target DB (not always required)
+4. Insert data into target
+5. Change LLDAP config to new target
```
nitnelave, line 13: ... and restart.

docs/database_migration.md
```diff
@@ -6,43 +6,53 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
 The process is as follows:
 
-1. Create a dump of existing data.
-2. Change all `CREATE TABLE ...` lines to `DELETE FROM tablename;`. We will later have LLDAP create the schema for us, so we want to clear out existing data to replace it with the original data.
-3. Do any syntax fixes for the target db syntax
-4. Change your LLDAP config database_url to point to the new target and restart.
-5. After LLDAP has started, stop it.
-6. Execute the manicured dump file against the new database.
+1. Create empty schema on target database
+2. Dump existing values
```
nitnelave, line 10: Stop LLDAP before? I think SQLite doesn't like it when you dump during a write.

server/src/infra/cli.rs
```diff
@@ -94,6 +97,13 @@ pub struct TestEmailOpts {
     pub smtp_opts: SmtpOpts,
 }
 
+#[derive(Debug, Parser, Clone)]
+pub struct CreateSchemaOpts {
+    /// Database connection URL
+    #[clap(short, long, env = "LLDAP_CREATE_SCHEMA_DATABASE_URL")]
```
nitnelave, line 103: I think this should just be `LLDAP_DATABASE_URL`, and we'll read from the config by default

nitnelave: We could actually just make that part of the general CLI options, to be able to override the DB url on the command line.

nitnelave: and you'll need to update `configuration.rs` to merge it into the config.

server/src/main.rs
```diff
@@ -189,12 +189,40 @@ fn run_healthcheck(opts: RunOpts) -> Result<()> {
     std::process::exit(i32::from(failure))
 }
 
+async fn create_schema(database_url: String) -> Result<()> {
+    let sql_pool = {
+        let mut sql_opt = sea_orm::ConnectOptions::new(database_url.clone());
+        sql_opt
+            .max_connections(5)
```
nitnelave, line 196: Nit: max_connections(1), we'll only ever need 1 connection.

nitnelave (changes requested):

martadinata666:
Can you also help check the database sanitization for mysql/mariadb. ☕ 

thanks

docs/database_migration.md
```diff
@@ -6,52 +6,87 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
-## Generate New Schema
+### To MySQL
 
-Modify your `database_url` in `lldap_config.toml` (or `LLDAP_DATABASE_URL` in the env) to point to your new database. Restart LLDAP and check the logs to ensure there were no errors connecting and creating the tables. After that, stop LLDAP. Now we can import our original data!
+MySQL mostly cooperates, but it gets some errors if you don't escape the `groups` table. Run the
+following command to wrap all table names in backticks for good measure, and wrap the inserts in
+a transaction:
+
+```
+sed -i -r -e 's/^INSERT INTO ([a-zA-Z0-9_]+) /INSERT INTO `\1` /' \
```
nitnelave, line 64: What does this line do? Can you break it down for me?

docs/database_migration.md
```diff
@@ -6,52 +6,87 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
+or 
+
+`psql -d <database> -U <username> -W -f /path/to/dump.sql`
+
 ### MySQL
 
-`mysql -u < -p <database> < /path/to/dump.sql`
+`mysql -u <username> -p <database> < /path/to/dump.sql`
+
+NOTE: MySQL will limit avatar size to 64Kb. You may need to remove avatars for some users. (Untested: you could probably change the column type to LONGBLOB )
```
nitnelave, line 86: Rather than Untested, maybe refer to the issue you just opened.

server/src/main.rs
nitnelave: Still 5?

server/src/main.rs
```diff
@@ -189,12 +189,45 @@ fn run_healthcheck(opts: RunOpts) -> Result<()> {
...
+    debug!("CLI: {:#?}", &opts);
+    let config = infra::configuration::init(opts)?;
+    infra::logging::init(&config)?;
+    let database_url = config.database_url;
+
+    actix::run(
+        create_schema(database_url).unwrap_or_else(|e| error!("Could not create schema: {:#}", e)),
+    )?;
+
+    info!("End.");
```
nitnelave, line 220: Schema created successfully?

nitnelave (approved):

server/src/main.rs
pixelrazor: oops, accidentally reverted the line when testing.

docs/database_migration.md
pixelrazor: yep! it goes from 
```
INSERT INTO groups VALUES
```
to
```
INSERT INTO `groups` VALUE
```

I think `groups` is a special word in mysql, it threw syntax errors for JUST the groups table when not escaped like that. It's listed as a [reserved keyword](https://dev.mysql.com/doc/refman/8.0/en/keywords.html).

This line generalizes it, but i could easily JUST update the `groups` insert statements

nitnelave: oh, yeah, the `sed` part I had, it was more about the couple of lines below, sorry.

pixelrazor: Oh, the next one adds START TRANSACTION; to the beginning, and the one after adds COMMIT; to the end

nitnelave (approved):

docs/database_migration.md
```diff
@@ -6,52 +6,87 @@ NOTE: [pgloader](https://github.com/dimitri/pgloader) is a tool that can easily
...
+or 
+
+`psql -d <database> -U <username> -W -f /path/to/dump.sql`
+
 ### MySQL
 
-`mysql -u < -p <database> < /path/to/dump.sql`
+`mysql -u <username> -p <database> < /path/to/dump.sql`
+
+NOTE: MySQL will limit avatar size to 64Kb. You may need to remove avatars for some users. See #486
```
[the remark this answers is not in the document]
nitnelave: I'll remove it myself to unblock the PR

nitnelave, line 86:
```suggestion

```

PR #487, opened 2023-03-20
Dark Theme

pixelrazor:
Resolves https://github.com/nitnelave/lldap/issues/250

Uses [bootstrap nighshade](https://github.com/vinorodrigues/bootstrap-dark-5/blob/main/docs/bootstrap-nightshade.md). Initial state is set by browser preference, but can be overridden with user choice.

Some of the -dark and -muted classes did not have any contrast in dark mode, so they were changed

THIS IS A DRAFT UNTIL I MIGRATE AND ADDRESS COMMENTS FROM #483

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -34,6 +34,11 @@
     <link
       rel="stylesheet"
       href="/static/style.css" />
+    <script>
+      function inDarkMode(){
+        return darkmode.inDarkMode;
+      }
+    </script>
     <meta name="viewport" content="width=device-width, initial-scale=1">
 </head>
 
--- a/app/index.html
+++ b/app/index.html
@@ -6,14 +6,18 @@
     <title>LLDAP Administration</title>
     <script src="/static/main.js" type="module" defer></script>
     <link
-      href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.1/dist/css/bootstrap.min.css"
+      href="https://cdn.jsdelivr.net/npm/bootstrap-dark-5@1.1.3/dist/css/bootstrap-nightshade.min.css"
       rel="preload stylesheet"
-      integrity="sha384-+0n0xVW2eSR5OomGNYDnhzAbDsOXxcvSN1TPprVMTNDbiYZCxYbOOl7+AMvyTG2x"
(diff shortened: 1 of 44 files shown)


app/index.html
```diff
@@ -30,6 +34,11 @@
     <link
       rel="stylesheet"
       href="/static/style.css" />
+    <script>
+      function inDarkMode(){
+        return darkmode.inDarkMode;
+      }
+    </script>
```
pixelrazor, line 41: > does that load the browser's default settings?

It gets the current setting, whether it be browser default or user settings if overriden. It's all from [this](https://github.com/vinorodrigues/bootstrap-dark-5/blob/main/docs/darkmode.js.md#indarkmode) helper lib - I didn't know how to bind to that variable so this function gives me a way to bind. We shouldn't circumvent these libs since they manage the storage of user perefence - they should be our source of truth

app/src/components/app.rs
```diff
@@ -29,6 +30,25 @@ use yew_router::{
     BrowserRouter, Switch,
 };
 
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = darkmode)]
+    fn toggleDarkMode(doSave: bool);
```
pixelrazor, line 36: > Save is done through what? Cookies, local storage?

[ref for the js function](https://github.com/vinorodrigues/bootstrap-dark-5/blob/main/docs/darkmode.js.md#indarkmode)
It looks like it could use either cookies or local storage, depending on user consent. Either way, it is stored under the key `bs.prefers-color-scheme` (customizable)

app/src/components/app.rs
```diff
@@ -323,7 +345,7 @@ impl App {
 
     fn view_footer(&self) -> Html {
         html! {
-          <footer class="text-center text-muted fixed-bottom bg-light py-2">
+          <footer class="text-center fixed-bottom bg-light py-2">
```
pixelrazor, line 348: > Isn't that going to be too bold/visible? I don't want it to be intrusive.

Maybe? It was not visible at all in dark mode before.

> We'll need to change the bg-light as well, it's awful in dark mode :D

I can do that

pixelrazor (commented):
> The general links (to users, groups, etc) are hard to read in dark mode, the contrast is not high enough.

Ack. I'll find a different class

nitnelave:
Looks great!

nitnelave (approved):

Issue #486, opened 2023-03-20
MySQL schema uses BLOB instead of LONGBLOB


pixelrazor:
Discovered when testing #484. MySQL BLOB is limited to 64kb, which is far too small for avatar. It looks like [avatar is already configured properly](https://github.com/nitnelave/lldap/blob/b2cfc0ed03b41db25b49a03da31e57582c661c29/server/src/domain/model/users.rs#L59), but it that doesn't appear to be effecting the schema creation.

pixelrazor:
Digging more i found that the issue is probable [here](https://github.com/nitnelave/lldap/blob/b2cfc0ed03b41db25b49a03da31e57582c661c29/server/src/domain/sql_migrations.rs#L105) where we create it without the binary size. Do we need to change that or the v2 migration to do the right type? or maybe add a v3 migration to upgrade that type?

nitnelave:
We'll probably need to do a fancy dance like we do with display_name at the end of the migration file to move the columns around and change the type, using a v3 migration marker.

PR #489, opened 2023-03-20
server: ensure first/last name nullable, make avatar long blob in DB

### LINKED ISSUE #474: User creation requires firstname, lastname
I have recently updated to 0.4.1.

At some point it seems that creating a new user has gone from requiring uid, email, and display name, to also requiring first name and last name.

I can't see any notice of this change, is it intended?

for my own use case first name and last name are not used....I only require username and email from my users. (I am happy to add username as display name as well, as that allows for the possibility to change  at a later date).

It is also not possible to delete firstname or lastname after user creation: 

```
Errors: [updateUser:2:3: Database error: `Execution Error: error returned from database: (code: 1299) NOT NULL constraint failed: users.first_name`]
```

A consequence of this, for my use case, is that I am currently copying the user id into three other settings, in order to create a user (sub-optimal)...and am concerned that the absence of these fields in previously created users will cause other spin off problems (some users are reporting seeing some problems logging into their profile, and I'm concerned that this may be the reason).

nitnelave:
Fixes #474, #486.

### DIFF
--- a/server/src/domain/sql_migrations.rs
+++ b/server/src/domain/sql_migrations.rs
@@ -11,7 +11,7 @@ use tracing::{info, instrument, warn};
 
 use super::sql_tables::LAST_SCHEMA_VERSION;
 
-#[derive(Iden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone)]
+#[derive(Iden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum Users {
     Table,
     UserId,
@@ -27,7 +27,7 @@ pub enum Users {
     Uuid,
 }
 
-#[derive(Iden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone)]
+#[derive(Iden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum Groups {
     Table,
     GroupId,
@@ -36,7 +36,7 @@ pub enum Groups {
     Uuid,
 }
 
-#[derive(Iden)]
+#[derive(Iden, Clone, Copy)]
 pub enum Memberships {
     Table,
     UserId,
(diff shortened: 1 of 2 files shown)


pixelrazor:
Has the schema for mysql, sqlite, and postgres been verified?

server/src/domain/sql_migrations.rs
```diff
@@ -334,6 +334,132 @@ pub async fn upgrade_to_v1(pool: &DbConnection) -> std::result::Result<(), sea_o
...
+    //  - rename the column to a temporary name
+    //  - create the column with the new definition
+    //  - copy the data from the temp column to the new one
+    //  - update the new one if there are changes needed
+    //  - drop the old one
+    let builder = pool.get_database_backend();
+    pool.transaction::<_, (), sea_orm::DbErr>(move |transaction| {
+        Box::pin(async move {
+            #[derive(Iden)]
+            enum TempUsers {
```
pixelrazor, line 354: Is this function a general usage function?/ i looks like it is, but this enum is names for the Users table

server/src/domain/sql_migrations.rs
```diff
@@ -334,6 +334,132 @@ pub async fn upgrade_to_v1(pool: &DbConnection) -> std::result::Result<(), sea_o
...
+        ColumnDef::new(Users::LastName).string_len(255).to_owned(),
+        [builder.build(
+            Query::update()
+                .table(Users::Table)
+                .value(Users::LastName, Option::<String>::None)
+                .cond_where(Expr::col(Users::LastName).eq("")),
+        )],
+    )
+    .await?;
+    // Change Avatar from binary to blob(long), because for MySQL this is 65kb.
```
pixelrazor, line 449:
```suggestion
    // Change Avatar from binary to blob(long), because for MySQL this is 64kb.
```

server/src/domain/sql_migrations.rs
nitnelave: Hmm, the name doesn't actually matter, but you're right that it seems to imply that it does. I'll change that.

server/src/domain/sql_migrations.rs
nitnelave: Thanks!

nitnelave:
> Has the schema for mysql, sqlite, and postgres been verified?

I might have been a bit hasty, but I'm pretty sure it's okay. Would love to be proven wrong, though :) At least the CI tests work, so standing up the server works with all 3 DBs.

PR #519, opened 2023-04-02
added Shaarli configuration example

Hobbabobba:


example_configs/shaarli.md
```diff
@@ -0,0 +1,11 @@
+# Configuration for shaarli
+
+LDAP configuration is in ```/data/config.json.php```
+
+Just add the following lines:
+```
+    "ldap": {
+        "host": "ldap://lldap_server:3890",
+        "dn": "uid=user_you_want_to_login_to_shaarli,ou=people,dc=example,dc=com"
```
nitnelave, line 9: Shouldn't this be `uid=%s,ou=people,dc=example,dc=com`?

PR #529, opened 2023-04-05
Use jq in CI to extract json deterministically

pixelrazor:
cut relies on the string being a fixed length, which is subject to change in the  future


nitnelave (approved):

nitnelave:
Nit: For next time, the commit message should be prefixed with `github:`

Issue #524, opened 2023-04-04
docker healthcheck fails when ldaps is enabled


leanderjanssen:
After enabling and configuring ldaps the Docker healthcheck is consistently failing with an `invalid dns name` error when it tries to resolve the `localhost` hostname.

Making a ldaps connection to the server works just fine.

Here is the debug output when manually running the healthcheck.
```
docker container exec -it lldap /app/lldap healthcheck --config-file /data/lldap_config.toml
```

```
2023-04-04T14:21:16.276794237+00:00  INFO     ｉ [info]: Starting healthchecks
2023-04-04T14:21:16.276901747+00:00  INFO     check_ldaps [ 121µs | 100.00% ]
2023-04-04T14:21:16.277022159+00:00  ERROR    ┕━ 🚨 [error]:  | error: invalid dns name
2023-04-04T14:21:16.277207994+00:00  DEBUG    🐛 [debug]: resolving host="localhost"
2023-04-04T14:21:16.277614454+00:00  DEBUG    🐛 [debug]: flushed 60 bytes
2023-04-04T14:21:16.276839569+00:00  INFO     check_ldap [ 405µs | 100.00% ]
[...]

PR #530, opened 2023-04-05
feat: support ec private keys

### LINKED ISSUE #525: Support ECKey as private key for lldaps
Caddy defaults to [ed25519 tls key type](https://caddyserver.com/docs/caddyfile/options#key-type). 
It would be cool if the certificates generated with standard configuration of Caddy would work with lldap as well.

Michsior14:
Tested locally using [self-signed certificate](https://www.scottbrady91.com/openssl/creating-elliptical-curve-keys-using-openssl).

Resolves #525

### DIFF
--- a/server/src/infra/ldap_server.rs
+++ b/server/src/infra/ldap_server.rs
@@ -94,7 +94,7 @@ where
 }
 
 fn read_private_key(key_file: &str) -> Result<PrivateKey> {
-    use rustls_pemfile::{pkcs8_private_keys, rsa_private_keys};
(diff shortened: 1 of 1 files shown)


nitnelave (approved):

leanderjanssen:
Does the lldap healthcheck also work with a self-signed cert?

nitnelave:
> Does the lldap healthcheck also work with a self-signed cert?

I think you know the answer https://github.com/lldap/lldap/issues/524 :) This PR doesn't change that fact.

nitnelave:
I'm waiting to merge this PR until we figure out the issues with pushing docker containers, we are running into trouble deploying to ghcr and dockerhub over several labels at once.

PR #532, opened 2023-04-05
github: Leverage metadata

martadinata666:
* Reduce mouthfull of image tagging.
* Tags
  * Debian
  ```
  vX-debian
  vX.Y-debian
  vX.Y.Z-debian
  stable-debian
  latest-debian
  ```
  * Alpine
  ```
  vX-alpine
  vX.Y-alpine
  vX.Y.Z-alpine
  stable-alpine
  latest-alpine
  latest
  stable
  ```
* Ensure database startup correctly on init test job
* Ensure database startup correctly on migration test job


.github/workflows/docker-build-static.yml
```diff
@@ -453,14 +463,18 @@ jobs:
...
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
-            type=sha
+            type=semver,pattern={{raw}}
+            type=semver,pattern=v{{major}}
+            type=semver,pattern=v{{major}}.{{minor}}
+            type=semver,pattern=v{{major}}.{{minor}}.{{patch}}
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
+            type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
pixelrazor, line 477:
```suggestion
            type=semver,pattern=v{{version}}
            type=semver,pattern=v{{major}}.{{minor}}
            type=semver,pattern=v{{major}}
            type=semver,pattern={{version}},value=stable
            type=raw,value=latest,enable={{is_default_branch}}
```
More consistent way of handling latest and [stable](https://github.com/marketplace/actions/docker-metadata-action#typesemver) IMO - stable could also be converted to raw and use `enable=${{ github.event_name == 'release' }}`
Also, i don't think we need teh raw symver, we can add `v` to the parsed version

.github/workflows/docker-build-static.yml
```diff
@@ -453,14 +463,18 @@ jobs:
           # list of Docker images to use as base name for tags
           images: |
             nitnelave/lldap
+            ghcr.io/lldap/lldap
```
pixelrazor, line 466:
```suggestion
            lldap/lldap
            ghcr.io/lldap/lldap
```

.github/workflows/docker-build-static.yml
```diff
@@ -425,6 +425,16 @@ jobs:
     needs: [build-ui, build-bin]
     name: Build Docker image
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container: ["debian","alpine"]
+        include:
+          - container: alpine
+            platforms: linux/amd64,linux/arm64
+            suffix: ""
```
pixelrazor, line 434:
```suggestion
            suffix: "-alpine"
```

.github/workflows/docker-build-static.yml
```diff
@@ -453,14 +463,18 @@ jobs:
           # list of Docker images to use as base name for tags
           images: |
             nitnelave/lldap
+            ghcr.io/lldap/lldap
+          flavor:
+            suffix= ${{ matrix.suffix }}
```
pixelrazor, line 468:
```suggestion
          flavor: |
            latest=false
            suffix=${{ matrix.suffix }}
```
latest to false since we manually handle it

.github/workflows/docker-build-static.yml
```diff
@@ -473,83 +487,47 @@ jobs:
...
-          platforms: linux/amd64,linux/arm64,linux/arm/v7
-          file: ./.github/workflows/Dockerfile.ci.debian
+          platforms: ${{ matrix.container.platforms }}
+          file: ./.github/workflows/Dockerfile.ci.${{ matrix.container }}
           tags: |
-            nitnelave/lldap:latest-debian
-            lldap/lldap:latest-debian
-            ghcr.io/${{ github.repository }}:latest-debian
-            ghcr.io/nitnelave/lldap:latest-debian
+                ${{ steps.meta.outputs.tags }}
```
pixelrazor, line 511: This can be just `tags: ${{ steps.meta.outputs.tags }}`

.github/workflows/docker-build-static.yml
```diff
@@ -473,83 +487,47 @@ jobs:
...
-            ghcr.io/nitnelave/lldap:stable, ghcr.io/nitnelave/lldap:stable-alpine, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-semantic }}, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}.${{ steps.slug.outputs.version-minor }}, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}.${{ steps.slug.outputs.version-minor }}.${{ steps.slug.outputs.version-patch }}, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-semantic }}-alpine, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}-alpine, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}-alpine.${{ steps.slug.outputs.version-minor }}-alpine, ghcr.io/nitnelave/lldap:v${{ steps.slug.outputs.version-major }}.${{ steps.slug.outputs.version-minor }}.${{ steps.slug.outputs.version-patch }}-alpine
-          cache-from: type=gha,mode=max
-          cache-to: type=gha,mode=max
-
-      - name: Build and push release debian
+      - name: Build and push release
```
pixelrazor, line 519: Hear me out - do we need this step? Wouldn't the previous build and push handle this correctly, since the metadata action will supply the appropriate tags when it's triggered by release?

pixelrazor (commented):
Okay, so I have an idea. I suggested the -alpine suffix, but we still need the tags without suffix. I'm thinking we add a variable to the matrix: `base: true` for alpine and `base: false` for debian. Then we duplicate tags and explicitly set suffix to empty and  enable based off that:
```yaml
            type=semver,pattern=v{{version}},suffix=,enable=${{ matrix.base }}
            type=semver,pattern=v{{major}}.{{minor}},suffix=,enable=${{ matrix.base }}
            type=semver,pattern=v{{major}},suffix=,enable=${{ matrix.base }}
            type=semver,pattern={{version}},value=stable,suffix=,enable=${{ matrix.base }}
            type=raw,value=latest,suffix=,enable=${{ github.ref == format('refs/heads/{0}', 'master') && matrix.base }}
```
(i don't know if the `{{is_default_branch}}` templating from the action works in expressions?)

pixelrazor (commented):
We can delete the `slug` step - it's unused with this change

.github/workflows/docker-build-static.yml
martadinata666: Already tried this `type=semver,pattern={{version}},value=stable`, the `value` must conform semver `vX.Y.Z`

This maybe useful,  `type=raw,value=latest,enable={{is_default_branch}}` checking which one is default branch, in this case `main` or future change it will follow.

```
            type=semver,pattern=v{{version}}
            type=semver,pattern=v{{major}}
            type=semver,pattern=v{{major}}.{{minor}}
            type=semver,pattern=v{{major}}.{{minor}}.{{patch}}
            type=raw,value=latest,enable={{ is_default_branch }}
            type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
This should be sufficient?

.github/workflows/docker-build-static.yml
martadinata666: do we already had `lldap` on docker? as `lldap/lldap` will lead to `index.docker.io/lldap/lldap`

.github/workflows/docker-build-static.yml
pixelrazor: EDIT: please change ` type=semver,pattern={{version}},value=stable` to the propsed alternative - i misunderstood how value was going to work in that context

.github/workflows/docker-build-static.yml
martadinata666: If `-alpine` enabled , need metadata tag to only produce `v{{version}}` only.

.github/workflows/docker-build-static.yml
pixelrazor: docker pull works for me with it

.github/workflows/docker-build-static.yml
pixelrazor: your alternative with the ref also works -roll with either

.github/workflows/docker-build-static.yml
pixelrazor: see the main comment i left with this batch - we'd need some extra stuff to get both plain and -alpine tags

.github/workflows/docker-build-static.yml
pixelrazor: I think we still want nitnelave/lldap since that's waht a LOT of people target still

.github/workflows/docker-build-static.yml
```diff
@@ -462,19 +462,19 @@ jobs:
...
+            lldap/lldap
             ghcr.io/lldap/lldap
           flavor: |
             latest=false
             suffix= ${{ matrix.suffix }}
           # generate Docker tags based on the following events/attributes
           tags: |
             type=ref,event=pr
-            type=semver,pattern={{raw}}
+            type=semver,pattern=v{{version}}
```
pixelrazor, line 473: pelase remove major,minor,patch version below. {{version}} covers that

.github/workflows/docker-build-static.yml
martadinata666:
```
type=ref,event=pr
type=semver,pattern=v{{version}}
type=semver,pattern=v{{major}}
type=semver,pattern=v{{major}}.{{minor}}
type=raw,value=latest,enable={{ is_default_branch }},suffix=${{ matrix.noflavour }}
type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }},suffix=${{ matrix.noflavour }}
```

So basically on `latest` and `stable` override the suffix, become `:latest` `:latest-debian` and `:stable` `:stable-debian`

.github/workflows/docker-build-static.yml
```diff
@@ -453,26 +484,39 @@ jobs:
...
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Login to LLDAP Docker Hub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_LLDAP_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_LLDAP_TOKEN }}
+
```
martadinata666, line 519: Prepare to login to both repo for `lldap/lldap` and `nitnelave/lldap`.

nitnelave: `lldap` is an organization on DockerHub too, so you can't log in as it. I *think* that being logged in as nitnelave is enough (we already pushed an image to lldap/lldap).

martadinata666: On update repo part
```

      - name: Update repo description
        if: github.event_name != 'pull_request'
        uses: peter-evans/dockerhub-description@v3
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PASSWORD }}
          repository: nitnelave/lldap

      - name: Update lldap repo description
        if: github.event_name != 'pull_request'
        uses: peter-evans/dockerhub-description@v3
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PASSWORD }}
          repository: lldap/lldap
```
Is this working correctly? Same secrets?

.github/workflows/docker-build-static.yml
```diff
@@ -453,20 +484,26 @@ jobs:
...
           images: |
             nitnelave/lldap
-          # generate Docker tags based on the following events/attributes
+            lldap/lldap
+            ghcr.io/lldap/lldap
+          # generate Docker tags based on the following events/attributes, with -alpine or -debian
+          # vX-debian / -alpine
+          # vX.Y-debian / -alpine
+          # vX.Y.Z-debian / -alpine
+          # Suffix override on latest and stable tag, stable stable-debian, latest, latest-debian
```
pixelrazor, line 493: We still need to have all these tags on alpine:
latest-alpine
stable-alpine
vX
vX.Y
v.X.Y.Z

Those are all tags that exist today and i don't think we should be changing that, at least not in the scope of this work. That's what i proposed in my last review, but  you can do you own take on whatever accomplishes it:

Okay, so I have an idea. I suggested the -alpine suffix, but we still need the tags without suffix. I'm thinking we add a variable to the matrix: `dobase: true` for alpine and `dobase: false` for debian. Then we duplicate tags and explicitly set suffix to empty and enable based off that:
```
            type=semver,pattern=v{{version}},suffix=,enable=${{ matrix.dobase }}
            type=semver,pattern=v{{major}}.{{minor}},suffix=,enable=${{ matrix.dobase }}
            type=semver,pattern=v{{major}},suffix=,enable=${{ matrix.dobase }}
            type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') && matrix.dobase }},suffix=,enable=${{ matrix.dobase }}
            type=raw,value=latest,suffix=,enable=${{ github.ref == format('refs/heads/{0}', 'main') && matrix.dobase }}
```

martadinata666: the `stable` and `latest` pretty much alike you suggested, just need to add for `vX` `vX.Y` `vX.Y.Z`.

pixelrazor: With this approach, the debian matrix will generate duplicate tags. That might be a problem, it might not; this action or the docker build might handle it just fine. I used a bool in my suggegstion to enable or disable the extra tags to avoid this - if the build is happy though, i won't oobject. One last nit here - i don';t think this generates `latest-alpine` or `stable-alpine`

martadinata666: ah skipped that, and yeah `:latest` and `:stable` will duplicate. But no more because of some magic 🪄. 🤣  

Joke aside, essentially we should pass the `definition tags` to metadata generation step.

.github/workflows/docker-build-static.yml
```diff
@@ -425,6 +443,33 @@ jobs:
...
+        container: ["debian","alpine"]
+        include:
+          - container: alpine
+            platforms: linux/amd64,linux/arm64
+            suffix: "-alpine"
+            tags: |
+                  type=ref,event=pr
+                  type=semver,pattern=v{{version}}
+                  type=semver,pattern=v{{major}}
+                  type=semver,pattern=v{{major}}.{{minor}}
```
pixelrazor, line 457:
```suggestion
                  type=semver,pattern=v{{version}}
                  type=semver,pattern=v{{major}}
                  type=semver,pattern=v{{major}}.{{minor}}
                  type=semver,pattern=v{{version}},suffix=
                  type=semver,pattern=v{{major}},suffix=
                  type=semver,pattern=v{{major}}.{{minor}},suffix=
```

I believe we still want version tags without base OS too.

Very nice idea with moving the tags to the matrix! An further change you COULD do (feel free to ignore if yo uwant):

Remove the suffix from the matrix, and just hardcode the suffix into the list of tags here and below. Then you can remove the suffix from the flavor field in the metadata step. My thinking here is the matrix is providing the list of tags anyways - why continue to split that logic between the matrix definition and the metadata step

.github/workflows/docker-build-static.yml
```diff
@@ -425,6 +443,36 @@ jobs:
     needs: [build-ui, build-bin]
     name: Build Docker image
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container: ["debian","alpine"]
+        include:
+          - container: alpine
+            platforms: linux/amd64,linux/arm64
+            suffix: "-alpine"
```
pixelrazor, line 452: remove suffix, it's dead after your latest commit

pixelrazor (approved):

nitnelave (approved):
Very nice!

nitnelave:
It needs to be rebased to solve the conflicts, but thanks a lot for this PR.

martadinata666:
please push docker image correctly 🤞🏼

PR #549, opened 2023-04-12
server: WIP implementation of server key from rng seed

nitnelave:


pixelrazor (approved):
needs documentation in the toml file as well

PR #552, opened 2023-04-13
server: enforce email and uuid unicity

### LINKED ISSUE #537: Security issue: Users can set their own arbitrary email addresses
Users are able to log in to lldap and change their own email address to anything.
This presents two security concerns:
1 - Since many applications use email address as a unique identifier, this can allow impersonating the application accounts of other users.
2 - An attacker could set their email to a target's address and then abuse the password reset feature to flood their inbox

The simplest, most secure fix is to add a config option to prevent users from modifying their own email address.

A more sophisticated fix would be to allow users to use a process similar to the password-reset flow to update their email, and to rate-limit the number of email updates which can be requested per-account.

nitnelave:
Fixes #537 and #466 .

### DIFF
--- a/server/src/domain/sql_migrations.rs
+++ b/server/src/domain/sql_migrations.rs
@@ -2,14 +2,13 @@ use crate::domain::{
     sql_tables::{DbConnection, SchemaVersion},
     types::{GroupId, UserId, Uuid},
 };
+use anyhow::Context;
 use sea_orm::{
-    sea_query::{self, ColumnDef, Expr, ForeignKey, ForeignKeyAction, Query, Table, Value},
+    sea_query::{self, ColumnDef, Expr, ForeignKey, ForeignKeyAction, Index, Query, Table, Value},
     ConnectionTrait, FromQueryResult, Iden, Statement, TransactionTrait,
 };
 use serde::{Deserialize, Serialize};
 use tracing::{info, instrument, warn};
-
-use super::sql_tables::LAST_SCHEMA_VERSION;
 
 #[derive(Iden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum Users {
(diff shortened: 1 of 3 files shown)


server/src/domain/sql_migrations.rs
```diff
@@ -460,30 +459,81 @@ async fn migrate_to_v3(pool: &DbConnection) -> anyhow::Result<()> {
     Ok(())
 }
 
+async fn migrate_to_v4(pool: &DbConnection) -> anyhow::Result<()> {
+    let builder = pool.get_database_backend();
+    // Make emails and UUIDs unique.
+    pool.execute(
```
nitnelave, line 465: Note that I can't use `replace_column`, since you can't create column with "unique" constraints in sqlite. You get the same effect by creating a unique index, though.

server/src/domain/sql_migrations.rs
```diff
@@ -460,30 +459,81 @@ async fn migrate_to_v3(pool: &DbConnection) -> anyhow::Result<()> {
     Ok(())
 }
 
+async fn migrate_to_v4(pool: &DbConnection) -> anyhow::Result<()> {
```
pixelrazor, line 462: From what i think, this will fail if two users already exist with the same email, correct? It might be worth making note, and making sure this is labelled as a breaking change?

pixelrazor (commented):
What do the errors look like when attempting to create a user with an existing email or updating email to a used one? We should make sure the UI feedback is clear and actionable.

Totally not needed, but part of me thinks validation on the email field (similar to HIBP with gloo-timers) would be nice to have - once again, totally not needed, permitting we give a good response to the user when edit/create fails

server/src/domain/sql_migrations.rs
nitnelave: Yeah, that's what the `.context` in the body means: it adds some context to the error message, which will show up in the logs.

And yes, it'll be a breaking change.

PR #558, opened 2023-04-14
Add OPNsense example config.

tylerpace:
Added an example configuration for using LLDAP in OPNsense.

### DIFF
--- a/example_configs/opnsense.md
+++ b/example_configs/opnsense.md
@@ -1,0 +1,117 @@
+# Configuration for OPNsense
+
+## Create a LDAP Server
+
+- Login to OPNsense
+- Navigate to: `System > Access > Servers`
+- Create a new server by clicking on the `+` icon
+
+## Server Config
+
+- Descriptive Name: `A Descriptive Name`
+- Type: `LDAP`
+- Hostname or IP address: `Hostname or IP for your LLDAP host`
+- Port value: `Your LLDAP port`
+  - Default: `3890`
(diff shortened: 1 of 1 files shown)


example_configs/opnsense.md
```diff
@@ -0,0 +1,117 @@
...
+
+It is recommended that you create a separate user account (e.g, `bind_user`) instead of `admin` for sharing Bind credentials with other services. The `bind_user` should be a member of the `lldap_strict_readonly` group to limit access to your LDAP configuration in LLDAP.
+
+#### Password
+
+```
+xxx
+```
+
+Use the password that you set for the user specified in the User DN field. Your password was likely set via:
```
nitnelave, line 39: I think this is redundant, most people know what the password for the admin account is. Moreover, if they changed it since, it will be different from the env variable. And lastly, if they follow your advice above of creating a bind_user, the password will have nothing to do with the env variable.

In short, I think something along the lines of "The password for the user above" is enough.

nitnelave (approved):

example_configs/opnsense.md
tylerpace: Done!

nitnelave:
Thank you everyone for the contribution!
This feature is now available in the latest release, [v0.5.0](https://github.com/lldap/lldap/releases/tag/v0.5.0).
You can support LLDAP by starring our repo, contributing some configuration examples and becoming a sponsor.

PR #564, opened 2023-04-25
fixed check of memberrole

Hobbabobba:


nitnelave (approved):
Is this really needed? It should be equivalent. But I'll trust you if you say it works with cn but not with uid

Hobbabobba:
It is necessary, unfortunately. The old settings works for any user in any group. The new setting works as intended.

PR #565, opened 2023-04-27
correct env var names in docker template toml

bendlas:
found out about `LLDAP_KEY_FILE` by trial and error, figured `LLDAP_KEY_SEED` by inference

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -95,13 +95,13 @@ database_url = "sqlite:///data/users.db?mode=rwc"
 ## Randomly generated on first run if it doesn't exist.
 ## Alternatively, you can use key_seed to override this instead of relying on
 ## a file.
-## Env variable: LLDAP_SERVER_KEY_FILE
+## Env variable: LLDAP_KEY_FILE
 key_file = "/data/private_key"
 
 ## Seed to generate the server private key, see key_file above.
 ## This can be any random string, the recommendation is that it's at least 12
 ## characters long.
-## Env variable: LLDAP_SERVER_KEY_SEED
+## Env variable: LLDAP_KEY_SEED
 #key_seed = "RanD0m STR1ng"
 
 ## Ignored attributes.


lldap_config.docker_template.toml
```diff
@@ -95,13 +95,13 @@ database_url = "sqlite:///data/users.db?mode=rwc"
 ## Randomly generated on first run if it doesn't exist.
 ## Alternatively, you can use key_seed to override this instead of relying on
 ## a file.
-## Env variable: LLDAP_SERVER_KEY_FILE
+## Env variable: LLDAP_KEY_FILE
```
nitnelave, line 98: I don't think this one is correct. The other one yes, but not this one

nitnelave (changes requested):

lldap_config.docker_template.toml
bendlas: wait, something is weird:

When I `lldap run` with `LLDAP_SERVER_KEY_FILE=/var/lib/lldap/private_key`, I get `Error: Could not open `/var/lib/lldap/private_key` from config value `server_key_file`: stream did not contain valid UTF-8 in `LLDAP_` environment variable(s)`

When I `lldap run` with just a postgres connection string, I get `Error: Could not write the generated server setup to file 'server_key'; Read-only file system (os error 30)`. This would be consistent with trying to find a file in working directory, which is read-only in my setup.

But when I `lldap run` with `LLDAP_KEY_FILE=/var/lib/lldap/private_key`, it starts and creates `/var/lib/lldap/private_key`.

I'm on commit 5ad63d31d360d3706c153409f28deb99c446adf8. Any idea what could cause the inconsistent errors (and accidental success) here?

bendlas: btw, the success with `LLDAP_KEY_FILE=/var/lib/lldap/private_key` is the reason I opened this PR in the first place

pixelrazor: https://github.com/lldap/lldap/blob/22ae2c71241def35a7c8f203d06bb23dc62a37de/server/src/infra/configuration.rs#L94 shows the configuration using `key_file`, which seems to imply that this is actually correct?

nitnelave: I'm a bit confused as to what's happening, but for what it's worth I think the behavior with `LLDAP_KEY_FILE` is better than with `LLDAP_SERVER_KEY_FILE`: I just tried with a non-existing file, and it crashed on startup, whereas the `LLDAP_KEY_FILE` correctly created the file. I'm not entirely sure why, though... I'll merge this, and make a mental note (that I'll probably forget) to look at it closer when I have time.

nitnelave (approved):

nitnelave:
Thank you everyone for the contribution!
This feature is now available in the latest release, [v0.5.0](https://github.com/lldap/lldap/releases/tag/v0.5.0).
You can support LLDAP by starring our repo, contributing some configuration examples and becoming a sponsor.

PR #567, opened 2023-05-02
tests: allow dead code in common module

nitnelave:
We're running afoul of https://github.com/rust-lang/rust/issues/46379, where each test is compiled independently, so any test that doesn't use every helper method triggers a dead code warning.


pixelrazor (approved):
Well that's odd - it's in line with my reading of integration tests though.

Issue #67, opened 2021-10-21
Feature Request: Arbitrary and Duplicate Text Fields


dwmunster:
# Overview

As an LLDAP admin, I would like to be able to define arbitrary text fields that may be repeated to extend the user object schema. 

# Details

I would like to see the following:

* the ability to define arbitrary fields with string data
* the ability to restrict the ability for users to change these fields
* the ability to have a list or repeated values for the field

# Context

I am interested in using LLDAP as a user DB for my mail server and to replace my OpenLDAP instance with something much more manageable. However, my users may have more than one email address assigned to them. Currently, I have that implemented with a repeated `mailalias` attribute. For example, my user looks something like the following:

```
dn: uid=user,ou=people,dc=domain,dc=tld
objectClass: mailAccount
objectClass: person
objectClass: uidObject
objectClass: top
uid: user
mail: user@domain.tld
mailenable: TRUE
sn: SN
cn: CN
mailalias: user@domain1.tld
mailalias: user@domain2.tld
mailalias: user2@domain1.tld
```
Restricting the ability for users to change the `mailalias` attributes is also crucial to prevent users from impersonating other users and sending mail from arbitrary addresses.

Other use cases include:
- SSH Public Keys
- Home directories

Task breakdown:
 - [x] Prepare DB and migration
 - [x] Create Schema data structure + default schema
 - [x] Migrate first and last name to attributes
 - [x] Implement read in the domain
 - [x] Implement read in graphql
 - [x] Implement read in LDAP 
 - [x] Implement write in domain
 - [x] Implement write in graphql
 - [ ] Implement write in LDAP
 - [x] Implement front-end read/write
 - [x] Implement schema manipulation features
   - [x] domain
   - [x] graphql
   - [x] frontend
 - [ ] Implement aliases
   - [ ] Query from LDAP
   - [ ] create/delete/query domain
   - [ ] create/delete/query graphql
   - [ ] create/delete/display frontend
 - [x] Custom LDAP object classes
   - [x] backend
   - [x] frontend

nitnelave:
Here's an RFC for this feature: https://docs.google.com/document/d/1BdizgwX6Pvy5hqH8_HzSg3LCa66tQSDhrTsdhF80ySY/edit?usp=drivesdk

Samonitari:
Wow!

I was searching for [ldap-ui](https://github.com/dnknth/ldap-ui) when I stumbled upon this project, which seems pretty awesome.
[...]

PR #574, opened 2023-05-08
Add Ejabberd

poVoq:
Basic auth only for now. Will add more once I figure out how to make it work.


example_configs/ejabberd.md
```diff
@@ -0,0 +1,30 @@
...
+[Main documentation here.](https://docs.ejabberd.im/admin/configuration/ldap/)
+
+For simple user auth add this to main ejabberd.yml:
+
+```
+host_config:
+  xmpp.example.org:
+    auth_method: [ldap]
+    ldap_servers:
+      - 127.0.0.1
```
nitnelave, line 12: Let's use "<your ldap host>" here, since it's not always localhost

example_configs/ejabberd.md
```diff
@@ -0,0 +1,30 @@
...
+    ldap_port: 3890
+    ldap_uids:
+      - uid
+    ldap_rootdn: "uid=admin,ou=people,dc=example,dc=org"
```
nitnelave, line 16: You could mention that any member of "lldap_admin" or "lldap_readonly" would work here (that still allows to change passwords)

example_configs/ejabberd.md
```diff
@@ -0,0 +1,30 @@
...
+    ldap_password: "secret"
+    ldap_base: "ou=people,dc=example,dc=org"
+```
+
+## vCard from LDAP
+Theoretically possible, [see the documentation.](https://docs.ejabberd.im/admin/configuration/ldap/#vcard-in-ldap)
+
+TODO
```
nitnelave, line 24: The birthday is not supported yet. If it's required, then that's blocking on #67

nitnelave (changes requested):

example_configs/ejabberd.md
poVoq: Thats all optional. I actually only plan to get the photo and display name from LDAP. But good to know that additional values are worked on.

example_configs/ejabberd.md
poVoq: I think having an actual functional value there is easier to understand and modify than something that doesn't explain what exactly to put there, but I will add a comment to explain that this is the ip or hostname of the LLDAP server.

PR #578, opened 2023-05-12
github: create actions to publish crates

nitnelave:


.github/workflows/rust.yml
```diff
@@ -112,3 +112,39 @@ jobs:
...
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  publish-crates:
+    name: Publish on crates.io
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' || (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release' }}
+    needs: pre_job
+    steps:
+      - uses: actions/checkout@v3.5.2
+      - uses: actions-rs/toolchain@v1
```
nitnelave, line 122: Note: we might want (in a future PR) to migrate our rust installation to this action instead.

fgardt: actually you probably don't want to use `actions-rs` because its [unmaintained](https://github.com/actions-rs/toolchain/issues/216)

nitnelave:
We're hitting https://github.com/katyo/publish-crates/issues/660 so I had to get rid of the wildcard versions.

PR #584, opened 2023-05-16
github: enable armv7 musl build

martadinata666:
~~Building with gnu compiler~~
* [x] ~~Can remove musl-gcc from development image?~~
* [x] Build armv7? Success
* [x] ~~Set image to alpine only?~~ Keep all
* [x] Some cleanup

Issue after change
* [x] Some sqlite3 missing? Incompabily linking with gnu gcc. (use musl-gcc)
* [x] ~~musl-cross armv7l need more older musl [stackoverflow thread](https://stackoverflow.com/questions/61934997/undefined-reference-to-stat-time64-when-cross-compiling-rust-project-on-mu),~~ this is rust issue, confirmed with rust nightly 1.71 (musl lib upgraded to 1.2.3)
* [x] Due sqlite3, can't remove musl compiler (sqlite3 blackmagic)
```
[...]

PR #586, opened 2023-05-17
dev image: prep for 1.71

martadinata666:
Tasks:
* [x] wait rust 1.71 stable release
* [x] Full musl, remove arm64 and armhf gnu deps
* [x] Fix cc and linker to target musl gcc binaries

Deprecated warning:
* Warning: the following packages contain code that will be rejected by a future version of Rust: nom v2.2.1

Related:
* https://github.com/lldap/lldap/pull/584

PS: Locally tested that armv7 musleabihf compiled successfully

### DIFF
--- a/.github/workflows/Dockerfile.dev
+++ b/.github/workflows/Dockerfile.dev
@@ -1,45 +1,55 @@
 # Keep tracking base image
-FROM rust:1.66-slim-bullseye
+FROM rust:1.71-slim-bookworm
 
 # Set needed env path
-ENV PATH="/opt/aarch64-linux-musl-cross/:/opt/aarch64-linux-musl-cross/bin/:/opt/x86_64-linux-musl-cross/:/opt/x86_64-linux-musl-cross/bin/:$PATH"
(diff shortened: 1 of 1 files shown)


nitnelave:
Why do you want to update to 1.68 or 1.71? Any specific reason?

martadinata666:
> Why do you want to update to 1.68 or 1.71? Any specific reason?

oh, it still 1.68? my bad, I will update to 1.71 later.

The specific reason for rust 1.71 is because https://blog.rust-lang.org/2023/05/09/Updating-musl-targets.html which means the old `time64` thing and maybe other few things that broke armv7-musl compilation are hopefully solved.

martadinata666:
As I locally tested locally that armv7l musl build successfully, should we drop the gnu build all together?

nitnelave (approved):

nitnelave:
> As I locally tested locally that armv7l musl build successfully, should we drop the gnu build all together?

Sure, let's clean up as much as we can.

PR #598, opened 2023-06-08
Update jellyfin.md example

gdunstone:
The LdapAdminBaseDN either doesn't work properly or is used incorrectly here.

This change will make it work.

see: 
- https://github.com/jellyfin/jellyfin-plugin-ldapauth/issues/145

### DIFF
--- a/example_configs/jellyfin.md
+++ b/example_configs/jellyfin.md
@@ -37,7 +37,7 @@ Otherwise, just use:
 ```
 ### Admin Base DN
 
-The DN to search for your admins. If your admins are also users use:
+The DN to search for your admins.
 ```
 ou=people,dc=example,dc=com
 ```
@@ -49,6 +49,7 @@ that), use:
 ```
 (memberof=cn=media_admin,ou=groups,dc=example,dc=com)
 ```
+Bear in mind that admins must also be a member of the users group if you use one.
 
 Otherwise, you can use LLDAP's admin group:
 ```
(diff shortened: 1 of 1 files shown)


example_configs/jellyfin.md
```diff
@@ -37,9 +37,9 @@ Otherwise, just use:
 ```
 ### Admin Base DN
 
-The DN of your admin group. If you have `media_admin` as your group you would use:
+The DN to search for your admins. If your admins are also users use:
```
nitnelave, line 40: Erm, the admins are all users, no?

nitnelave (approved):
Nice catch!

martadinata666:
So after playing a bit, this is how it works `cmiiw`. The admin was filtered twice, first filtered as normal user, then filtered as admin. So the `admin must be on two groups`, `normal user group` and `admin group`. If the admin not in normal user group it won't login. At least that's what happen in my jellfyin, I hope someone can reconfirm this, also add additional description about this condition.

example_configs/jellyfin.md
gdunstone: Well technically yes, but not strictly. You could have a user who is a member of the admin group but not a member of the user group.

You could also have no user group set at all and only designate an admin group.

nitnelave: I mean, admins are LDAP users, and all users are in that base DN, so you'll always need this value.

Also, regarding what @xavierus666 said, I think even admins have to pass a user check when logging in.

gdunstone: Fair. I have updated the language here.

nitnelave (approved):

PR #603, opened 2023-06-13
server: prepare DB schema for user attributes

nitnelave:
First step of #67.


fgardt (approved):
as mentioned this is a big chunk of changes but it looks pretty good so far! very promising.

pixelrazor (approved):
Awesome stuff!

PR #610, opened 2023-06-22
enhancement: Emails do not have a message_id

### LINKED ISSUE #608: Emails do not have a message_id
I noticed that password-reset-mails sent from my system have a higher spam score as expected.
This is because the mails are missing the `message-id header`.

Usually it is up to the originating client to add this header.
But many email-providers add it automatically if it is missing, because those mails would very likely get flagged as spam and lower their trust score.

Anyway it would be a good thing to not rely on the mailserver to "repair" the mails for us.

I can not implement it myself but here is how another project which is also using `lettre` solved it.
Maybe someone can get a PR going.
https://github.com/dani-garcia/vaultwarden/blob/1.28.1/src/mail.rs#L594-L599

References:
https://en.wikipedia.org/wiki/Message-ID
https://docs.rs/lettre/latest/lettre/message/header/struct.MessageId.html

tarunsamanta2k20:
Closes #608 

**Issue Description:**

I noticed that password-reset-mails sent from my system have a higher spam score as expected.
This is because the mails are missing the message-id header.

Usually it is up to the originating client to add this header.
But many email-providers add it automatically if it is missing, because those mails would very likely get flagged as spam and lower their trust score.

Anyway it would be a good thing to not rely on the mailserver to "repair" the mails for us.

I can not implement it myself but here is how another project which is also using lettre solved it.
Maybe someone can get a PR going.

### DIFF
--- a/server/src/infra/mail.rs
+++ b/server/src/infra/mail.rs
@@ -17,7 +17,7 @@ async fn send_email(to: Mailbox, subject: &str, body: String, options: &MailOpti
         &to, &from, &options.user, &options.server, options.port
     );
     let email = Message::builder()
-        .message_id(Some(format!("<{}@{}>", crate::util::get_uuid(), smtp_from.split('@').collect::<Vec<&str>>()[1])))
+        .message_id(Some(format!("<{}@{}>", uuid::Uuid::new_v4().to_string(), from.split('@').collect::<Vec<&str>>()[1])))
         .from(from)
         .reply_to(reply_to)
         .to(to)
--- a/server/src/infra/mail.rs
+++ b/server/src/infra/mail.rs
@@ -17,7 +17,7 @@ async fn send_email(to: Mailbox, subject: &str, body: String, options: &MailOpti
         &to, &from, &options.user, &options.server, options.port
     );
     let email = Message::builder()
-        .message_id(Some(format!("<{}@{}>", uuid::Uuid::new_v4().to_string(), from.split('@').collect::<Vec<&str>>()[1])))
+        .message_id(Some(format!("<{}@{}>", uuid::Uuid::new_v4().to_string(), from.to_string().split('@').collect::<Vec<&str>>()[1])))
         .from(from)
         .reply_to(reply_to)
         .to(to)
--- a/server/src/infra/mail.rs
+++ b/server/src/infra/mail.rs
@@ -17,6 +17,7 @@ async fn send_email(to: Mailbox, subject: &str, body: String, options: &MailOpti
         &to, &from, &options.user, &options.server, options.port
     );
     let email = Message::builder()
+        .message_id(Some(format!("<{}@{}>", crate::util::get_uuid(), smtp_from.split('@').collect::<Vec<&str>>()[1])))
         .from(from)
         .reply_to(reply_to)
         .to(to)


server/src/infra/mail.rs
```diff
@@ -17,6 +17,7 @@ async fn send_email(to: Mailbox, subject: &str, body: String, options: &MailOpti
         &to, &from, &options.user, &options.server, options.port
     );
     let email = Message::builder()
+        .message_id(Some(format!("<{}@{}>", crate::util::get_uuid(), smtp_from.split('@').collect::<Vec<&str>>()[1])))
```
nitnelave, line 20: I don't think we have a `crate::util::get_uuid()` function, and there is no local variable `smtp_from`.

Could you try compiling and testing locally?

nitnelave (changes requested):

server/src/infra/mail.rs
tarunsamanta2k20: **how to run and tet locally? **

nitnelave: You can check the section on building from source in the readme: https://github.com/lldap/lldap#from-source

Once you have it built, you can run tests with `cargo test`.

tarunsamanta2k20: @nitnelave why am i getting split() method not available? it is an inbuilt method.help me to resolve it.

nitnelave: It's a built-in method for which type? And what type are you calling it on?

tarunsamanta2k20: I am calling it like thet from.split('@').

nitnelave: Yes, I can see that. What is the type of `from` ?

tarunsamanta2k20: I think I have to convert it to string first then call the split(). Right?

nitnelave: That would be one way. But try answering the question first: what is the type of `from`?

tarunsamanta2k20: I think message type.

nitnelave: Nope. To make sure that you know which type it is, please send me a link to the definition of the type of `from`, or the docs for its type.

tarunsamanta2k20: Option<Mailbox>

nitnelave: That would be the type of `options.from`. The type of `from` is just `Mailbox`. Now, using the docs, can you find a different way to extract the domain from the address?

tarunsamanta2k20: Allow me some time i will let you know.

Masgalor:
Perhaps the thought does not make sense, but maybe we can not entirely rely on the sender address.
It is an optional field with no default value, so it might be empty.

Would it be a good idea to check if we did not get anything useful from `from` and use some fallback in that case?
Maybe the domain in `http_server` would be a good fallback, it has a default value and if not configured properly the password reset would not work anyway.

nitnelave:
> maybe we can not entirely rely on the sender address. It is an optional field with no default value, so it might be empty.

True, but for the purpose of simply message ID uniqueness, having "@lldap" as the domain name (the fallback) is probably enough.

Masgalor:
You are right, that would be within the requirements.
Just to add some context [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322) strongly recommends to use a domain or IP.

```
The message identifier (msg-id) itself MUST be a globally unique
   identifier for a message.  The generator of the message identifier
   MUST guarantee that the msg-id is unique.  There are several
   algorithms that can be used to accomplish this.  Since the msg-id has
   a similar syntax to addr-spec (identical except that quoted strings,
   comments, and folding white space are not allowed), a good method is
   to put the domain name (or a domain literal IP address) of the host
   on which the message identifier was created on the right-hand side of
   the "@" (since domain names and IP addresses are normally unique),
   and put a combination of the current absolute date and time along
   with some other currently unique (perhaps sequential) identifier
   available on the system (for example, a process id number) on the
   left-hand side.  Though other algorithms will work, it is RECOMMENDED
   that the right-hand side contain some domain identifier (either of
   the host itself or otherwise) such that the generator of the message
   identifier can guarantee the uniqueness of the left-hand side within
   the scope of that domain.
```

In my head it was not so much hassle to use `http_server` strip the protocol and use that.
But on the other hand, I do not know enough to do this myself, so it might be more complicated than I imagine.

nitnelave:
Fair enough. Getting the domain or IP from `http_server` would probably be ever-so-slightly more accurate here.

Although, coming back to the uniqueness point and the recommendation from the RFC, we probably want to use UUID v1 with a timestamp, rather than the fully-random v4 (even though collision chances are astronomically low).

tarunsamanta2k20:
@nitnelave what should i put in place of **"from"**?

nitnelave:
@tarunsamanta2k20 Just a reminder: open source authors are not here to fix your bugs or teach you how to write code. If you want to learn Rust, a very good first step is to work through the [Rust book](https://doc.rust-lang.org/book/) and build a small project yourself, so you get acquainted with the language.

I'm going to close this PR as it not only has problems of style (code not formatted), form (code not using the right APIs to achieve its goals), and content (wrong semantics), all in the span of a single line of code.

Once you feel more confident in Rust, I invite you to come back, have another look at the discussion here, and submit a PR that is:
  - working
  - tested (you reproduced the issue without the patch, and the same steps do not reproduce the issue with the patch, along with some output of your test)
  - properly formatted
  - taking this discussion into account.

tarunsamanta2k20:
@nitnelave please reopen it, I'll modify it. as a contributor i just asked for your suggestion. Better to finalize it from yourself rather than making unnecessary commits.

PR #615, opened 2023-06-28
Use attribute schema at the LDAP/GraphQL layers

nitnelave:
Step 2 in #67 implementation.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -128,7 +128,7 @@ features = ["dangerous_configuration"]
 
 [dev-dependencies]
 assert_cmd = "2.0"
-mockall = "0.11"
+mockall = "0.11.4"
 nix = "0.26.2"
 
 [dev-dependencies.graphql_client]
--- a/server/src/domain/ldap/user.rs
+++ b/server/src/domain/ldap/user.rs
@@ -5,11 +5,11 @@ use ldap3_proto::{
 use tracing::{debug, instrument, warn};
 
 use crate::domain::{
-    handler::{UserListerBackendHandler, UserRequestFilter},
+    handler::{Schema, UserListerBackendHandler, UserRequestFilter},
     ldap::{
         error::{LdapError, LdapResult},
         utils::{
-            expand_attribute_wildcards, get_group_id_from_distinguished_name,
+            expand_attribute_wildcards, get_custom_attribute, get_group_id_from_distinguished_name,
             get_user_id_from_distinguished_name, map_user_field, LdapInfo, UserFieldType,
         },
     },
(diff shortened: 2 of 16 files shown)


server/Cargo.toml
```diff
@@ -128,7 +128,7 @@ features = ["dangerous_configuration"]
 
 [dev-dependencies]
 assert_cmd = "2.0"
-mockall = "0.11"
+mockall = "0.11.4"
```
nitnelave, line 131: Without this update, cargo crashes :/

server/src/domain/ldap/user.rs
```diff
@@ -36,9 +37,13 @@ pub fn get_user_attribute(
         "uid" | "user_id" | "id" => vec![user.user_id.to_string().into_bytes()],
         "entryuuid" | "uuid" => vec![user.uuid.to_string().into_bytes()],
         "mail" | "email" => vec![user.email.clone().into_bytes()],
-        "givenname" | "first_name" | "firstname" => vec![user.first_name.clone()?.into_bytes()],
-        "sn" | "last_name" | "lastname" => vec![user.last_name.clone()?.into_bytes()],
-        "jpegphoto" | "avatar" => vec![user.avatar.clone()?.into_bytes()],
+        "givenname" | "first_name" | "firstname" => {
```
nitnelave, line 40: This will eventually be replaced by some dynamic lookup based on the attribute name and aliases. Hardcoding for now as a smallest step.

server/src/domain/types.rs
```diff
@@ -378,16 +392,20 @@ impl IntoActiveValue<Serialized> for JpegPhoto {
...
+    pub name: String,
+    pub value: Serialized,
+}
+
 #[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
 pub struct User {
     pub user_id: UserId,
     pub email: String,
     pub display_name: Option<String>,
-    pub first_name: Option<String>,
```
nitnelave, line 386: Meat of the diff: Removing these attributes in the `User` class, the values are stored in the `attributes` vector instead.

server/src/domain/types.rs
```diff
@@ -104,9 +104,23 @@ macro_rules! uuid {
...
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Serialized(Vec<u8>);
 
+impl std::fmt::Debug for Serialized {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Serialized")
+            .field(&String::from_utf8(self.0.clone()).unwrap_or_else(|_| {
+                format!("hash: {}", {
```
pixelrazor, line 114: sick idea!

pixelrazor (approved):
Thanks for the tight scope on this - was pretty easy to digest. Awesome work, I'm pumped for this!

server/src/domain/types.rs
nitnelave: Oh, that reminds me, I wanted to fiddle with that a bit, the output is not great. I also want to have that for LdapAttributes, currently it prints the bytes for everything because of the JpegPhoto. I remember talking about it with the LDAP library author, but I don't think it got implemented.

PR #616, opened 2023-07-01
example config for squid

Hobbabobba:
Squid as a caching proxy that uses LDAP to authenticate users. The procedure configures that only authenticated users can use the proxy.


nitnelave (approved):
There's a lot there that I don't know about, but overall it looks good!

nitnelave:
Thanks!

PR #617, opened 2023-07-04
Add a graphql method to get the schema

nitnelave:
I chose to decouple the domain-level configurable attributes, and the public-viewable schema which includes the attributes that are not stored as "user attributes". All the "hardcoded" attributes will be available as fields from grahpql, and the non-hardcoded will be only in the "user_attributes" (coming) field.

In terms of permissions, only users who have the permissions to see all the users (admin, password manager or readonly) can see the schema. Regular users cannot (they wouldn't have much use for it, and that way I don't have to filter the fields they're not allowed to see).

### DIFF
--- a/server/src/domain/handler.rs
+++ b/server/src/domain/handler.rs
@@ -210,49 +210,6 @@ pub trait BackendHandler:
 }
 
 #[cfg(test)]
-mockall::mock! {
-    pub TestBackendHandler{}
-    impl Clone for TestBackendHandler {
-        fn clone(&self) -> Self;
-    }
-    #[async_trait]
-    impl GroupListerBackendHandler for TestBackendHandler {
-        async fn list_groups(&self, filters: Option<GroupRequestFilter>) -> Result<Vec<Group>>;
-    }
-    #[async_trait]
-    impl GroupBackendHandler for TestBackendHandler {
-        async fn get_group_details(&self, group_id: GroupId) -> Result<GroupDetails>;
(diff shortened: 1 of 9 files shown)


server/src/domain/handler.rs
```diff
@@ -209,49 +209,6 @@ pub trait BackendHandler:
 {
 }
 
-#[cfg(test)]
```
nitnelave, line 212: Moved that to a new test_utils.rs file to deduplicate.

server/src/domain/ldap/utils.rs
```diff
@@ -200,6 +200,12 @@ pub fn get_custom_attribute(
     attribute_name: &str,
     schema: &Schema,
 ) -> Option<Vec<Vec<u8>>> {
+    let convert_date = |date| {
```
nitnelave, line 203: It doesn't strictly belong in this PR, but it's a small self-contained change.

server/src/infra/ldap_handler.rs
```diff
@@ -807,44 +749,6 @@ mod tests {
         setup_bound_handler_with_group(mock, "lldap_admin").await
     }
 
-    fn setup_default_schema(mock: &mut MockTestBackendHandler) {
```
nitnelave, line 810: Also moved to test_utils.rs

nitnelave:
Scratch that, I should implement getting the schema restricted to what the user can see, since I'll need it to resolve the attributes anyway.

PR #675, opened 2023-09-12
server: Only create the default admin if there are no admins

### LINKED ISSUE #563: Deleted admin account gets recreated with default password
First of all, thanks for developing this great software. It's easy to use and deploy.
However, I just noticed that when you delete the default admin "admin", it gets recreated with the default password and privileges. 
That means, to e.g. protect against brute force attacks, deleting it and using another admin account (with different name) instead, this recreates the default one.
This seems dangerous, as people might not suspect that behavior and allows access to ldap via the typical admin/password combo. 

Currently I changed it's password and removed it from the administrator group. But lldap should just check if there is at least one other user member of the administrator group.

nitnelave:
Fixes #563

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -9,7 +9,7 @@ use crate::{
     domain::{
         handler::{
             CreateUserRequest, GroupBackendHandler, GroupListerBackendHandler, GroupRequestFilter,
-            UserBackendHandler,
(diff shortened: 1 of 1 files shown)


nitnelave:
I tested it by creating a new admin, and deleting the "admin" user. After restarting the server, the admin user didn't get re-created. Note that only admin users can delete users, and you can't delete your own user, so there should always be an admin.

Issue #521, opened 2023-04-03
Support password policy


S0obi:
Hello there,

It will be nice if lldap can actually support a global password policy. For now, lldap only [checks](https://github.com/lldap/lldap/blob/main/app/src/components/change_password.rs#L39) that the password is 8 characters or more. It should be customizable by configuration with classic conditions : 

- Minimum number of characters
- x lowercase characters
- x uppercase characters
- x numeric characters
- x special characters

nitnelave:
Sure, we can implement something like that (I'm open for PRs!)

Some things to keep in mind:
- when changing passwords on the web interface, the server doesn't actually
see the password, so it's completely possible for a client to register a
weak password. The restrictions can be applied to the web UI, but that
won't stop a motivated user from registering a weak password.
- when changing passwords through LDAP, what should we do if the password
doesn't match the criteria? Refuse it? If that's the case, then we need to
be able to surface the restrictions to the end user, otherwise that's a
very frustrating experience. We should probably accept whatever password is
sent (maybe over 8 characters anyway) and count on the frontend of the
service to enforce the policy.

pixelrazor:
Some random thoughts on this:

We can , probably even today afaik, have one of  those password strength bars for indicators. We coul;d probably pass  it to the front end for live feedback, but with or without that piece, we could just implement it in the graphql api?

leanderjanssen:
What about [`zxcvbn`](https://github.com/dropbox/zxcvbn)?  
It's a password strength estimator developed by dropbox.
There is a rust library available at: https://github.com/shssoichiro/zxcvbn-rs

nitnelave:
@leanderjanssen nice! I think that's a good candidate for the "measuring entropy"-equivalent category. I think also checking if the password was leaked would help on top.

binaryben:
I'd actually make an argument against this suggestion. LDAP is LDAP. Let this tool just focus on providing LDAP and mark this as out of scope. LDAP should be used with other tools like Authelia (imho) which can do the security checks and password strength enforcement. Personally I'm going as far as disabling the password reset functionality from LLDAP in my config and leaving it to other tools.

S0obi:
My main point about creating this issue is that LLDAP will actually become a minimum viable option for users that have simple needs and are looking for an easy drop-in replacement for OpenLDAP. I understand the point behind using Authelia or Keycloak for such "advanced" mechanisms, however I will consider "strong password" requirement as something that should belong the the core project.

[...]

PR #692, opened 2023-09-28
Remove password length limitation in lldap_set_password tool

stuart938503:
The password length limitation is arbitrary and non-configurable, and there are legitimate requirements to set a password shorter than 8 characters. This change will make it simple to achieve without the need for a third party tool.

I would update the web UI code also, but am unfamiliar with Rust and the wider LLDAP project to confidently make the change and not break something. The change for this tool appears simple, however.

### DIFF
--- a/set-password/src/main.rs
+++ b/set-password/src/main.rs
@@ -97,10 +97,6 @@ pub fn register_finish(
 fn main() -> Result<()> {
     let opts = CliOpts::parse();
     ensure!(
-        opts.password.len() >= 8,
-        "New password is too short, expected at least 8 characters"
-    );
-    ensure!(
         opts.base_url.scheme() == "http" || opts.base_url.scheme() == "https",
         "Base URL should start with `http://` or `https://`"
     );
--- a/set-password/src/main.rs
+++ b/set-password/src/main.rs
@@ -30,6 +30,10 @@ pub struct CliOpts {
     /// New password for the user.
     #[clap(short, long)]
     pub password: String,
+
+    /// Should minimum password length be bypassed?
+    #[clap(long)]
+    pub bypass_password_policy: bool,
 }
 
 fn append_to_url(base_url: &Url, path: &str) -> Url {
@@ -96,6 +100,12 @@ pub fn register_finish(
 
 fn main() -> Result<()> {
     let opts = CliOpts::parse();
+    if opts.bypass_password_policy != true {
+        ensure!(
+            opts.password.len() >= 8,
+            "New password is too short, expected at least 8 characters"
+        );
+    }
     ensure!(
         opts.base_url.scheme() == "http" || opts.base_url.scheme() == "https",
         "Base URL should start with `http://` or `https://`"
--- a/set-password/src/main.rs
+++ b/set-password/src/main.rs
@@ -31,7 +31,7 @@ pub struct CliOpts {
     #[clap(short, long)]
     pub password: String,
 
-    /// Should minimum password length be bypassed?
+    /// Bypass password requirements such as minimum length. Unsafe.
     #[clap(long)]
     pub bypass_password_policy: bool,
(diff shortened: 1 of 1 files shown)


nitnelave:
Hi! Sorry, I'm not sure I understand. Why is the password length a problem? What is a real situation where you cannot have a password length of 8? It's really a minimum threshold, anything below is woefully insecure.

Granted, we should probably be measuring entropy and/or checking against HIBP for password leaks, but the length limit is an absolute minimum.

stuart938503:
First - thank you for your work on this project. A simple LDAP server is exactly what I want.

The issue is I want to set passwords shorter than 8 characters but this is not possible within LLDAP. Security is a case for password length limitations, but is a minor concern for me. Setting a password that is 6 characters long has proven incredibly challenging today, and while I _could_ use an 8 character password I would rather avoid it.

Having an 8 character length limitation is fairly arbitrary, and does not appear to be configurable. I have spent an hour or two trying to work around this limitation today - the web UI (which is compiled to WASM and makes on the fly changes hard/impossible), this tool, LDAP Admin (which was returned an _unsupported operation_ error from LLDAP), looking at other LDAP client tools that had various obstacles e.g. java/other libraries/etc.

LLDAP itself (as I understand it reading the home page) is meant to be a simple identity store, with other features handled by other tools e.g. Authelia. Whether it should also provide an enforceable password policy appears to be a fairly debatable topic (see #521), and so on the surface I (respectfully) don't think having _just_ a non-configurable password length limitation is in keeping with the spirit (as I understand it) of LLDAP, and is a frustrating limitation.

On balance, I feel the change proposed by this PR is very minor but removes a (potentially) annoying limitation from the tool. The 8 character limitation is still present in the web UI, and this PR serves as a good workaround until/if a configurable password policy is implemented (if that is the direction LLDAP wants to go).

I hope I my comments are articulated well enough that they interpreted in the respectful and constructive manner they were intended :).

nitnelave:
I'm still not sure why you can't use an 8 character password, but I guess that's up to you :)

While LLDAP tries to be a simple identity store, it also tries to have safe defaults, and to the point that it's simple, those defaults are not always configurable.

In this case (since the server doesn't receive the password anyway), any added complexity would go to the password setting tool, not the server, so it's probably fine.

However, I'd like to keep the default of a minimum of 8 characters. To that end, I propose a command line flag like "--bypass-password-policy".

WDYT?

stuart938503:
I think that would be a completely agreeable alternative :)

stuart938503:
I have added that flag. 

Caveat: while I have tested this on my own machine, I have never touched Rust before.

set-password/src/main.rs
```diff
@@ -96,10 +100,12 @@ pub fn register_finish(
 
 fn main() -> Result<()> {
     let opts = CliOpts::parse();
-    ensure!(
-        opts.password.len() >= 8,
-        "New password is too short, expected at least 8 characters"
-    );
+    if opts.bypass_password_policy != true {
```
nitnelave, line 103:
```rust
ensure(opts.bypass_passwors_policy || opts.password.len() >= 8, ...
```

nitnelave (changes requested):

set-password/src/main.rs
```diff
@@ -30,6 +30,10 @@ pub struct CliOpts {
     /// New password for the user.
     #[clap(short, long)]
     pub password: String,
+
+    /// Should minimum password length be bypassed?
```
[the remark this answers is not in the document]
stuart938503: Agree - updated

set-password/src/main.rs
stuart938503: Updated

nitnelave (approved):

nitnelave:
Thanks! Merged :)

PR #699, opened 2023-10-05
Basic support for Paged Results Control (RFC 2696)

broeng:
This implements rudimentary support for the Paged
Results Control.

No actual pagination is performed, and we ignore
any requests for specific window sizes for paginated results.

Instead, the full list of search results is returned for any searches, and a control is added to the
SearchResultsDone message, informing the client that there is no further results available.

### DIFF
--- a/server/src/infra/ldap_server.rs
+++ b/server/src/infra/ldap_server.rs
@@ -13,7 +13,7 @@ use actix_rt::net::TcpStream;
 use actix_server::ServerBuilder;
 use actix_service::{fn_service, ServiceFactoryExt};
 use anyhow::{anyhow, Context, Result};
-use ldap3_proto::{proto::LdapMsg, LdapCodec};
+use ldap3_proto::{proto::LdapControl, proto::LdapMsg, proto::LdapOp, LdapCodec};
 use rustls::PrivateKey;
 use tokio_rustls::TlsAcceptor as RustlsTlsAcceptor;
 use tokio_util::codec::{FramedRead, FramedWrite};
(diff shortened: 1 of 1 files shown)


broeng:
Haven't tested yet. Only pushed to gather feedback.

server/src/infra/ldap_handler.rs
```diff
@@ -188,7 +188,8 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
             },
             LdapPartialAttribute {
                 atype: "supportedControl".to_string(),
-                vals: vec![],
+                // Paged Results Control (RFC 2696)
+                vals: vec![b"1.2.840.113556.1.4.319".to_vec()],
```
nitnelave, line 192: Hmmm... I wouldn't go as far as listing it in the supported controls. Our implementation is... rudimentary :D

server/src/infra/ldap_server.rs
```diff
@@ -39,12 +39,22 @@ where
             if result.is_empty() {
                 debug!("No response");
             }
+            let results: i64 = result.len().try_into().unwrap();
             for response in result.into_iter() {
                 debug!(?response);
+                let controls =
```
nitnelave, line 45: If I wanted to be pedantic, I would ask you to assert that it's the last element in the response.

Thankfully, I'll have to maintain that code, and the simpler the better ;) I like this solution.

nitnelave (approved):
LGTM!

Can you test it with your setup to make sure that it's accepted?

nitnelave:
You'll need to run `cargo fmt`: https://github.com/lldap/lldap/actions/runs/6420405003/job/17432425224?pr=699

server/src/infra/ldap_handler.rs
broeng: Good point, was thinking it'd be necessary, but can't really imagine it is. :)

broeng:
I have pushed a formatted update, that also removes it from the listed features.

I'll get it tested as soon as possible (hopefully tomorrow) -- let's wait with merge until I can verify it solves the problem. :-)

broeng:
> Can you test it with your setup to make sure that it's accepted?

Tested, and solves the/my problem. :+1: :-)

nitnelave (approved):

nitnelave:
Thanks! Merged. It'll be in the `latest` image in a few minutes

PR #703, opened 2023-10-08
Add notes on fixing postgresql primary keys.

MI3Guy:


nitnelave (commented):
Can we package that in a single SQL statement? Then we can add it to the migration script and we don't even need the special instructions

MI3Guy:
Updated so it is in the sed command. I had thought there would be more tables, but as far as I can tell, only the groups table uses an auto increment id so it should be the only one with the issue.

nitnelave (approved):

nitnelave:
Thanks! Merged

PR #709, opened 2023-10-22
chore: update repository references

fgardt:
There were still some references to the old repository at `nitnelave/lldap` _(most notably in the footer of the web interface)_.


nitnelave (approved):
Thanks!

fgardt:
that clippy fail seems unrelated but I can fix that quick if you want to?

nitnelave:
If you want, but in another PR. It triggered due to a clippy update. I'll merge as is

PR #711, opened 2023-10-23
add date-based tagging to docker matrix jobs

### LINKED ISSUE #613: Improved Image Versioning and Release Cycle
https://hub.docker.com/r/nitnelave/lldap/tags

Currently only the latest images seem to be frequently updated, but it would be nice to have daily, or per PR images.
I use renovate bot to automatically update images, but when using latest that doesn't work.
But if I were to use the named images I would be months behind in the development cycle, which would be annoying.

Grafana images have very nice versioning as an example:
https://hub.docker.com/r/grafana/grafana/tags

Also, hello from hackernews :D

aroberts:
fixes #613 

adds additional date tag when publishing "latest" docker images


nitnelave (approved):

nitnelave:
Thanks!

PR #718, opened 2023-10-29
Add bootstrap script

Sad-Soul-Eater:
Since I run lldap in my home-lab Kubernetes cluster, I made this script to manage my lldap instance in a git-ops manner. 

The overall process is next:
- The script running in the Kubernetes job managed by Argo CD
- Whenever new changes are committed to config files, Argo CD relaunches the job.
- The script then creates/updates or deletes users, and groups, and adds/removes users from corresponding groups.

The script can be used for maintaining lldap in the Kubernetes cluster, in the way I mentioned before. Or it also can be used for initial bootstrapping of the lldap instance in other ways, by manually running from the terminal for example.


martadinata666:
IMO this should create a new folder.
`example_configs/argocd/bootstrap.md and the other files` to prevent confusion as many `example configs` are directly related to the application/service name. Maybe some other way to prevent confusion also will do.

Sad-Soul-Eater:
> IMO this should create a new folder. `example_configs/argocd/bootstrap.md and the other files` to prevent confusion as many `example configs` are directly related to the application/service name. Maybe some other way to prevent confusion also will do.

I agree that the `example_configs` dir is a _little_ messy, but the script isn't directly related to Argo CD nor even Kubernetes, it can be run manually after starting the lldap container, by systemd unit, Terraform, Ansible, etc. The script doesn't have Kubernetes or Argo CD specific code, I use them just to automatically re-run the script on config files changing.

So that directory shouldn't called `example_configs/argocd`.

Also, I added two usage examples in `bootstrap.md`, how to run the script manually and in the docker container with docker-compose.

example_configs/bootstrap.md
```diff
@@ -0,0 +1,253 @@
...
+  ]
+}
+
+```
+
+## Usage example
+
+### Manually
+
+The script can be run manually in the terminal for initial bootstrapping your lldap instance.
```
nitnelave, line 102: Nit: bootstrapping _of_ your lldap instance.

example_configs/bootstrap.md
```diff
@@ -0,0 +1,253 @@
...
+You only should ensure that [required packages](#required-packages) are installed
```
nitnelave, line 103: You should make sure that _the_ required packages are installed and _the_ environment variables _are_ configured properly.

example_configs/bootstrap.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+You should mount `bootstrap` dir to lldap container and set the corresponding `env` variables:
+
+```yaml
+version: "3"
+
+services:
+  lldap:
+    image: lldap/lldap:v0.5.0
```
nitnelave, line 144: Can you explicitly refer to the alpine image, since it's a requirement (for apk)? Also it might be nice to mention it above in the docs.

example_configs/bootstrap.sh
```diff
@@ -0,0 +1,414 @@
...
+  else
+    printf 'Group "%s" (%s) successfully created\n' "$group_name" "$(printf '%s' "$response" | jq --raw-output '.data.createGroup.id')"
+  fi
+}
+
+delete_group() {
+  local group_name="$1" id=''
+
+  if [[ $(is_group_exists "$group_name") != true ]]; then
+    printf 'Group "%s" is not exists\n' "$group_name"
```
nitnelave, line 110: nit: "_does_ not exist"

example_configs/bootstrap.sh
```diff
@@ -0,0 +1,414 @@
...
+get_group_list() {
+  local query='{"query":"query GetGroupList {groups {id displayName}}","operationName":"GetGroupList"}'
+  make_query <(printf '%s' "$query") <(printf '{}')
+}
+
+get_group_array() {
+  get_group_list | jq --raw-output '.data.groups[].displayName'
+}
+
+is_group_exists() {
```
nitnelave, line 77: I'd name it just `group_exists`

example_configs/bootstrap.sh
```diff
@@ -0,0 +1,414 @@
...
+    printf '%s\n' "$error"
+  else
+    printf 'Group "%s" (%s) successfully created\n' "$group_name" "$(printf '%s' "$response" | jq --raw-output '.data.createGroup.id')"
+  fi
+}
+
+delete_group() {
+  local group_name="$1" id=''
+
+  if [[ $(is_group_exists "$group_name") != true ]]; then
```
nitnelave, line 109: `if [[ ! $(group_exists "$group_name") ]]; then`

nitnelave (changes requested):
This shell script is waaay past the point where I would have written it in a saner language. If not Rust, then at least Python. Just a fair warning that I probably won't touch it, even if it breaks with newer updates (now, if it was written in python/rust, I would probably update it).

But as an example bootstrap script, sure.

Let's put it in a `bootstrap` folder, in example_configs, though.

example_configs/bootstrap.md
Sad-Soul-Eater: Done :ok_hand:

example_configs/bootstrap.md
Sad-Soul-Eater: Done :ok_hand:

example_configs/bootstrap.md
Sad-Soul-Eater: Added support for installing packages by apt and mention about packages in the docs

example_configs/bootstrap.sh
Sad-Soul-Eater: Done :ok_hand:

example_configs/bootstrap.sh
Sad-Soul-Eater: Done :ok_hand:

example_configs/bootstrap.sh
Sad-Soul-Eater: There are no booleans in bash, so this `if` condition would be always false since the returned string is always not empty. All this `true` and `false` in the script is just a string.

nitnelave (approved):
Maybe you want to add something to the readme? In the section https://github.com/lldap/lldap#with-kubernetes for instance.

README.md
```diff
@@ -162,6 +162,10 @@ front-end.
 
 See https://github.com/Evantage-WS/lldap-kubernetes for a LLDAP deployment for Kubernetes
 
+You can bootstrap your lldap instance (users, groups)
+using [bootstrap.sh](example_configs/bootstrap/bootstrap.md#kubernetes-job).
+It Can be run by Argo CD for managing users in git-opt way, or like one shot job.
```
nitnelave, line 167: nit: "It can" (capital)
"or as a one-shot job"

example_configs/bootstrap/bootstrap.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Group config file example
+
+Group configs are used to define groups that will be created by the script
+
+Fields description:
+
+* `name`: name of the group (**MANDATORY**)
+
+```json
```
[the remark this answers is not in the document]
nitnelave: Ping.

nitnelave (changes requested):

example_configs/bootstrap/bootstrap.md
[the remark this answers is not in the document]
Sad-Soul-Eater: The example is valid, it's not in JSON standard but because to iterate over all JSONs `jq--compact-output` (lines 387 and 445)  will merge all files in one big file like this:

```json
{"name":"group-1"}
{"name":"group-2"}
{"name":"group-3"}
```

Then in the loop (lines 382 and 411), the script reads each line separately, so each line is only one JSON structure. It doesn't matter if there are many nested JSONs in one file or each in a separate one all this will work in the same manner.

README.md
Sad-Soul-Eater: Done :ok_hand:

PR #752, opened 2023-12-05
app: make it possible to serve lldap behind a sub-path

MinerSebas:
This adds support for placing lldap on a sub-path.

To use it, a User needs to set the `http_url` setting to a URL with a sub-path (`http://example.com/lldap`) and setup the reverse-proxy to strip the sub-path.
Caddy example:
```
handle_path /lldap* {
  reverse_proxy http://127.0.0.1:17170
}
```

Future Work:
Currently, the sub-path is injected every time `index.html` and `main.js` are called, but this could be cached instead.
I tried it, but I never used actix before, and thus couldn't make it work.

### DIFF
--- a/app/src/infra/api.rs
+++ b/app/src/infra/api.rs
@@ -18,6 +18,10 @@ fn get_claims_from_jwt(jwt: &str) -> Result<JWTClaims> {
 
 const NO_BODY: Option<()> = None;
 
+fn base_url() -> String {
+    yew_router::utils::base_url().unwrap_or_default()
+}
+
 async fn call_server(
     url: &str,
     body: Option<impl Serialize>,
@@ -97,7 +101,7 @@ impl HostService {
         };
         let request_body = QueryType::build_query(variables);
         call_server_json_with_error_message::<graphql_client::Response<_>, _>(
-            "/api/graphql",
+            &(base_url() + "/api/graphql"),
             Some(request_body),
             error_message,
         )
@@ -109,7 +113,7 @@ impl HostService {
         request: login::ClientLoginStartRequest,
     ) -> Result<Box<login::ServerLoginStartResponse>> {
         call_server_json_with_error_message(
-            "/auth/opaque/login/start",
+            &(base_url() + "/auth/opaque/login/start"),
             Some(request),
             "Could not start authentication: ",
         )
@@ -118,7 +122,7 @@ impl HostService {
 
     pub async fn login_finish(request: login::ClientLoginFinishRequest) -> Result<(String, bool)> {
         call_server_json_with_error_message::<login::ServerLoginResponse, _>(
-            "/auth/opaque/login/finish",
+            &(base_url() + "/auth/opaque/login/finish"),
             Some(request),
             "Could not finish authentication",
         )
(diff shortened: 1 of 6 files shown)


server/src/infra/tcp_server.rs
```diff
@@ -12,22 +12,31 @@ use crate::{
...
 
-async fn index() -> actix_web::Result<NamedFile> {
-    let path = PathBuf::from(r"app/index.html");
-    Ok(NamedFile::open(path)?)
+async fn index<Backend>(data: web::Data<AppState<Backend>>) -> actix_web::Result<impl Responder> {
+    let mut file = std::fs::read_to_string(r"./app/index.html")?;
+
+    if data.server_url.path() != "/" {
+        file = file
+            .split_terminator("<base href=\"/\">")
```
nitnelave, line 33: I think you can use https://doc.rust-lang.org/std/string/struct.String.html#method.replace instead.

server/src/infra/tcp_server.rs
```diff
@@ -68,6 +77,22 @@ pub(crate) fn error_to_http_response(error: TcpError) -> HttpResponse {
...
 }
 
+async fn main_js_handler<Backend>(
+    data: web::Data<AppState<Backend>>,
+) -> actix_web::Result<impl Responder> {
+    let mut file = std::fs::read_to_string(r"./app/static/main.js")?;
+
+    if data.server_url.path() != "/" {
+        file = file
+            .split_terminator("/pkg/")
```
nitnelave, line 87: Same here

app/src/infra/api.rs
```diff
@@ -18,6 +18,10 @@ fn get_claims_from_jwt(jwt: &str) -> Result<JWTClaims> {
 
 const NO_BODY: Option<()> = None;
 
+fn base_url() -> String {
+    yew_router::utils::base_url().unwrap_or_default()
```
nitnelave, line 22: This reads from the `<base>` tag?

nitnelave (changes requested):
Looks nice!

Can you make sure the email reset works with the subpath?
Also, make sure to try in a private window to clear the caches, in case.

app/src/infra/api.rs
MinerSebas: Yes, that's what it does.

MinerSebas:
> Can you make sure the email reset works with the subpath?

Confirmed on Discord that email resetting works correctly.

> Also, make sure to try in a private window to clear the caches, in case.

Everything works correctly.

server/src/infra/auth_service.rs
```diff
@@ -352,19 +352,24 @@ where
     let token = create_jwt(data.get_tcp_handler(), &data.jwt_key, name, groups).await;
     let refresh_token_plus_name = refresh_token + "+" + name.as_str();
 
+    let mut path = data.server_url.path().to_string();
+    if path != "/" {
```
nitnelave, line 356: If the URL ends with a slash, is it trimmed when calling path()? Or could you end up with 2 slashes here?

It seems more straightforward to say `if !path.ends_with("/")`

nitnelave (approved):

nitnelave:
Thank you for the contribution!

PR #757, opened 2023-12-11
example for kasm config

jakob42:
I created an example of how to use kasm with lldap


nitnelave (changes requested):
Looks great! Can you add it to the readme?

jakob42:
Sure!

nitnelave (approved):

nitnelave:
Thanks a lot :)

jakob42:
It's just a drop compared to your work. Thanks!

PR #770, opened 2023-12-19
Add example conf for zitadel

Halyul:
Finally found a sso solution that baby can use 😆

### DIFF
--- a/example_configs/zitadel.md
+++ b/example_configs/zitadel.md
@@ -1,0 +1,38 @@
+# Configuration for Zitadel
+In Zitadel, go to `Instance > Settings` for instance-wide LDAP setup or `<Organization Name> > Settings` for organization-wide LDAP setup.
+
+## Identity Providers Setup
+Click `Identity Providers` and select `Active Directory/LDAP`.
+
+Replace every instance of `dc=example,dc=com` with your configured domain.
+**Group filter is not supported at the time of writing.**
+### Connection
+* Name: The name to identify your identity provider
+* Servers: `ldaps://<FQDN or Host IP>:<Port for LADPS>` or `ldap://<FQDN or Host IP>:<Port for LADP>` 
+* BaseDn: `dc=example,dc=com`
+* BindDn: `cn=admin,ou=people,dc=example,dc=com`. It is recommended that you create a separate user account (e.g, `bind_user`) instead of `admin` for sharing Bind credentials with other services. The `bind_user` should be a member of the `lldap_strict_readonly` group to limit access to your LDAP configuration in LLDAP.
+* Bind Password: `<user password>`
+
+### User binding
+* Userbase: `ou=people,dc=example,dc=com`
+* User filters: `uid`. `mail` will not work.
+* User Object Classes: `inetOrgPerson`
+
+### LDAP Attributes
+* ID attribute: `uid`
+* Avatar Url attribute: `jpegPhoto`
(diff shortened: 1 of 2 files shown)


example_configs/zitadel.md
```diff
@@ -0,0 +1,38 @@
...
+
+### User binding
+* Userbase: `ou=people,dc=example,dc=com`
+* User filters: `uid`. `mail` will not work.
+* User Object Classes: `inetOrgPerson`
+
+### LDAP Attributes
+* ID attribute: `uid`
+* Avatar Url attribute: `jpegPhoto`
+* Displayname attribute: `cn`
```
nitnelave, line 24: `displayName` is more self-explanatory

example_configs/zitadel.md
```diff
@@ -0,0 +1,38 @@
...
+* Bind Password: `<user password>`
+
+### User binding
+* Userbase: `ou=people,dc=example,dc=com`
+* User filters: `uid`. `mail` will not work.
+* User Object Classes: `inetOrgPerson`
+
+### LDAP Attributes
+* ID attribute: `uid`
+* Avatar Url attribute: `jpegPhoto`
```
nitnelave, line 23: Does that work? It's not a URL, it's a raw jpeg given as bytes.

example_configs/zitadel.md
```diff
@@ -0,0 +1,38 @@
...
+* Displayname attribute: `cn`
+* Email attribute: `mail`
+* Given name attribute: `givenName`
+* Family name attribute: `sn`
```
nitnelave, line 27: You can use `lastName` here

example_configs/zitadel.md
```diff
@@ -0,0 +1,38 @@
...
+
+### optional
+* Account creation allowed [x]
```
nitnelave, line 30: Does that mean creating an account in Zitadel creates a user in LLDAP? Have you tested that?

We have limited support for new user creation through LDAP, so it may work. However, you'll have to mention it above, you'll need `lldap_admin` permissions rather than `strict_readonly`.

example_configs/zitadel.md
```diff
@@ -0,0 +1,38 @@
+# Configuration for Zitadel
+In Zitadel, go to `Instance > Settings` for instance-wide LDAP setup or `<Organization Name> > Settings` for organization-wide LDAP setup.
+
+## Identity Providers Setup
+Click `Identity Providers` and select `Active Directory/LDAP`.
+
+Replace every instance of `dc=example,dc=com` with your configured domain.
+**Group filter is not supported at the time of writing.**
```
nitnelave, line 8: Is that an LLDAP limitation, or a Zitadel one?

nitnelave (changes requested):
Looks good! A couple of comments/questions but that's fine

example_configs/zitadel.md
Halyul: Just tested. It doesn't work 😢

example_configs/zitadel.md
Halyul: `Account creation allowed` means an account will be created in zitadel, not lldap

example_configs/zitadel.md
Halyul: it's zitadel

nitnelave (approved):
Thanks!

PR #771, opened 2023-12-19
Example homeassistant update

Halyul:
Updated the homeassistant script so it can support assigning users to internal regular and admin groups

### DIFF
--- a/example_configs/home-assistant.md
+++ b/example_configs/home-assistant.md
@@ -17,18 +17,18 @@ homeassistant:
     - type: command_line
       command: /config/lldap-ha-auth.sh
       # arguments: [<LDAP Host>, <regular user group>, <admin user group>, <local user group>]
-      # <regular user group>: Only allow users in the <regular user group> (e.g., 'homeassistant_user')
-      # group to login. Users will have the default 'system-users' permission.
+      # <regular user group>: Find users that has permission to access homeassistant, anyone inside
+      # this group will have the default 'system-users' permission in homeassistant.
       #
-      # <admin user group>: Allow users in the <regular user group> (e.g., 'homeassistant_user') with
-      # <admin user group> (e.g., 'homeassistant_admin') to login and have the default 'system-admin' 
-      # permission.
+      # <admin user group>: Allow users in the <regular user group> to be assigned into 'system-admin' group.
+      # Anyone inside this group will not have the 'system-users' permission as only one permission group
+      # is allowed in homeassistant
       #
       # <local user group>: Users in the <local user group> (e.g., 'homeassistant_local') can only access
       # homeassistant inside LAN network.
       #
-      # Change to ["https://lldap.example.com"] to allow all users. All of them will have the default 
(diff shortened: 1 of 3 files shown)


example_configs/home-assistant.md
```diff
@@ -16,9 +16,20 @@ homeassistant:
...
     - type: command_line
       command: /config/lldap-ha-auth.sh
-      # Only allow users in the 'homeassistant_user' group to login.
-      # Change to ["https://lldap.example.com"] to allow all users
-      args: ["https://lldap.example.com", "homeassistant_user"]
+      # arguments: [<LDAP Host>, <regular user group>, <admin user group>, <local user group>]
+      # <regular user group>: Only allow users in the <regular user group> (e.g., 'homeassistant_user')
+      # group to login. Users will have the default 'system-users' permission.
+      #
+      # <admin user group>: Allow users in the <regular user group> (e.g., 'homeassistant_user') with
```
nitnelave, line 23: allow users belonging to both the `<regular user group>` and `<admin user group>`

WDYT?

example_configs/home-assistant.md
```diff
@@ -16,9 +16,20 @@ homeassistant:
...
+      # <admin user group> (e.g., 'homeassistant_admin') to login and have the default 'system-admin' 
+      # permission.
+      #
+      # <local user group>: Users in the <local user group> (e.g., 'homeassistant_local') can only access
+      # homeassistant inside LAN network.
+      #
+      # Change to ["https://lldap.example.com"] to allow all users. All of them will have the default
```
nitnelave, line 30: How about:
> Only the first argument is required. `["https://lldap.example.com"]` allows all users to log in from anywhere and have 'system-users' permissions.

example_configs/lldap-ha-auth.sh
```diff
@@ -66,5 +66,26 @@ fi
 
 DISPLAY_NAME=$(jq -r .displayName <<< $USER_JSON)
 
+IS_ADMIN=false
+if [[ ! -z "$3" ]] && jq -e '.groups|map(.displayName)|index("'"$3"'")' <<< "$USER_JSON" > /dev/null 2>&1; then
```
nitnelave, line 70: I'm not going to pretend I understand this. I assume you tested it?

nitnelave (changes requested):
Some formatting comments.

example_configs/home-assistant.md
Halyul: My expression was inaccurate. `<regular user group>` is a list of users that are permitted to access `homeassistant`. Upon that, `<admin user group>` enables the users within `<regular user group>` to be in the admin group.

example_configs/lldap-ha-auth.sh
Halyul: i tested the case where a regular user with the local group and an admin user without the local group logged into homeassistant. the behaviour is correct.

nitnelave (approved):
Thanks!

PR #773, opened 2023-12-21
server: Add an option to force reset the admin password

### LINKED ISSUE #748: [FEATURE REQUEST] Force reset admin user/password
**Motivation**
Allow server admins to reset the admin user and password when something goes wrong

When a config value is set (CLI and env as well), force the recreation/password reset of the admin user to the values in the config

Delete all the admin users. Not ideal.

**Additional context**
Usually a password reset is enough, but sometimes it's not setup/doesn't work.

nitnelave:
Fixes #748

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -78,6 +78,12 @@
 ## is just the default one.
 #ldap_user_pass = "REPLACE_WITH_PASSWORD"
 
+## Force reset of the admin password.
+## Break glass in case of emergency: if you lost the admin password, you
+## can set this to true to force a reset of the admin password to the value
+## of ldap_user_pass above.
+# force_reset_admin_password = false
+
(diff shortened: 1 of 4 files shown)


nitnelave:
@martadinata666 WDYT?

martadinata666:
I saw the issue thread,  the solution is kinda prone to accidental moves. In my personal opinion admin password reset should be triggered manually, `the admin really want to reset`. Something like `lldap --admin-password-reset` or add `FORCE_RESET` environment. The simplest thing is, `FORCE_RESET` just go back to `password` rather than taking value from CLI and ENV.

This will line up with the #765 to keep simplest.

martadinata666:
Oh I see, yes. That is also valid. Accidental like ENV missing, edit file config when env exist, so on, there are always somehow magically happen. 🤣

PR #774, opened 2023-12-25
docs(zitadel): make the doc more comprehensive

Halyul:
- fixed `Userbase` attribute
- added `Preferred username attribute`
- added `Automatic creation`

### DIFF
--- a/example_configs/zitadel.md
+++ b/example_configs/zitadel.md
@@ -4,8 +4,9 @@ In Zitadel, go to `Instance > Settings` for instance-wide LDAP setup or `<Organi
 ## Identity Providers Setup
 Click `Identity Providers` and select `Active Directory/LDAP`.
 
-Replace every instance of `dc=example,dc=com` with your configured domain.
 **Group filter is not supported in `Zitadel` at the time of writing.**
+
+Replace every instance of `dc=example,dc=com` with your configured domain.
 ### Connection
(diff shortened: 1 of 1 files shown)


example_configs/zitadel.md
```diff
@@ -14,16 +15,17 @@ Replace every instance of `dc=example,dc=com` with your configured domain.
 * Bind Password: `<user password>`
 
 ### User binding
-* Userbase: `ou=people,dc=example,dc=com`
+* Userbase: `dn`
```
nitnelave, line 18: I'm not sure I understand this one. What's the doc for this?

Halyul: Their [doc](https://zitadel.com/docs/guides/integrate/identity-providers/ldap#create-a-new-ldap-provider) isn't really helpful, or maybe just I misinterpreted their doc.

I took a look on lldap's logs last night, and notice that when `ou=people,dc=example,dc=com` is used, lldap reported an error related to `unknown attribute` and asked me to use `ignore user attribute`. So after some researching, I found `dn` is the correct one used here.

nitnelave: Ah yeah, they do mention "dn"

Their vocabulary is all over the place, it's all kinds of wrong...

Thanks for doing the legwork!

nitnelave (approved):

Issue #755, opened 2023-12-10
[FEATURE REQUEST] Build `rootless` container image


onedr0p:
**Motivation**

LLDAP cannot not run as nonroot with strict security constraints because it uses gosu and chowns data on start. As a tool that is built for security, best practices should be followed and gosu should be removed in favor of the supported method provided by the Docker runtime.

https://github.com/lldap/lldap/blob/70d85524db4d7070f1fda499702bbbd3ec2dd8b5/docker-entrypoint.sh#L24-L26

The error I get is when I set `readOnlyRootFilesystem` to `true` in Kubernetes `securityContext`: 

```
k logs -n security lldap-6bf866cc88-lnqtv
Defaulted container "main" out of: main, init-db (init)
> Setup permissions..
chown: /app: Read-only file system
```

**Describe the solution you'd like**

Remove gosu and support the official method of using the `user` option in Docker Compose:

> user
> Optional User and Group ID you want to run the container as. lldap will run using this UID:GID and any files it creates in your /config volume will also be owned by this user and group. The default for this, if not specified, is 1000:1000.

```yaml
networks:
  lldap:
    name: lldap
    external: true

services:
  recyclarr:
    image: ghcr.io/lldap/lldap
    container_name: lldap
    user: 1000:1000
    networks: [lldap]
    volumes:
      - ./lldap_data:/data
    environment:
      - TZ=America/Santiago
```

onedr0p:
I am willing to PR this change with the supporting docs around it.

martadinata666:
Initially LLDAP also adopt that conformity, v0.3 *cmiiw. The `gosu` was used to solve installation on synology, or NAS software alike, essentially when deployed LLDAP got permission error to read/write `/data` and `/app` even when specify `user: xxxx:xxxx`. Tbf I don't use syno, so I don't know how their mounting permission works. If this can be cleared up maybe we can adopt docker method.

onedr0p:
This would be a breaking change for some users of the project, because they will need to adopt the official way to handle it. It's been years since I ran Syno but IIRC it is just docker-compose. So as long as the user sets that `user` var and `chown`s the data folder themselves prior to deploying lldap or uses the `uid:gid` that already is set on the lldap folder they should not have an issue.

nitnelave:
The current image uses gosu in order to solve common permission problems
that made installation harder. I'd like to keep it that way, in part
because of compatibility reasons.

However, I'm entirely open to also having a non root image that you can opt
[...]

PR #775, opened 2023-12-26
Rootless container

martadinata666:
Testing build for rootless container
1. Add rootless entrypoint 
2. Add Dockerfile for rootless
3. Add rootless meta
4. Build rootless images

Run as user 1000 (a.k.a lldap) by default. Deployer will be responsible for permission setup, only check `is config file exist?`.

#755 

Things to do
- [x] Temp base image. Local service registry route.
- [x] Rename docker-entry-point for rootless.
- [x] Add comments push behaviour on dummy registry
- [x] Gosu fetch from github and only on root image

Some resources:
* https://docs.docker.com/build/ci/github-actions/share-image-jobs/
* https://docs.docker.com/build/ci/github-actions/local-registry/


docker-entrypoint.sh-rootless
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CONFIG_FILE=/data/lldap_config.toml
+
+if [[ ( ! -w "/data" ) ]] || [[ ( ! -d "/data" ) ]]; then
```
nitnelave, line 6: For compatibility, we can use `[` instead of `[[`, and I don't think we need the inner parentheses.

nitnelave: Also, we don't really need `/data` to be writable: it's only used to set up the initial config.

How about:
- If the config file is missing:
  - if the folder is writable, copy the template
  - otherwise, error.

docker-entrypoint.sh-rootless
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if [[ ! -f "$CONFIG_FILE" ]]; then
+  echo "[entrypoint] Copying the default config to $CONFIG_FILE"
+  echo "[entrypoint] Edit this file to configure LLDAP."
+  cp /app/lldap_config.docker_template.toml $CONFIG_FILE
+fi
+
+if [[ ! -r "$CONFIG_FILE" ]]; then
+  echo "[entrypoint] Config file is not readable. Check the permissions"
```
nitnelave, line 20: We need the file name/path in the error.

.github/workflows/Dockerfile.ci.alpine-rootless
```diff
@@ -0,0 +1,115 @@
+FROM debian:bullseye AS lldap
```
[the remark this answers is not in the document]
martadinata666: Yes, we can. Some other ways, take the standard image and then override `USER` and `ENTRYPOINT` or create a base image then create the root and rootless image based on that. 🤔

docker-entrypoint.sh-rootless
martadinata666: How just just check if config file exist? In the end if it fail likely and mostly is permission issue.
```
lldap-lldap-1  | [entrypoint] Copying the default config to /data/lldap_config.toml
lldap-lldap-1  | [entrypoint] Edit this /data/lldap_config.toml to configure LLDAP.
lldap-lldap-1  | cp: can't create '/data/lldap_config.toml': Permission denied
lldap-lldap-1 exited with code 1
```

nitnelave: Yeah, but the error is not as nice. You can say "Could not create the default config file /data/lldap_config.toml, please copy the template from the repository to get started" or something similar.

.github/workflows/Dockerfile.ci.alpine-rootless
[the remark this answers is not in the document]
nitnelave: That sounds good too.

docker-entrypoint.sh-rootless
martadinata666: Something like this I think sufficient?
```
#!/usr/bin/env bash
set -euo pipefail

CONFIG_FILE=/data/lldap_config.toml

if [ ! -f "$CONFIG_FILE" ]; then
  echo "[entrypoint] Copying the default config to $CONFIG_FILE"
  echo "[entrypoint] Edit this $CONFIG_FILE to configure LLDAP."
  if cp /app/lldap_config.docker_template.toml $CONFIG_FILE; then
     echo "Configuration copied successfully."
  else
     echo "Fail to copy configuration, check permission on /data or manually create one by copying from LLDAP repository"
     exit 1
  fi
fi

echo "> Starting lldap.."
echo ""
exec /app/lldap "$@"
exec "$@"
```

nitnelave: Looks good!

martadinata666: Oh, on my testing I saw something weird, maybe just a nitpick
```
Attaching to lldap-lldap-1
lldap-lldap-1  | [entrypoint] Copying the default config to /data/lldap_config.toml
lldap-lldap-1  | [entrypoint] Edit this /data/lldap_config.toml to configure LLDAP.
lldap-lldap-1  | cp: can't create '/data/lldap_config.toml': Permission denied
lldap-lldap-1  | Fail to copy configuration, check permission on /data or manually create one by copying from LLDAP repository
####I Forgot to Break Here ####
lldap-lldap-1  | > Starting lldap..
lldap-lldap-1  | 
###################
# Why LLDAP still tell load from /data/lldap_config.toml, as `/data/lldap_config.toml` non-exist.
###################
lldap-lldap-1  | Loading configuration from /data/lldap_config.toml
lldap-lldap-1  | Configuration: Configuration {
lldap-lldap-1  |     ldap_host: "0.0.0.0",
lldap-lldap-1  |     ldap_port: 3890,
lldap-lldap-1  |     http_host: "0.0.0.0",
lldap-lldap-1  |     http_port: 17170,
lldap-lldap-1  |     jwt_secret: ***SECRET***,
lldap-lldap-1  |     ldap_base_dn: "dc=example,dc=com",
lldap-lldap-1  |     ldap_user_dn: UserId(
lldap-lldap-1  |         CaseInsensitiveString(
lldap-lldap-1  |             "admin",
lldap-lldap-1  |         ),
lldap-lldap-1  |     ),


```

docker-entrypoint.sh-rootless
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
nitnelave, line 1: For the name of the file, maybe docker_entrypoint_rootless.sh ?
Let's keep the shell extension.

nitnelave (approved):

.github/workflows/Dockerfile.ci.alpine-base
```diff
@@ -0,0 +1,112 @@
...
+    && for file in $(cat /lldap/app/static/libraries.txt); do wget -P app/static "$file"; done \
+    && for file in $(cat /lldap/app/static/fonts/fonts.txt); do wget -P app/static/fonts "$file"; done \
+    && chmod a+r -R .
+
+FROM alpine:3.16
+WORKDIR /app
+ENV UID=1000
+ENV GID=1000
+ENV USER=lldap
+ENV GOSU_VERSION 1.14
```
nitnelave, line 67: I think we can move gosu to the root container only, the rootless one doesn't need it.

.github/workflows/docker-build-static.yml
```diff
@@ -488,10 +497,40 @@ jobs:
...
+          # list of Docker images to use as base name for tags
+          images: |
+            localhost:5000/lldap/lldap
+          tags: ${{ matrix.container }}-base
+
+      - name: Build ${{ matrix.container }} Base Docker Image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          #push: ${{ github.event_name != 'pull_request' }}
```
nitnelave, line 518: Uncomment

.github/workflows/docker-build-static.yml
```diff
@@ -523,6 +562,39 @@ jobs:
...
+          # vX-debian
+          # vX.Y-debian
+          # vX.Y.Z-debian
+          # latest-debian
+          # stable-debian
+          # YYYY-MM-DD-debian
+          #################
+          # Check matrix for tag list definition
+          flavor: |
+            latest=false
```
nitnelave, line 594: Latest should be false?

nitnelave (commented):
Looks really nice! Thanks for the hard work, I saw the stream of commits :)

A couple of minor comments, then we can try it in prod

.github/workflows/docker-build-static.yml
martadinata666: This part is for our dummy registry, it doesn't really push to the real registry. Then enable `#push: ${{ github.event_name != 'pull_request' }}` expect to fail in PR, we already know will work fine. Going to add some comments.

.github/workflows/docker-build-static.yml
martadinata666: Yes, the latest thing taken care on matrix, this part.
```
444     strategy:
445       matrix:
446         container: ["debian","alpine"]
447         include:
448           - container: alpine
449             platforms: linux/amd64,linux/arm64,linux/arm/v7
450             tags: |
451                   type=ref,event=pr
452                   type=semver,pattern=v{{version}}
453                   type=semver,pattern=v{{major}}
454                   type=semver,pattern=v{{major}}.{{minor}}
455                   type=semver,pattern=v{{version}},suffix=
456                   type=semver,pattern=v{{major}},suffix=
457                   type=semver,pattern=v{{major}}.{{minor}},suffix=
458                   type=raw,value=latest,enable={{ is_default_branch }}
459                   type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
460                   type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }},suffix=
461                   type=raw,value=latest,enable={{ is_default_branch }},suffix=
462                   type=raw,value={{ date 'YYYY-MM-DD' }},enable={{ is_default_branch }}
463                   type=raw,value={{ date 'YYYY-MM-DD' }},enable={{ is_default_branch }},suffix=
464           - container: debian
465             platforms: linux/amd64,linux/arm64,linux/arm/v7
466             tags: |
467                   type=ref,event=pr
468                   type=semver,pattern=v{{version}}
469                   type=semver,pattern=v{{major}}
470                   type=semver,pattern=v{{major}}.{{minor}}
471                   type=raw,value=latest,enable={{ is_default_branch }}
472                   type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
473                   type=raw,value={{ date 'YYYY-MM-DD' }},enable={{ is_default_branch }}
```

nitnelave:
LGTM , but it doesn't seem to pass :D

Is that related to your push comment?

Ah, I see how the intermediate images need to be pushed.

As long as the final images are not pushed, it's fine. Can you restore the push?

nitnelave:
Thanks!

PR #787, opened 2024-01-08
server: update ldap3_proto dependency

nitnelave:
This will fix the issue with some unhandled controls.


nitnelave:
This should help with #739

PR #803, opened 2024-01-18
Added maddy example config

elmodor:
Added an example configuration for the [Maddy Mail Server](https://maddy.email/)

### DIFF
--- a/example_configs/maddy.md
+++ b/example_configs/maddy.md
@@ -1,0 +1,83 @@
+# Configuration for Maddy Mail Server
+
+Documentation for maddy LDAP can be found [here](https://maddy.email/reference/auth/ldap/).  
+Maddy will automatically create an imap-acct if a new user connects via LDAP.  
+Replace `dc=example,dc=com` with your LLDAP configured domain.
+
+
+## Simple Setup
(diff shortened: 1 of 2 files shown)


elmodor:
Please hold, I found an issue with different mail clients. Will adjust the config to be more flexible

example_configs/maddy.md
```diff
@@ -0,0 +1,57 @@
...
+Documentation for maddy LDAP can be found [here](https://maddy.email/reference/auth/ldap/).  
+Maddy will automatically create an imap-acct if a new user connects via LDAP.
+
+### URL
+Change to your LLDAP URL
+```
+urls ldap://lldap.example.com:3890
+```
+
+#### Docker
```
nitnelave, line 12: No need for this section. People should know what the URL for their LLDAP server is, at least.

nitnelave (approved):

elmodor:
Since different mail clients seem to send different usernames to maddy (which then uses these for the LDAP auth) I extended the config with the different possibilities I found:
* Either a simple setup with no proper lookup
* A proper lookup because the username is send as E-Mail
* And allowing both (if different mail clients are used for example)

nitnelave (approved):
Great, thanks!

PR #807, opened 2024-01-21
server: Add the attribute schema to the attributes in graphql

nitnelave:
And make sure that we only request the schema once per top-level query


nitnelave:
@pixelrazor that will give you the schema along with the attributes. Should simplify your life.

PR #815, opened 2024-01-24
Fix docker-mailserver example

HighwayStar:
* Fixes following issues:
 - double braces around mail= filter cause: 
 ldap_search_ext: Bad search filter (-7)
 - too wide/upper level base DN cause, changed to ou= level helps   
 result: 53 Server is unwilling to perform 
 text: Unsupported group attribute for substring filter: "mail"


nitnelave (approved):
Thanks!

PR #818, opened 2024-01-26
app: update forms to use new components

### LINKED ISSUE #812: [CLEANUP] Use new form components across all components
New form components were added in #802 to simplify the html of forms. Where applicable, switch old form inputs over to the new components.

pixelrazor:
Most of the spots switched to the new components, though the login form didn't quite fit with the new components. Instead of forcing it to fit i left it out for the time being.

fixes #812

### DIFF
--- a/app/src/components/login.rs
+++ b/app/src/components/login.rs
@@ -1,5 +1,8 @@
 use crate::{
-    components::router::{AppRoute, Link},
+    components::{
+        form::submit::Submit,
+        router::{AppRoute, Link},
+    },
     infra::{
         api::HostService,
         common_component::{CommonComponent, CommonComponentParts},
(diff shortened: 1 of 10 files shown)


app/src/components/login.rs
```diff
@@ -155,68 +158,62 @@ impl Component for LoginForm {
...
-                    <Field
-                      class="form-control"
-                      class_invalid="is-invalid has-error"
-                      class_valid="has-success"
-                      form={&self.form}
-                      field_name="password"
-                      input_type="password"
-                      placeholder="Password"
-                      autocomplete="current-password" />
+              <form class="form center-block col-sm-4 col-offset-4">
```
pixelrazor, line 161: This file looks like it has a lot of changes. All i did was add the new submit component, but i moved the class attribute for the form and reduced all of the indentation by 1 level, which makes a pretty big diff

nitnelave (approved):

PR #825, opened 2024-02-01
app: create group attribute schema page

pixelrazor:
For #67

Mostly copy and paste of the user table. There's probably some common code here, but i opted to copy and try to share more code when these components are refactored with function components and/or better forms.

Also, i'm not sure what happened, but the name and type form fields on the user attribute form for messed up, so that is corrected here

EDIT: I swear the dropdown for type was in my screenshot... oh well

Screenshot 2024-02-01 9 20 16 AM
Screenshot 2024-02-01 9 20 34 AM


app/src/components/create_group_attribute.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+#[derive(Model, Validate, PartialEq, Eq, Clone, Default, Debug)]
+pub struct CreateGroupAttributeModel {
+    #[validate(length(min = 1, message = "attribute_name is required"))]
+    attribute_name: String,
+    #[validate(custom = "validate_attribute_type")]
+    attribute_type: String,
+    is_list: bool,
+    is_visible: bool,
```
nitnelave, line 46: On second thought, only privileged accounts can see a group's details, so this flag doesn't make sense either. That should be reflected in the backend, if not in the API. Could you create an issue for that?

app/src/components/create_group_attribute.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn validate_attribute_type(attribute_type: &str) -> Result<(), ValidationError> {
+    let result = AttributeType::from_str(attribute_type);
```
nitnelave, line 50: from_str(...).map_err( ValidationError(...))?;
Ok(())

app/src/components/group_schema_table.rs
```diff
@@ -0,0 +1,198 @@
...
+                match self.attributes {
+                    None => {
+                        log!(format!("Attribute {attribute_name} was  deleted but component has no attributes"));
+                        Err(anyhow!("invalid state"))
+                    }
+                    Some(_) => {
+                        self.attributes
+                            .as_mut()
+                            .unwrap()
+                            .retain(|a| a.name != attribute_name);
```
nitnelave, line 68: We don't actually check that we removed an element. Not that I expect it to happen (got a deletion response for an element we don't have), unless maybe the user spams the delete button. It's probably fine like that.

app/src/components/delete_group_attribute.rs
```diff
@@ -0,0 +1,172 @@
...
+            id={"deleteGroupAttributeModal".to_string() + &ctx.props().attribute_name}
+            tabindex="-1"
+            aria-labelledby="deleteGroupAttributeModalLabel"
+            aria-hidden="true"
+            ref={self.node_ref.clone()}>
+            <div class="modal-dialog">
+              <div class="modal-content">
+                <div class="modal-header">
+                  <h5 class="modal-title" id="deleteGroupAttributeModalLabel">{"Delete group attribute?"}</h5>
+                  <button
```
nitnelave, line 139: Should we disable the button while we have a request in flight?

nitnelave (approved):

app/src/components/create_group_attribute.rs
pixelrazor: Will do. I got thrown off at first when it looked like users can read visible group attributes in the backend handlers.

app/src/components/delete_group_attribute.rs
pixelrazor: it is - see line 114

app/src/components/group_schema_table.rs
pixelrazor: Yeah, i don't think it's necessary.

app/src/components/create_group_attribute.rs
pixelrazor: Done, and i also moved this function to schema.rs to share it with create_group_attribute too

app/src/components/create_group_attribute.rs
pixelrazor: #826

Issue #821, opened 2024-01-29
Attributes are not sorted, users are not sorted, or previous user didn't consume all the attributes


dvv:
**Describe the bug**

Panic "Attributes are not sorted, users are not sorted, or previous user didn't consume all the attributes" after migration.

**To Reproduce**
1. Migrate from another LDAP via the migration tool
2. Restart the server
3. Login

**Expected behavior**

Server just works.
Is there a way to repack the database (postgresql) to fix the problem?

nitnelave:
Hmm, I can't reproduce the panic. Could you post a sql dump of the DB, or send it to me privately on Discord? Don't worry, I won't be able to get the passwords.

dvv:
@nitnelave sorry, no, I don't own it.

Indeed I saw no sorting in migration tool hence user ids come assorted from the source LDAP.
At first I tried to sort them:
```diff
--- a/migration-tool/src/main.rs                                                                                                                                      
+++ b/migration-tool/src/main.rs
@@ -23,11 +23,12 @@ fn ask_generic_confirmation(name: &str, message: &str) -> Result<bool> {
 fn get_users_to_add(users: &[User], existing_users: &[String]) -> Result<Option<Vec<User>>> {
     let existing_users = HashSet::<&String>::from_iter(existing_users);
     let num_found_users = users.len();
-    let input_users: Vec<_> = users
+    let mut input_users: Vec<_> = users
         .iter()
         .filter(|u| !existing_users.contains(&u.user_input.id))
         .map(User::clone)
         .collect();
+    input_users.sort_by(|u1, u2| u1.user_input.id.cmp(&u2.user_input.id));
     println!(
         "Found {} users, of which {} new users: [\n  {}\n]",
         num_found_users,
```
but the error persisted.

Then I commented out the assertion:
```diff
--- a/server/src/domain/sql_user_backend_handler.rs
+++ b/server/src/domain/sql_user_backend_handler.rs
@@ -145,11 +145,13 @@ impl UserListerBackendHandler for SqlBackendHandler {
         let mut attributes_iter = attributes.into_iter().peekable();
         use itertools::Itertools; // For take_while_ref
         for user in users.iter_mut() {
+            /*
             assert!(attributes_iter
                 .peek()
                 .map(|u| u.user_id >= user.user.user_id)
                 .unwrap_or(true),
                 "Attributes are not sorted, users are not sorted, or previous user didn't consume all the attributes");
+            */

             user.user.attributes = attributes_iter
[...]

PR #827, opened 2024-02-02
server: Fix panic due to database collation

nitnelave:
When the database's collation is not "C", the DB order is not the same as the Rust order. As such, asserting that the elements are in increasing order fails. However, since both queries get the order from the database, they should be in the same order.

With too many users, the query had a giant filter `IN (u1, u2, u3, ...)`. In PostgreSQL, we can pass the users as an array instead, but that doesn't work with SQLite. Instead, we repeat the filter from the previous query to get the same users/groups, as a subquery.


nitnelave:
Fixes #821

PR #856, opened 2024-03-07
Update docker-build-static.yml add FreeBSD static build

n-connect:
Add x86_64-unknown-freebsd static build, create tar.gz from static builds


nitnelave:
Pardon my ignorance, but since this is a statically linked binary without dependencies, can't you run the Linux version of the binary? How is freebsd special in that respect?

nitnelave:
Oh, duh, different kernels, the syscalls are different. Pardon the derp

.github/workflows/docker-build-static.yml
```diff
@@ -34,7 +34,7 @@ env:
 ### upload artifacts
 
 # build-bin
-## build-armhf, build-aarch64, build-amd64 , create binary for respective arch
+## build-armhf, build-aarch64, build-amd64 , x86_64-unknown-freebsd create binary for respective arch
```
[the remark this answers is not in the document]
nitnelave: Can you rephrase this to remove the mention of the job names and just talk about the architectures?

nitnelave:
You'll need to install rust for freebsd as well:  "consider downloading the target with `rustup target add x86_64-unknown-freebsd`"

n-connect:
Hi @nitnelave 

Yes, your are right. As there are different low level kernel elements when you change the CPU/SOC family: armv7 aarm64 vs amd64, here even a completely different OS underlying. But luckily its a Rust code, so after a `rustup target add x86_64-unknown-freebsd` it should just compile it. If a toolchain exist in Rust/cargo/cross theoretically it (compilation) should just work. Checkout other codes like this https://github.com/nextcloud/notify_push/releases/tag/v0.6.9

Answering the last one:
> It's not AMD, can you fix the binary name?

It is amd64 and actual build start with this `cargo --build x86_64-unknown-freebsd` I used the naming based on that and the other already used naming conventions in the "docker-build-static.yml". I'm happy to change I think the tar.gz filename should stand

n-connect:
@nitnelave,

Based on your review, the build name changes: instead of "build_x86_freebsd" of "x86_64-unknown-freebsd" in the commended line 37.

I think a `rustup target add x86_64-unknown-freebsd` line should be added somewhere too, at least on my fork the GitHub Actions are failed because lack of that.

nitnelave (approved):

n-connect:
It seems there's a linker issue (ld) in the building docker image. I've checked [Dockerfile.dv](https://github.com/lldap/lldap/blob/main/.github/workflows/Dockerfile.dev). There you are specifying cc compilers per target. Following the links, it could be [x86_64-linux-muslx32-cross.tgz](https://more.musl.cc/x86_64-linux-musl/x86_64-linux-muslx32-cross.tgz) it needs. Although I'm not sure I've never used them like that before. I've used to target, or installed cross. Will check with a local build using rust docker.

martadinata666:
So freebsd approach wise likely want to use CLANG rather normal gcc. The best sample I can get is from https://github.com/smartmontools/docker-build/blob/master/Dockerfile#L77. That in simple term use clang, fetch freebsd libs. Then following suit to development file
```
    CARGO_TARGET_ARMV7_UNKNOWN_LINUX_MUSLEABIHF_LINKER=armv7l-linux-musleabihf-gcc \
    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-musl-gcc \
    CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=x86_64-linux-musl-gcc \
    CARGO_TARGET_X86_64_UNKNOWN_FREEBSD_LINKER=clang++ \
    CC_armv7_unknown_linux_musleabihf=armv7l-linux-musleabihf-gcc \
    CC_x86_64_unknown_linux_musl=x86_64-linux-musl-gcc \
    CC_aarch64_unknown_linux_musl=aarch64-linux-musl-gcc \
    CC_x86_64_unknown_freebsd=clang++
```
Maybe just clang? or clang++? Trial error begins.

There is also https://github.com/marketplace/actions/freebsd-vm but this route gonna be start from scratch.

n-connect:
@martadinata666 

Thanks for the details, checked the smartmontools and the  Wezm repo too. Unfortunately ran into several linked problems while tried to build one. In this direction this seems to most up-to-date [Docker x86_64-unknown-freebsd](https://github.com/mcandre/snek/blob/master/docker/x86_64-unknown-freebsd/Dockerfile)  by mcandre.

Honestly pervious projects I tried to get a FreeBSD build, were constucted different from Github actions point of view, editing only the YAML files of github action to add target x86_64-unknown-freebsd was simply enough.

Here I had to go under the hood, so there are three ways:
1. Use cross, justt tested successfully. Caveats: cross wants to spin up another docker -> ghcr.io/cross-rs/x86_64-unknown-freebsd . To be able to do, one needs to enable the Docker-in-Docker. It needs from host to run with the docker socket passed into the building Docker (lldap_dev), also docker need to be installed, and cross installed by passing and ENV that Docker-in-Docker will happen

2. Building freebsd-clang. Finally I've had the same linker error messages as before
3. Building freebsd-cc. Haven't tried yet, seemed the most complex for me.

martadinata666:
When freebsd-cc will complicates stuff as it need rebuild many things just to achieve x86_64 platform, thankfully  musl had cross cc available.

I also want to suggest using cross action, but instead doing inside this main yml. Do separately and using workflow call to build freebsd.

```
name: build freebsd

on:
  workflow_call:

env:
  CARGO_TERM_COLOR: always

jobs:
  build-bin:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        target: [x86_64-unknown-freebsd]

    steps:
      - uses: actions/checkout@v3
      - uses: actions/cache@v3
        with:
          key: lldap-bin-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
          restore-keys: |
            lldap-bin-${{ matrix.target }}-
      - name: Install podman
        run: |
             sudo apt update
             sudo apt install -y podman
      - name: Use cross?
        run: cargo install cross --git https://github.com/cross-rs/cross
      - name: Build with cross
        run: /home/runner/.cargo/bin/cross build --release --target x86_64-unknown-freebsd -p lldap -p lldap_migration_tool -p lldap_set_password
      - name: check path
        run: ls -alr target/release

      - name: upload ${{ matrix.target}} lldap artifacts
        uses: actions/upload-artifact@v3
        with:
          name: ${{ matrix.target}}-lldap-bin
          path: target/${{ matrix.target }}/release/lldap

      - name: upload ${{ matrix.target }} migration tool artifacts
        uses: actions/upload-artifact@v3
        with:
          name: ${{ matrix.target }}-migration-tool-bin
          path: target/${{ matrix.target }}/release/lldap_migration_tool

      - name: upload ${{ matrix.target }}-set-password-bin
        uses: actions/upload-artifact@v3
        with:
          name: ${{ matrix.target }}-lldap-set-password-bin
          path: target/${{ matrix.target }}/release/lldap_set_password
```

n-connect:
@nitnelave

I have successful complete builds for FreeBSD by downloading/updating the actual [Dockerfile.dev](https://github.com/lldap/lldap/blob/main/.github/workflows/Dockerfile.dev):
- installing docker.io
- adding rust x86_64-unknown-freebsd
- setting environment variable about Docker-in-Docker
- install cross and wasm-pack via cargo
- build the Docker, run the Docker and build app via cross (+ app/build.sh)
- pieces tar.gz'd out from docker: the lldap binaries, and the app/build.sh results. Uploaded [here](https://github.com/n-connect/rustd-hbbx/blob/main/x86_64-freebsd_lldap.tar.gz). The bootstrap and font files arer missing, I beleive those are put together with actions/upload-artifact@v4 (fixme)

Help me out: how can I find what actions/upload-artifact@v4 do, or how the additional fonts and bootstrap files get into under app/static & app/pkg for a release?

The modified [Dockerfile.dev](https://github.com/n-connect/rustd-hbbx/blob/main/Dockerfile.dev) is in another temporary repo, to not contaminate this PR

Right now it is out of the way of current static build, but can be integrated:
- leave out changes from Dockerfile.dev, which are in [docker-build-static.yml](https://github.com/lldap/lldap/blob/main/.github/workflows/docker-build-static.yml) already
- modify docker-build-static.yml to have DinD instructed cross installed
- add line to cross build x86_64-unkown-freebsd
- BUT this docker (job: build-bin) need to be run via `docker run -v /var/run/docker.sock:/var/run/docker.sock ...` [FAQ - cross-rs/cross Wiki ](https://github.com/cross-rs/cross/wiki/FAQ#docker-in-docker)

Finally: is this important, or is it this [LICENCE](https://github.com/lldap/lldap/blob/main/LICENSE) file it is talking about?
`[INFO]: Installing wasm-bindgen...
[INFO]: Optimizing wasm binaries with `wasm-opt`...
[INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
[INFO]: :-) Done in 1m 45s
[INFO]: :-) Your wasm pkg is ready to publish at /tmp/lldap/app/pkg.`

Should I provide the licence file? I'm not famailiar with that part.

n-connect:
@martadinata666 

Yeah, had similar idea, but were thinking about extending the original static-build YAML. What I didn't catched really, how the existing static build dockers called in life and by what. If there's no issue with running them via mapping /var/run/docker.sock into them to do DinD, that seems to be the simplest extension for this.

However if you do have some more details on the freebsd-cc way to share with, I'm interested in that too, it could help the BSD/SunOS/IllumOS area further.

> When freebsd-cc will complicate stuff as it needs to rebuild many things just to achieve the x86_64 platform, thankfully musl had cross cc available.
> 
> I also want to suggest using cross action, but instead doing inside this main yml. Do it separately and use workflow call to build freebsd on this main yml.
> 
> ```

example_configs/lldap
martadinata666, line 1: What is this file? Text?

.github/workflows/Dockerfile.dev
```diff
@@ -37,5 +37,19 @@ RUN rustup target add x86_64-unknown-linux-musl && \
...
+    apt clean && \
+    rm -rf /var/lib/apt/lists/*
+##  Error starting systemctl daemon, lack of PID 1 cause of docker.
+##  But docker cli commands works cause of host running with -v /var/run/docker.sock:/var/run/docker.sock
+#   systemctl start docker && \
+
+### Add cross with Docker-in-Docker
+
+ENV CROSS_CONTAINER_IN_CONTAINER=true
+RUN cargo install cross && \
```
martadinata666, line 51: By adding cross to  development image will this make freebsd binary generated? Or need someworkaround? As last conversation seems need to mount /var/run/docker.sock that like need different CI implementation.

martadinata666:
Kindly check the run result https://github.com/lldap/lldap/actions/runs/8360008854

example_configs/lldap
n-connect: Yes, its a rc.d style service script used in FreeBSD to run & control daemons/services. At first upload I made and error, need to be fixed to change the content in [my fork repo](https://github.com/n-connect/lldap/blob/main/example_configs/lldap). However I do not see the right content through the PR commits or the Files/changed tab (here).

.github/workflows/Dockerfile.dev
n-connect: Yes, possible. It seems your suggestion with a separate yml and one plus workflow call will be the solution  here.

Maybe the best to close this PR and create a new one with that way. That can fix the strange file issue with the lldap rc.d script - where the latest commit not appears here, but exist in the PR source/fork repo.

martadinata666:
PR closed due weird freebsd rc.d taken as binary rather than text.

PR #883, opened 2024-03-28
FreeBSD rc.d service script

n-connect:
Follow-up after closing PR #856 to have this service script independently. Manual/local **lldap FreeBSD build** available [here](https://github.com/n-connect/rustd-hbbx/blob/main/x86_64-freebsd_lldap-0.5.1.tar.gz)

The script in its actual form takes into consideration, that lldap release has a complete structure:
- app subdir for web content, 
- possible data subdir for other elements
- and it expects the above sudirs to be in the same folder as the main lldap binary.

Use:
- extract (the future) FreeBSD release package into /usr/local/ -> so your files will be under /usr/local/lldap_server/
- save/copy this rc.d script file into /usr/local/etc/rc.d/
- finally `cat lldap_enable=YES >> /etc/rc.conf`
- the service script set to run the lldap server as "www" user - make sure the whole lldap_server directory is accessible/runnable by "www". Simplest to run `chown -R www:www /usr/local/lldap_server`

Note:
When ldap will accept config file parameter speficying the web content dir, this file need to be updated as well.


nitnelave:
Sorry for the late review.

I feel that all of the comments in this PR (assumptions, how-to and so on) should be in that file as comments. WDYT?

nitnelave:
I just think that most people are going to look at the file, not the PR, and will miss the assumptions. It'll be more discoverable to include it in the file itself.

n-connect:
> I just think that most people are going to look at the file, not the PR, and will miss the assumptions. It'll be more discoverable to include it in the file itself.

Yepp, done - check it out both the PR description and the rc.d file has the info

example_configs/lldap
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
nitnelave, line 1: Shouldn't this file be named lldap.rc.d ? Right now it's just lldap and a bit confusing

nitnelave (changes requested):

nitnelave:
Sorry for the delay in reviewing! Life got in the way

example_configs/lldap
n-connect: I think no, cause then you have to rename it back to "lldap" when you copy it over to "/usr/local/etc/rc.d/" dir

n-connect:
> Sorry for the delay in reviewing! Life got in the way

No worries.
I'm planning some other PR's to raise around let be the web artifact its own configurable http_root directory...

example_configs/lldap
nitnelave: Still, it's better to rename it to something clearer. Maybe "freebsd_lldap.rc.d", and then add in the instructions that you should rename it. Otherwise, think about it: the path to this file is "example_configs/lldap": what is it about? How can a freebsd user find the config?

nitnelave: Speaking of discoverability, can you add it to the readme as well? It's not exactly a "supported integration", but it'll fit there just fine.

n-connect: @nitnelave 

If you want to add FB install instuctions into to the readme file, that's nice. With that step we should just delete them (the install instructions) from the rc.d file itself as it should not go there at the first place. 

People insterested FreeBSD lldap build can find a fully working build with the rc.d script inside included in my [temporary repo](https://github.com/n-connect/rustd-hbbx/blob/main/x86_64-freebsd_lldap-0.5.1.tar.gz). Just until a working release script/build can be included here :)

Please do make up your mind, and write me back what you'd like to close this PR. :)

nitnelave: No, I was just thinking of adding a link to the (renamed) file, in the installation section (e.g. "for freebsd, you have a service file there")

n-connect: Ok. What about adding this linking into readme.md with the renamed file and have a separate text file for install instructions (freebsd_install.txt), so the rc.d script can be as it should be?

nitnelave: In that case, what about a freebsd directory in example configs, with a readme and the rc script?

n-connect: Sounds great, lets do that.

One question though, what's your actual way of adding relative links into (Readme).md files? I've tried to do it in a separate repo, but never worked for me the way the [GitHub docs described](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links)

nitnelave: It's worked that way for images at least, I guess for everything it would work as well.

nitnelave:
Looks good! Can you add a link in the readme? Maybe the installation section.

example_configs/lldap
n-connect: Ok, all done, pls. check them. The main Readme.md files internal link, I'm not sure..

n-connect:
> Looks good! Can you add a link in the readme? Maybe the installation section.

It's there, check it

README.md
```diff
@@ -208,6 +209,14 @@ Each package places lldap's configuration file at `/etc/lldap.toml` and offers
 [systemd service](https://wiki.archlinux.org/title/systemd#Using_units)
 `lldap.service` to (auto-)start and stop lldap.
 
+### With FreeBSD
+
+You can also install it as a rc.d service in FreeBSD, see
+[FreeBSD-install.md](example_configs/freebsd/freebsd-install.md).
+
+The rc.d script file 
+[rcd_lldap](example_configs/freebsd/rcd_lldap).
```
nitnelave, line 218: Link should be to `rc.d_lldap`, not `rcd_lldap`.

You can check the links by clicking the three dots ... in the upper right corner, then View the file: you'll get the rendered markdown, and you can click on the links (which, being relative, will find the files in your PR)

nitnelave (changes requested):

README.md
n-connect: Yepp, mistyped, beside looking previews. Thanks! Fixed.

nitnelave (approved):

nitnelave:
Thanks a lot, merged!

PR #892, opened 2024-04-06
Docker Compose update

bsncubed:
Hi, I updated the docker compose part of the README file to include SMTP environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ services:
       # You can also set a different database:
       # - LLDAP_DATABASE_URL=mysql://mysql-user:password@mysql-server/my-database
       # - LLDAP_DATABASE_URL=postgres://postgres-user:password@postgres-server/my-database
-      # If useing SMTP, set the following variables
+      # If using SMTP, set the following variables
       # - LLDAP_SMTP_OPTIONS__ENABLE_PASSWORD_RESET=true
       # - LLDAP_SMTP_OPTIONS__SERVER=smtp.example.com
       # - LLDAP_SMTP_OPTIONS__PORT=465 #Check your smtp providor's documentation for this setting 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,15 @@ services:
       # You can also set a different database:
       # - LLDAP_DATABASE_URL=mysql://mysql-user:password@mysql-server/my-database
       # - LLDAP_DATABASE_URL=postgres://postgres-user:password@postgres-server/my-database
+      # If useing SMTP, set the following variables
+      # - LLDAP_SMTP_OPTIONS__ENABLE_PASSWORD_RESET=true
+      # - LLDAP_SMTP_OPTIONS__SERVER=smtp.example.com
+      # - LLDAP_SMTP_OPTIONS__PORT=465 #Check your smtp providor's documentation for this setting 
+      # - LLDAP_SMTP_OPTIONS__SMTP_ENCRYPTION=TLS #How the connection is encrypted, either "NONE" (no encryption), "TLS" or "STARTTLS".
+      # - LLDAP_SMTP_OPTIONS__USER=no-reply@example.com #The SMTP user, usually your email address
+      # - LLDAP_SMTP_OPTIONS__PASSWORD=PasswordGoesHere #The SMTP password
+      # - LLDAP_SMTP_OPTIONS__FROM=no-reply <no-reply@example.com> #The header field, optional: how the sender appears in the email. The first is a free-form name, followed by an email between <>.
+      # - LLDAP_SMTP_OPTIONS__TO=admin <admin@example.com> # Same for reply-to, optional.
 ```
 
 Then the service will listen on two ports, one for LDAP and one for the web


README.md
```diff
@@ -161,6 +161,15 @@ services:
       # You can also set a different database:
       # - LLDAP_DATABASE_URL=mysql://mysql-user:password@mysql-server/my-database
       # - LLDAP_DATABASE_URL=postgres://postgres-user:password@postgres-server/my-database
+      # If using SMTP, set the following variables
+      # - LLDAP_SMTP_OPTIONS__ENABLE_PASSWORD_RESET=true
+      # - LLDAP_SMTP_OPTIONS__SERVER=smtp.example.com
+      # - LLDAP_SMTP_OPTIONS__PORT=465 #Check your smtp providor's documentation for this setting 
+      # - LLDAP_SMTP_OPTIONS__SMTP_ENCRYPTION=TLS #How the connection is encrypted, either "NONE" (no encryption), "TLS" or "STARTTLS".
```
nitnelave, line 168: Maybe we should expand a bit on this bit of config, it's often confusing due to the wrong names used in email. SMTP settings usually list "TLS" when they mean STARTTLS, and then SSL means TLS.

bsncubed: Yeah, good idea. I'll have a go at it over the next few days. Should it be in the main readme, or a new file in the docs folder?

nitnelave: Maybe in the main readme, and in the config template at the top level

polarathene: > SMTP settings usually list "TLS" when they mean STARTTLS, and then SSL means TLS.

Do you have some references for that?

SSL / TLS is usually used interchangeably (_although it should be TLS these days, users tend to be more familiar with using the term SSL in my experience_).

- STARTTLS is explicit TLS, aka opportunistic (_connection starts unencrypted and may potentially upgrade to TLS to be encrypted_), ports 25 and 587
- TLS is implicit TLS, aka mandatory (_connection is established as TLS from the start, like with HTTPS_), port 465

Some service providers may offer non-standard ports, but the expectation for ports are:
- Port 25 is for receiving inbound mail delivery, or sending mail outbound to another MTA.
- Ports 587 and 465 are for mail submission (_client submits mail to be sent by the MTA, or the MTA connects to a relay service like SendGrid_), these require authentication.

[`lettre` offers 4 choices](https://docs.rs/lettre/latest/lettre/transport/smtp/client/enum.Tls.html), `None`, `Opportunistic`, `Required` (_StartTLS but upgrade to TLS is mandatory_), `Wrapped` (_implicit TLS_). For authentication with TLS the last two make sense, you shouldn't need `Opportunistic` for connections where authentication is required, and `None` is only relevant for troubleshooting / testing without TLS.

---

Given the above, what are you proposing to adjust here? You should stick to `TLS` and not confuse users by associating it to `STARTTLS` while also referring to `SSL` for port 465?

Perhaps you're referring to some outdated info around the web regarding port 465?

Before an RFC in 2018, the port was historically assigned for SMTPS (_no authentication required, just implicit TLS for port 25 as an alternative_). That was during a draft over 20 years ago, but while in the draft state it was implemented in software/services, even though AFAIK it never actually was published officially into an RFC standard.

It didn't really make sense for SMTP over TLS on a separate port for delivery.... so to compliment the existing port 587 intended for mail submission, we got an implicit TLS equivalent with port 465 standardized in 2018. However by that point in time, some service providers offered this port or documented/blogged about it with the historical use, I haven't checked if those sources have been updated since, but I often came across the misunderstanding.

Some linux distros may not have updated the definition in `/etc/services` to have port 465 assigned as `submissions`, but I know some have.

`NONE`, `TLS`, `STARTTLS` is clear enough.

If the user isn't familiar with StartTLS, they should understand TLS and None choices well enough, and those are the ones they'll likely want today, or they follow the documentation of a service provider that mentions StartTLS for the port of choice and go with that.

nitnelave: I agree with you that SSL/TLS should be used interchangeably, but if you look at SMTP configuration guides, they often use SSL to mean implicit TLS and TLS to mean starttls. 

See for example the Google guide: https://developers.google.com/gmail/imap/imap-smtp
Protonmail has a weird call-out https://proton.me/support/smtp-submission
Zoho has a section on SSL and one on TLS: https://www.zoho.com/mail/help/zoho-smtp.html?zredirect=f&lb=nl&zsrc=bheader#alink2
Similarly, I've seen many SMTP clients have SSL and TLS configuration options.

I just want to make the distinction clear to users and how they're used.
I agree that it's not the right meaning (which is why the enum values are the right ones), but I want to bridge to how they're actually used in the wild, in the specific context of SMTP

polarathene: Thank you for sharing those references, I'm quite surprised Google has that messed up like that too, and Protonmail to be encouraging STARTTLS over TLS 🤔 

It's honestly bizarre how these leading brands are failing so hard with their docs / advice 😭 

## Suggestions / references

If it's helpful, you could also reference the [Docker Mail Server docs on this subject](https://docker-mailserver.github.io/docker-mailserver/v14.0/config/security/understanding-the-ports/), I'm a maintainer there and spent a fair bit of time looking into the mess myself a while back 😅 

I also made some notes myself that I [committed as comments for maintainers](https://github.com/docker-mailserver/docker-mailserver/blob/22383c28e7cf6d38b674f8a0d28ff522bbad3929/target/scripts/helpers/relay.sh#L288-L349), just looked that up to provide as an additional reference and I also cited the SSL / TLS mix you mentioned there. So there's some additional providers that add to the list, along with misleading/outdated blog posts (_which may omit original published date_).

Maybe something like:

```yaml
      # NOTE: Use the correct PORT and SMTP_ENCRYPTION pairs (Port 465 recommended):
      #  Port 587 --> STARTTLS (sometimes mistakenly called TLS)
      #  Port 465 --> TLS (sometimes referred to as SSL)
      # - LLDAP_SMTP_OPTIONS__PORT=465 #Check your smtp providor's documentation for this setting 
      # - LLDAP_SMTP_OPTIONS__SMTP_ENCRYPTION=TLS #How the connection is encrypted, either "NONE" (no encryption), "TLS" or "STARTTLS".
```

For DMS we presently only support port 587 for outbound relay hosts (_unless the user manually adjusts some config to support port 465_), but we make the ports inbound to DMS pretty clear in our [example `compose.yaml`](https://github.com/docker-mailserver/docker-mailserver/blob/22383c28e7cf6d38b674f8a0d28ff522bbad3929/compose.yaml#L8-L15):

```yaml
    # More information about the mail-server ports:
    # https://docker-mailserver.github.io/docker-mailserver/latest/config/security/understanding-the-ports/
    ports:
      - "25:25"    # SMTP  (explicit TLS => STARTTLS, Authentication is DISABLED => use port 465/587 instead)
      - "143:143"  # IMAP4 (explicit TLS => STARTTLS)
      - "465:465"  # ESMTP (implicit TLS)
      - "587:587"  # ESMTP (explicit TLS => STARTTLS)
      - "993:993"  # IMAP4 (implicit TLS)
```

Maybe for your example it'd be simpler to just copy/paste the two ENV pairs to show a port 465 vs 587 example? Perhaps with a comment that some services may incorrectly use STARTTLS on port 465

---

## Rant

Just checked the [SendGrid blog post again](https://sendgrid.com/en-us/blog/whats-the-difference-between-ports-465-and-587#span-stylefontweight-400port-465-message-submission-over-tls-protocolspan) and the publish date is now May 2023, while my git commit history was 2 years ago, so you can see that the date is misleading users as if it were recent content, while the post content itself still informs users with misinformation on port 465 (_despite referencing RFC 8314, and mentioning the reassignment, they didn't seem to quite understand the difference from SMTPS_), meanwhile [another blog post of theirs on StartTLS](https://sendgrid.com/en-us/blog/what-is-starttls) still has this misunderstanding:

> _Port 465 is the second most commonly used port for StartTLS._

>  _If the connection is not encrypted, the mail will be blocked from sending. This method is much more secure than Opportunistic TLS, but does lead to more mail being dropped._

>_Unencrypted and TLS connections are accepted on ports 25, 587, and 2525. Or, you can connect through SSL on port 465._

At the end it implies that they're describing configuring how to have SendGrid deliver mail over a port to a recipient... which is weird since it should always be via port 25 outbound for a MTA delivery. The link for "Enforced TLS" doesn't resolve to the API docs anymore, probably due to changes since SendGrid was acquired by Twilio, claims to be published since Jan 2023 though 🤷‍♂️

nitnelave: For what it's worth, I now have much better error messages that should point people to the right combination of port and protocols, so the clarification is not as crucial anymore (if you still want to merge this :D)
We can instead just use the implicit TLS/465 values by default in the example.

PR #905, opened 2024-04-26
pfsense.md, add warning about error about `Organizational Units` 

Torstein-Eide:
Add warning about `Organizational Units` that will fail. 
Minor formating

### DIFF
--- a/example_configs/pfsense.md
+++ b/example_configs/pfsense.md
@@ -40,7 +40,11 @@ ou=people
 
 > [!Note]
 > The `Select a container` seach fuction will not work for selecting containers. You enter the `Authentication containers` directly into the text field.
-> This is do to Pfsense running the following filter `"(|(ou=*)(cn=Users))"`, and `Organizational Units` is currently not supported.
+> This is due to Pfsense running the following filter `"(|(ou=*)(cn=Users))"`, and `Organizational Units` is currently not supported.Could not connect to the LDAP server. Please check the LDAP configuration.
+
+> [!WARNING] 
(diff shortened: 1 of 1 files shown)


example_configs/pfsense.md
```diff
@@ -27,18 +32,21 @@ Entire Subtree
...
 ### Authentication containers
 
 ```
 ou=people
 ```
 
-Note: The `Select a container` box may not work for selecting containers. You can just enter the `Authentication containers` directly into the text field.
+> [!Note]
+> The `Select a container` seach fuction will not work for selecting containers. You enter the `Authentication containers` directly into the text field.
+> This is do to Pfsense running the following filter `"(|(ou=*)(cn=Users))"`, and `Organizational Units` is currently not supported.
```
nitnelave, line 43: due to

example_configs/pfsense.md
```diff
@@ -112,6 +126,9 @@ Go to `System > User Manager > Settings` page. Add your LLDAP server configurati
 
 pfSense includes a built-in feature for testing user authentication at `Diagnostics > Authentication`. Select your LLDAP server configuration in the `Authentication Server` to test logins for your LLDAP users. The groups (only the ones you added to pfSense) should show up when tested.
 
+> [!WARNING] 
+> When running `Save and test`, the `Attempting to fetch Organizational Units from` will fail. This is do to Pfsense running the following filter `"(|(ou=*)(cn=Users))"`, and `Organizational Units` is currently not supported.
```
nitnelave, line 130: due to

nitnelave (changes requested):
Thanks!

nitnelave (approved):

PR #907, opened 2024-05-02
Add user_id for nextcloud example

ibizaman:
Context for why I believe this is a good idea:

- When logging with a new user "ibizaman" through the database backend, the ID used by Nextcloud is `ibizaman`.
- Through the LDAP backend, Nextcloud chooses the `UUID` field.
- Through the SSO backend, integrated with LDAP, Nextcloud chooses `ibizaman`.

I noticed this issue while migrating users from database backend, to LDAP backend, to SSO backend where the ID was inconsistent. Also, it means that if a user was created through LDAP and tries to login through SSO, a new user will be created!

This inconsistency makes no sense IMO. By overriding the `UUID` field to be the `user_id` field, the ID assigned to a new user is consistent whatever the backend used.

### DIFF
--- a/example_configs/nextcloud.md
+++ b/example_configs/nextcloud.md
@@ -74,6 +74,7 @@ occ ldap:set-config s01 ldapUserDisplayName displayname
 occ ldap:set-config s01 ldapUserFilterMode 1
 occ ldap:set-config s01 ldapUuidGroupAttribute auto
 occ ldap:set-config s01 ldapUuidUserAttribute auto
+occ ldap:set-config s01 ldapExpertUsernameAttr user_id
 ```
 With a bit of of luck, you should be able to log in your nextcloud instance with LLDAP accounts in the `nextcloud_users` group.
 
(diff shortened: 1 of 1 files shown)


nitnelave (approved):
Nice!

kesselb:
Please be careful when using `ldapExpertUsernameAttr`.

> When configuring this, be aware that the username in Nextcloud is considered immutable and cannot be changed afterwards. This can cause issues when using an attribute that might change, e.g. the email address of a user that will get changed during name change.

https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/user_auth_ldap.html#expert-settings

> I noticed this issue while migrating users from database backend, to LDAP backend, to SSO backend where the ID was inconsistent. Also, it means that if a user was created through LDAP and tries to login through SSO, a new user will be created!

That sounds more like a configuration issue in user_ldap or user_saml.

nitnelave:
User IDs in LLDAP are immutable, so it's reasonable to use them as an attribute. And it leads to a more readable folder architecture than UUIDs.

ibizaman:
@kesselb I agree that’s a concern but in this case, the LLDAP ID is indeed unique and not modifiable so it does the bill. And importantly it’s compatible with other methods of authentication for Nextcloud.

PR #923, opened 2024-06-25
Update jellyfin.md

ThorpeJosh:
Great project. Many thanks to the maintainers.

I have added some missing fields and restructured to match the jellyfin plugin UI.

### DIFF
--- a/example_configs/jellyfin.md
+++ b/example_configs/jellyfin.md
@@ -36,10 +36,6 @@ uid
 ```
 uid
 ```
-### LDAP Password Attribute
-```
-userPassword
-```
 
 ### LDAP Admin Base DN
 The DN to search for your admins.
--- a/example_configs/jellyfin.md
+++ b/example_configs/jellyfin.md
@@ -1,10 +1,12 @@
 # Configuration for Jellyfin
 
-Replace `dc=example,dc=com` with your LLDAP configured domain.
+Replace all instances of `dc=example,dc=com` with your LLDAP configured domain.
 
+## LDAP Server Settings
 ### LDAP Bind User
+Create an ldap user for Jellyfin to run search queries (and optionally reset passwords). For example `jellyfin_bind_user`
 ```
-uid=admin,ou=people,dc=example,dc=com
+uid=jellyfin_bind_user,ou=people,dc=example,dc=com
 ```
 
 ### LDAP Base DN for searches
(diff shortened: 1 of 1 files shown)


example_configs/jellyfin.md
```diff
@@ -1,42 +1,47 @@
...
 ```
-(memberof=cn=media,ou=groups,dc=example,dc=com)
+uid
 ```
-
-Otherwise, just use:
+### LDAP Password Attribute
 ```
-(uid=*)
+userPassword
```
nitnelave, line 41: Is that the main difference? It's a bit hard to tell given that the fields were reordered. 

Anyway, this field is not, will not and cannot be available through LLDAP by design.

nitnelave (changes requested):
I'm all for a little cleaning up, but the userPassword field cannot work. Have you tried it?

example_configs/jellyfin.md
ThorpeJosh: I agree it is difficult to see the diff. I recommend just looking at the entire rendered file side by side.

90% of changes is just reordering and adding some `## Headings` to match the Jellyfin UI sections.
The only modification to existing was to change the default bind user from `admin` to `jellyfin_bind_user`.
The only addition was adding the following block:
```markdown
### LDAP Password Attribute
\```
userPassword
\```
```

The reason I added it in the pr is:
- It was the only attribute from jellyfin UI that wasn't listed in `jellyfin.md`
- `userPassword` is the default jellyfin value and it is functioning as expected for me.
- It is mentioned to at the bottom of `jellyfin.md` under the `password change` section:
```markdown
`LDAP Password Attribute` Needs to be set to `userPassword`
```

ThorpeJosh: @nitnelave Given that `userPassword` is not available in lldap, did you want me to remove my addition and the last line of the `Password change` section introduced in c440df6 ?

nitnelave: The last line is fine: you can change passwords through jellyfin, but you can't read them (which the attribute is about) 

So yes, please remove the `userPassword` section, the rest of the PR looks fine

ThorpeJosh: I've removed the requested section. Let me know if anything else needs changing.

nitnelave (approved):
Thanks, LGTM !

nitnelave (approved):

PR #935, opened 2024-07-22
docs: fix example of ldap dex integration

fengxsong:
Host and optional port of the LDAP server in the form "host:port".


nitnelave (approved):
Thanks!

Issue #937, opened 2024-07-23
[BUG] generate_secrets.sh does not run when default shell is dash


ChevySSinSD:
**Describe the bug**
The generate_secrets.sh does not run in recent Debian-based distributions, generating this error:
`./generate_secrets.sh: 3: Syntax error: "(" unexpected`

**Additional Context**
The shebang on line 1 calls the default shell: `#! /bin/sh`

However, the function definition on line 3 uses the bash convention, which is not compatible with dash, the default shell for recent Debian-based linux distributions (and others):
`function print_random () {`

I have modified generate_secrets.sh in two separate ways, both solutions work on Debian:
- Change the shebang on line 1 to `#! /bin/bash`
- Change line 3 to `print_random () {` (i.e., remove "function" in front of the function definition). Reference: [Function Definition Command](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_05).

nitnelave:
Sounds good! Can you make a PR with change number 2? Let's keep it POSIX and as compatible as possible.

ChevySSinSD:
Will do

PR #938, opened 2024-07-23
Update generate_secrets.sh

ChevySSinSD:
Updated print_random function definition to be compatible with multiple default shells.

#937


nitnelave (approved):
Thanks!

PR #940, opened 2024-07-30
UI: User attributes form

nitnelave:
It's still buggy...

### DIFF
--- a/app/src/components/form/attribute_input.rs
+++ b/app/src/components/form/attribute_input.rs
@@ -1,0 +1,190 @@
+use crate::{
+    components::form::{date_input::DateTimeInput, file_input::JpegFileInput},
+    infra::{schema::AttributeType, tooltip::Tooltip},
+};
+use web_sys::Element;
+use yew::{
+    function_component, html, use_effect_with_deps, use_node_ref, virtual_dom::AttrValue,
+    Component, Context, Html, Properties,
+};
+
+#[derive(Properties, PartialEq)]
+struct AttributeInputProps {
+    name: AttrValue,
+    attribute_type: AttributeType,
+    #[prop_or(None)]
+    value: Option<String>,
+}
+
+#[function_component(AttributeInput)]
+fn attribute_input(props: &AttributeInputProps) -> Html {
+    let input_type = match props.attribute_type {
+        AttributeType::String => "text",
+        AttributeType::Integer => "number",
+        AttributeType::DateTime => {
+            return html! {
+                <DateTimeInput name={props.name.clone()} value={props.value.clone()} />
+            }
+        }
+        AttributeType::Jpeg => {
+            return html! {
+                <JpegFileInput name={props.name.clone()} value={props.value.clone()} />
+            }
+        }
+    };
+
+    html! {
+        <input
+            type={input_type}
+            name={props.name.clone()}
+            class="form-control"
+            value={props.value.clone()} />
+    }
+}
+
+#[derive(Properties, PartialEq)]
+struct AttributeLabelProps {
+    pub name: String,
+}
+#[function_component(AttributeLabel)]
+fn attribute_label(props: &AttributeLabelProps) -> Html {
+    let tooltip_ref = use_node_ref();
+
+    use_effect_with_deps(
+        move |tooltip_ref| {
+            Tooltip::new(
+                tooltip_ref
+                    .cast::<Element>()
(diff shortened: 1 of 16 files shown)


nitnelave:
@bojidar-bg I pulled your PR, updated it and did a good once-over, it should be ready to merge now (as soon as I fix the warning/test), if you want to have a look.

app/src/infra/tooltip.rs
```diff
@@ -0,0 +1,15 @@
...
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = bootstrap)]
+    pub type Tooltip;
+
+    #[wasm_bindgen(constructor, js_namespace = bootstrap)]
+    pub fn new(e: web_sys::Element) -> Tooltip;
+
+    #[wasm_bindgen(method, js_namespace = bootstrap)]
+    pub fn toggle(this: &Tooltip);
```
bojidar-bg, line 14: We don't use this one anymore actually. (It was only used in some code I commented out, IIRC.)

app/src/components/form/attribute_input.rs
```diff
@@ -0,0 +1,193 @@
...
+            move |tooltip_ref| {
+                let tooltip_ref = tooltip_ref
+                    .cast::<Element>()
+                    .expect("Tooltip element should exist");
+                *tooltip_js.borrow_mut() = Some(Tooltip::new(tooltip_ref));
+                || {}
+            },
+            tooltip_ref.clone(),
+        );
+    }
```
bojidar-bg, line 66: We actually don't need `tooltip_js` (value is never used; was only used in aforementioned commented-out code).

So you can probably simplify this to:
```suggestion
    use_effect_with_deps(
        move |tooltip_ref| {
            let tooltip_ref = tooltip_ref
                .cast::<Element>()
                .expect("Tooltip element should exist");
            Tooltip::new(tooltip_ref);
            || {}
        },
        tooltip_ref.clone(),
    );
```

Yay less magic! (And do note that using `use_node_ref` with `use_effect_with_deps` is suggested by [Yew's documentation](https://yew.rs/docs/0.19.0/concepts/function-components/pre-defined-hooks#use_node_ref))

bojidar-bg:
It's looking good! And much cleaner than what I had :sparkles: Thanks for taking care of those last few cleanup steps, I was dreading getting around to them... :sweat_smile:  ~~Pretty sure there are plenty of good Rust practices for me to pick up if I just read your changes.. :grin:~~

Only thing I would suggest is, if you want to credit all the code authors, yet still retain the singular commit, you could maybe use the [`Co-authored-by:`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line) syntax? (:

server/src/infra/graphql/query.rs
```diff
@@ -661,18 +665,136 @@ pub fn serialize_attribute(
...
-                    _phantom: std::marker::PhantomData,
-                },
-                _phantom: std::marker::PhantomData,
-            }),
-            None => Err(FieldError::from(format!("Unknown attribute {}", &a.name))),
-        }
+    fn from_schema(a: DomainAttributeValue, schema: &DomainAttributeList) -> Option<Self> {
+        schema
+            .get_attribute_schema(&a.name)
+            .map(|s| AttributeValue::<Handler>::from_domain(a, s.clone()))
```
bojidar-bg, line 671: I'm not 100% convinced by the way we are filtering out `!is_visible` fields here, TBH. Wouldn't it be cleaner to do it in `UserRestrictedListerBackendHandler::list_users`, then use that for `Query<Handler>::users` instead of `ReadonlyBackendHandler::list_users`? (:

app/src/components/form/attribute_input.rs
nitnelave: Thanks!

server/src/infra/graphql/query.rs
nitnelave: You can't make sense of a (serialized) attribute without the schema, and the schema is restricted to only visible items in `UserRestrictedListerBackendHandler` (access_control.rs:324)

nitnelave:
Thanks for the `Co-authored-by:`, I was looking for something similar, but it's the first time I'm doing it :)

PR #942, opened 2024-07-30
Completing the work on the user_attributes branch

bojidar-bg:
Not sure what's left to fix; some fixes are probably best cherry-picked into separate PRs.

### DIFF
--- a/app/src/components/create_user.rs
+++ b/app/src/components/create_user.rs
@@ -1,21 +1,36 @@
 use crate::{
     components::{
-        form::{field::Field, submit::Submit},
+        form::{attribute_input::{ListAttributeInput, SingleAttributeInput}, field::Field, submit::Submit},
         router::AppRoute,
-    },
-    infra::{
+    }, convert_attribute_type, infra::{
         api::HostService,
-        common_component::{CommonComponent, CommonComponentParts},
-    },
+        common_component::{CommonComponent, CommonComponentParts}, schema::AttributeType,
+    }
 };
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use gloo_console::log;
 use graphql_client::GraphQLQuery;
 use lldap_auth::{opaque, registration};
 use validator_derive::Validate;
+use web_sys::{FormData, HtmlFormElement};
 use yew::prelude::*;
 use yew_form_derive::Model;
 use yew_router::{prelude::History, scope_ext::RouterScopeExt};
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "../schema.graphql",
+    query_path = "queries/get_user_attributes_schema.graphql",
+    response_derives = "Debug,Clone,PartialEq,Eq",
+    custom_scalars_module = "crate::infra::graphql"
+)]
+pub struct GetUserAttributesSchema;
+
+use get_user_attributes_schema::ResponseData;
+
+pub type Attribute = get_user_attributes_schema::GetUserAttributesSchemaSchemaUserSchemaAttributes;
+
+convert_attribute_type!(get_user_attributes_schema::AttributeType);
 
 #[derive(GraphQLQuery)]
 #[graphql(
@@ -29,17 +44,16 @@ pub struct CreateUser;
 pub struct CreateUserForm {
     common: CommonComponentParts<Self>,
     form: yew_form::Form<CreateUserModel>,
+    attributes_schema: Option<Vec<Attribute>>,
+    form_ref: NodeRef,
 }
 
 #[derive(Model, Validate, PartialEq, Eq, Clone, Default)]
 pub struct CreateUserModel {
     #[validate(length(min = 1, message = "Username is required"))]
     username: String,
-    #[validate(email(message = "A valid email is required"))]
-    email: String,
-    display_name: String,
-    first_name: String,
-    last_name: String,
+    // #[validate(email(message = "A valid email is required"))]
+    // email: String,
     #[validate(custom(
         function = "empty_or_long",
         message = "Password should be longer than 8 characters (or left empty)"
@@ -59,6 +73,7 @@ fn empty_or_long(value: &str) -> Result<(), validator::ValidationError> {
 
 pub enum Msg {
     Update,
+    ListAttributesResponse(Result<ResponseData>),
     SubmitForm,
     CreateUserResponse(Result<create_user::ResponseData>),
     SuccessfulCreation,
@@ -79,21 +94,48 @@ impl CommonComponent<CreateUserForm> for CreateUserForm {
     ) -> Result<bool> {
         match msg {
             Msg::Update => Ok(true),
+            Msg::ListAttributesResponse(schema) => {
+                self.attributes_schema = Some(schema?.schema.user_schema.attributes.into_iter().collect());
+                Ok(true)
+            }
             Msg::SubmitForm => {
                 if !self.form.validate() {
                     bail!("Check the form for errors");
                 }
+
+                let form = self.form_ref.cast::<HtmlFormElement>().unwrap();
+                let form_data = FormData::new_with_form(&form)
+                    .map_err(|e| anyhow!("Failed to get FormData: {:#?}", e.as_string()))?;
+                let all_values = get_values_from_form_data(
+                    self.attributes_schema
+                        .iter()
+                        .flatten()
+                        .filter(|attr| !attr.is_hardcoded || attr.is_editable)
+                        .collect(),
+                    &form_data,
+                )?;
+                let attributes =  if all_values.is_empty() {
+                    None
+                } else {
+                    Some(
+                        all_values
+                            .into_iter()
+                            .filter(|(_, value)| !value.is_empty())
+                            .map(|(name, value)| create_user::AttributeValueInput { name, value })
+                            .collect(),
+                    )
+                };
+
                 let model = self.form.model();
-                let to_option = |s: String| if s.is_empty() { None } else { Some(s) };
                 let req = create_user::Variables {
                     user: create_user::CreateUserInput {
                         id: model.username,
-                        email: model.email,
-                        displayName: to_option(model.display_name),
-                        firstName: to_option(model.first_name),
-                        lastName: to_option(model.last_name),
+                        email: None,
+                        displayName: None,
+                        firstName: None,
+                        lastName: None,
                         avatar: None,
-                        attributes: None,
+                        attributes,
                     },
                 };
                 self.common.call_graphql::<CreateUser, _>(
@@ -177,11 +219,20 @@ impl Component for CreateUserForm {
     type Message = Msg;
     type Properties = ();
 
-    fn create(_: &Context<Self>) -> Self {
-        Self {
+    fn create(ctx: &Context<Self>) -> Self {
+        let mut component = Self {
             common: CommonComponentParts::<Self>::create(),
             form: yew_form::Form::<CreateUserModel>::new(CreateUserModel::default()),
-        }
+            attributes_schema: None,
+            form_ref: NodeRef::default(),
+        };
+        component.common.call_graphql::<GetUserAttributesSchema, _>(
+            ctx,
+            get_user_attributes_schema::Variables {},
+            Msg::ListAttributesResponse,
+            "Error trying to fetch user schema",
+        );
+        component
     }
 
     fn update(&mut self, ctx: &Context<Self>, msg: Self::Message) -> bool {
@@ -192,38 +243,15 @@ impl Component for CreateUserForm {
         let link = &ctx.link();
         html! {
           <div class="row justify-content-center">
-            <form class="form py-3" style="max-width: 636px">
+            <form class="form py-3" //style="max-width: 636px"
+              ref={self.form_ref.clone()}>
               <Field<CreateUserModel>
                 form={&self.form}
                 required=true
                 label="User name"
                 field_name="username"
                 oninput={link.callback(|_| Msg::Update)} />
-              <Field<CreateUserModel>
-                form={&self.form}
-                required=true
-                label="Email"
-                field_name="email"
-                input_type="email"
-                oninput={link.callback(|_| Msg::Update)} />
-              <Field<CreateUserModel>
-                form={&self.form}
-                label="Display name"
-                field_name="display_name"
-                autocomplete="name"
-                oninput={link.callback(|_| Msg::Update)} />
-              <Field<CreateUserModel>
-                form={&self.form}
-                label="First name"
-                field_name="first_name"
-                autocomplete="given-name"
-                oninput={link.callback(|_| Msg::Update)} />
-              <Field<CreateUserModel>
-                form={&self.form}
-                label="Last name"
-                field_name="last_name"
-                autocomplete="family-name"
-                oninput={link.callback(|_| Msg::Update)} />
+              {self.attributes_schema.iter().flatten().filter(|a| !a.is_hardcoded || a.is_editable).map(|a| get_custom_attribute_input(a)).collect::<Vec<_>>()}
               <Field<CreateUserModel>
                 form={&self.form}
                 label="Password"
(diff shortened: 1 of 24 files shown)


app/src/components/form/attribute_input.rs
```diff
@@ -56,8 +56,9 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
     html! {
         <div class="row mb-3">
             <label for={props.name.clone()}
-                class="form-label col-4 col-form-label">
-                {&props.name}{":"}
+                class="form-label col-4 col-form-label"
+                title={props.name.clone()}>
+                {props.name[0..1].to_uppercase() + &props.name[1..].replace("_", " ")}{":"}
```
nitnelave, line 61: You don't have to do it in this PR, but I'd like to have a little (i) tooltip with the technical name of the field, as well as any LDAP aliases (not currently part of the API)

app/src/components/form/attribute_input.rs
```diff
@@ -68,3 +69,59 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
...
+    pub name: String,
+    pub attribute_type: AttributeType,
+    pub values: Vec<String>,
+}
+
+#[function_component(ListAttributeInput)]
+pub fn list_attribute_input(props: &ListAttributeInputProps) -> Html {
+
+    // let value_indices = use_memo(props.values, |vals| use_state(|| (0..(vals.len() as isize)).collect::<Vec<_>>()));
+    let value_indices = use_state(|| (0..(props.values.len() as isize)).collect::<Vec<_>>() ); // use_memo(props.values)
```
nitnelave, line 84: I really have to learn to use `use_state` and company...

app/src/components/form/attribute_input.rs
```diff
@@ -68,3 +69,59 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
...
+    let new_index = use_mut_ref::<isize>(|| 0);
+
+    html! {
+        <div class="row mb-3">
+            <label for={props.name.clone()}
+                class="form-label col-4 col-form-label"
+                title={props.name.clone()}>
+                {props.name[0..1].to_uppercase() + &props.name[1..].replace("_", " ")}{":"}
```
nitnelave, line 92: Extract the name formatting to a function?

app/src/components/form/attribute_input.rs
```diff
@@ -68,3 +69,59 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
...
+            </label>
+            <div class="col-8">
+            {value_indices.iter().map(|i| {let i = *i; html! {
```
nitnelave, line 95: I *think* you can do `map(|&i| {` to avoid the `let i = *i;`

nitnelave: or `.iter().cloned()` but given that it's Copy, the version above should work?

app/src/components/form/attribute_input.rs
```diff
@@ -68,3 +69,59 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
...
+                <div class="input-group mb-2" key={i}>
+                <AttributeInput
+                    attribute_type={props.attribute_type.clone()}
+                    name={props.name.clone()}
+                    value={usize::try_from(i).ok().and_then(|i| props.values.get(i)).cloned().unwrap_or_else(|| String::from(""))} />
```
nitnelave, line 100: `unwrap_or_default`

nitnelave: Also, it's reasonable to `.expect()` here, since the `i` are coming from `usize` to start with.

nitnelave: So, just `i as usize` should work?

app/src/components/form/attribute_input.rs
```diff
@@ -68,3 +69,59 @@ pub fn single_attribute_input(props: &SingleAttributeInputProps) -> Html {
...
+                </div>
+            }}).collect::<Html>()}
+            <button
+                class="btn btn-secondary"
+                type="button"
+                onclick={
+                    let value_indices = value_indices.clone();
+                    move |_| {
+                        *new_index.borrow_mut() -= 1;
+                        value_indices.set((*value_indices).clone().into_iter().chain(std::iter::once(*new_index.borrow())).collect())
```
nitnelave, line 119: I think `.iter().cloned()` should be better: it'll clone the elements, not the container.

nitnelave: Also, I won't pretend I understand this code, since it's almost 11PM.

app/src/components/user_details_form.rs
```diff
@@ -413,6 +418,8 @@ impl UserDetailsForm {
         if let Some(avatar) = &self.avatar {
             user_input.avatar = Some(to_base64(avatar)?);
         }
+        user_input.removeAttributes = remove_attributes;
```
nitnelave, line 421: That might very well be what I was hitting last time I was trying to fix this code. You're the 3rd author of this chain of WIP PRs, btw :)

app/src/components/form/attribute_input.rs
nitnelave, line 1: I think we'll need a special attribute input for avatars: we need to handle uploading the file, and reacting to the end of the upload, as well as complaining that it's not the right file format and so on. The custom code present in master, basically.

nitnelave (commented):
Looks promising! Thanks a lot :)

app/src/components/form/attribute_input.rs
bojidar-bg: Something like this, already enabled by the title= on the line above? :grin: 
image

app/src/components/form/attribute_input.rs
bojidar-bg: Oh, duh! I didn't even realize I had duplicated that code. :sweat_smile:

app/src/components/form/attribute_input.rs
nitnelave: That's already great, but I was thinking a more visible "ⓘ", potentially with a dismissable modal? To be mobile-friendly

app/src/components/form/attribute_input.rs
bojidar-bg: `i` can be negative, and I don't wish to imagine the .expect() errors in case someone changes the underlying props.values list. Until we have it with a memo, or_default is better, and even after that we need to change for positive values.
...as I said, I'm not too happy with how value_indices turned out :sweat_smile:

app/src/components/form/attribute_input.rs
bojidar-bg: Ooh... I see. That makes the display commit worthy of its very own draft PR :joy:

nitnelave:
The list implementation looks much better now!

app/src/components/form/attribute_input.rs
bojidar-bg: Done. (or done-ish, who knows XD)

app/src/components/form/attribute_input.rs
bojidar-bg: Turns out, I'm a bit silly and didn't realize I can just use indices past the array length for the same result.

app/src/components/form/attribute_input.rs
bojidar-bg: It could still use a function, but at least it's not duplicated now (:

bojidar-bg:
So... extra pain point from the avatar implementation, I needed to throw in an extra input type="hidden" to get values back to FormData, duplicating the long base64 string used for the image in the process.

nitnelave:
A few comments for having tried it out:
- It's listing the "display_name" and "mail" attributes as fields (in addition to "Email" and "Display Name"), but when you try to fill them out, they get rejected on submission for not being part of the schema
  - If it helps, I don't mind having a hard-coded exception list to exclude them.
- It doesn't show the non-visible fields, even to admins. Same for non-editable.
- List works great :)
- Avatars seem to work well, but I guess I didn't upload anything chunky enough to see it loading. I assume it works.
- The date picker should either not be the default one, or we have to change the backend to be more permissive when parsing a date. Using firefox on MacOS, the date picker doesn't let me pick a second, only down to minutes, and it doesn't parse on the backend. I also can't modify it by hand (looking at the value in the inspector, it was "2024-08-13T01:03").
- Hoover/info is nice!
- User creation doesn't have the custom attributes (but that can be a follow-up PR)
- I don't think we need to care about the extra copy of the avatar, it's a local copy. If it becomes a problem we'll dig into it, but let's not spend time on it for now.
- There's a warning or 2 in the code, but it's a draft PR so no worries. Let's get them ironed out by the time it's a proper PR though.

That's all for now! :) Thanks a lot, that already looks much better than what I had :D

bojidar-bg:
* The "display_name" and "mail" attributes are now properly handled. I was initially going to do it in the graphql code, but since the schema.rs that lists them is in `domain/` , I figured it might be only proper to do it in the backend code instead. Either way, it's not too hard to move it around as needed.
* Visible but not editable fields are displayed as static values; fields editable by the admin are displayed as fields.
    * The way we assume that of all `is_hardcoded` fields, only those that are marked `is_editable` are editable by the admin is a bit ugly; as there might potentially be admin-editable user-uneditable hardcoded fields one day.
* Avatars do not yet block the form; it's probably not too hard to do that by writing a dummy non-base64 value in the hidden input and watching it fail serialization... or well, passing down an extra callback as a prop which lets us set individual fields as "blocking form".
* The date picker has been fixed, and works at least under firefox. I'm not looking forward to implementing a custom datepicker, so if it doesn't work on any platform, someone might have to fix it in a follow-up PR.
* User creation has the custom attributes now! There is a bit of code duplication, not sure what the ideal way to extract it would be.
* Warnings were corrected.. maybe. Build is failing because I opted to drop the "merge main" commit for now.

app/src/components/form/file_input.rs
```diff
@@ -0,0 +1,250 @@
...
+            //     }
+            //   } else { html! {} }
+            // }
+        }
+    }
+}
+
+impl JpegFileInput {
+    fn upload_files(files: Option<FileList>) -> Msg {
+        if let Some(files) = files {
```
nitnelave, line 234: maybe with a match and a guard condition?
```
Some(files) if !files.is_empty() => FileSelected,
_ => Update
```

app/src/components/user_details_form.rs
```diff
@@ -220,63 +125,8 @@ impl Component for UserDetailsForm {
...
-                            alt="Avatar" />
-                        }
-                      } else { html! {} }
-                    }
-                    </div>
-                  </div>
-                </div>
-              </div>
-              {ctx.props().user_attributes_schema.iter().filter(|a| a.is_editable).map(|s| get_custom_attribute_input(s, &self.user.attributes)).collect::<Vec<_>>()}
+              {ctx.props().user_attributes_schema.iter().filter(|a| (ctx.props().is_admin && !a.is_hardcoded) || a.is_editable).map(|s| get_custom_attribute_input(s, &self.user.attributes)).collect::<Vec<_>>()}
```
nitnelave, line 128: Why the `!is_hardcoded` ? Which hardcoded attribute should not be editable, but is part of the schema?

We might have to either remove them or add a new field like `is_static_data`.

app/src/components/user_details_form.rs
```diff
@@ -341,39 +196,59 @@ fn get_custom_attribute_input(
...
+        // }
         let form = self.form_ref.cast::<HtmlFormElement>().unwrap();
         let form_data = FormData::new_with_form(&form)
             .map_err(|e| anyhow!("Failed to get FormData: {:#?}", e.as_string()))?;
         let mut all_values = get_values_from_form_data(
             ctx.props()
                 .user_attributes_schema
                 .iter()
-                .filter(|attr| attr.is_editable)
+                .filter(|attr| (ctx.props().is_admin && !attr.is_hardcoded) || attr.is_editable)
```
nitnelave, line 234: We'll need to extract that check to a function.

server/src/domain/handler.rs
```diff
@@ -101,7 +101,7 @@ impl From<bool> for GroupRequestFilter {
 pub struct CreateUserRequest {
     // Same fields as User, but no creation_date, and with password.
     pub user_id: UserId,
-    pub email: Email,
+    pub email: Option<Email>, // Required, but could be passed in attributes
```
nitnelave, line 104: Hard disagree. At this layer, the email should have been taken out of the attributes and into the field. We should have a single, canonical representation in the DB, and it's with email as a column of user, not in the attributes.

I think the correct place to do the translation is in the GrahpQL layer, in mutation.rs

nitnelave: Same for `display_name`, pick one, either a column or an attribute, but not both.

server/src/infra/graphql/query.rs
```diff
@@ -661,18 +662,18 @@ pub fn serialize_attribute(
...
                 schema: AttributeSchema::<Handler> {
                     schema: s.clone(),
                     _phantom: std::marker::PhantomData,
                 },
                 _phantom: std::marker::PhantomData,
-            }),
-            None => Err(FieldError::from(format!("Unknown attribute {}", &a.name))),
-        }
+            })
+        // None => Err(FieldError::from(format!("Unknown attribute {}", &a.name))),
```
nitnelave, line 676: You want to just swallow errors? That seems risky.

app/src/components/user_details_form.rs
nitnelave: Oh, the `creation_date` is in the schema... Yeah, not great, we need a `is_static` or something. Maybe not static since we'll have last_modification_time, then `is_readonly`?

nitnelave (changes requested):
That looks great!

The functionality is there, there's just some code questions to polish.

Thanks a lot for your work!

app/src/components/create_user.rs
```diff
@@ -79,21 +101,49 @@ impl CommonComponent<CreateUserForm> for CreateUserForm {
...
                 }
+
+                let form = self.form_ref.cast::<HtmlFormElement>().unwrap();
+                let form_data = FormData::new_with_form(&form)
+                    .map_err(|e| anyhow!("Failed to get FormData: {:#?}", e.as_string()))?;
+                let all_values = get_values_from_form_data(
+                    self.attributes_schema
+                        .iter()
+                        .flatten()
+                        .filter(|attr| !attr.is_hardcoded || attr.is_editable)
```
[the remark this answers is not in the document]
nitnelave: Added an `is_readonly`, I don't know if that helps

app/src/components/user_details_form.rs
nitnelave: added `is_readonly`

server/src/domain/handler.rs
nitnelave: I'll have a look if I have time to change the server side to handle that for you, outside of this PR

nitnelave: Found time to do it. You can now send the email and display name as attributes _or_ main field.

app/src/components/create_user.rs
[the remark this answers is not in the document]
bojidar-bg: That'll help a ton! Thanks! :tada:

app/src/components/form/attribute_input.rs
```diff
@@ -26,23 +32,68 @@ fn attribute_input(props: &AttributeInputProps) -> Html {
...
+    pub name: String,
+}
+#[function_component(AttributeLabel)]
+fn attribute_label(props: &AttributeLabelProps) -> Html {
+    let tooltip_ref = use_node_ref();
+    let tooltip_js = use_mut_ref(|| None);
+
+    {
+        let tooltip_js = tooltip_js.clone();
+        use_effect_with_deps(
```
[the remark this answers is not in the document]
bojidar-bg: If we update to newer yew (>=0.20) with `use_memo`, I should be able to avoid some of that magic. :grin:

app/src/components/form/date_input.rs
```diff
@@ -0,0 +1,41 @@
...
+        props
+            .value
+            .as_ref()
+            .and_then(|x| DateTime::<Utc>::from_str(x).ok())
+    });
+
+    html! {
+        <div class="input-group">
+            <input
+                type="hidden"
```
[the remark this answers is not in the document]
bojidar-bg: Same as the file field, actually.
Perhaps it could be worth to make all fields just use regular prop+onchange to pass values to the main component, instead of collecting them by `name` from the form...

server/src/infra/graphql/query.rs
bojidar-bg: Fields that are not `is_visible` are not in the schema if we are not an admin. Not sure where that was stemming from, but IIRC it comes from the method that returns and filters the schema based on the authenticated user earlier, so it felt better to just ignore those fields here than to change the backend that deals with permissions and move the filtering to the graphql frontend.

...now that I'm writing that out, I'm thinking that maybe I can add some filtering to the part of the backend that lists the attributes of the user and leave this unchanged... :thinking:

app/src/components/form/attribute_input.rs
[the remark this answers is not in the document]
nitnelave: Yeah, I looked into updating yew... That'll be a major piece of work, a huge PR with no in-between steps possible.

server/src/infra/graphql/query.rs
nitnelave: Wait, we return the _values_ for fields that are not in the schema because they're not visible?

Definitely broken... Can you file that as a bug?

nitnelave: oh, we're still inside the server, it's graphql's query.rs that's supposed to do the filtering, WAI

bojidar-bg:
Superseded by/merged into #940

PR #944, opened 2024-08-06
Adds example configuration for Netbox

shrunbr:
I noticed that there wasn't an example config for Netbox. It took me about 20 minutes to get this working and I figured I would add it so others have an easier time. Any modifications are welcome!


nitnelave (commented):
That's a very nice guide!
I feel like it's not the first Django based config we have, do you think we should consolidate? Are there many things that are netbox-specific here?

nitnelave (approved):

Issue #945, opened 2024-08-07
[BUG] Rust 1.80 compatibility


bendlas:
**Describe the bug**

When compiling on rust 1.80, the client builds, but the server build errors in the `time` dependency. After updating dependencies with a `cargo update`, the server builds, but the client build has multiple errors in `app/src/components`.

**To Reproduce**
1. Use a rust 1.80 toolchain
2. `cargo build` -- server build error
3. `./app/build.sh` -- client build success
4. `cargo update`
5. `cargo build` -- server build success
6. `./app/build.sh` -- client build error

**Expected behavior**

Client build succeeds after `cargo update`

**Logs**

Error logs in gist: https://gist.github.com/bendlas/3630da2c9d9a399ac01e6ff33700faa3

PR #947, opened 2024-08-08
cargo: update time

alyssais:
Fixes building with Rust 1.80.0.

See https://github.com/rust-lang/rust/issues/127343.

Closes: https://github.com/lldap/lldap/issues/945

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,7 +233,7 @@ dependencies = [
  "serde_urlencoded",
  "smallvec",
  "socket2",
- "time 0.3.19",
+ "time 0.3.36",
  "url",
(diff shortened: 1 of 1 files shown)


nitnelave (approved):

bendlas:
Thanks! Might we also have a `0.5.1` tag by any chance?

nitnelave:
> Thanks! Might we also have a `v0.5.1` tag by any chance?

0.5.0 is actually 0.5.0-beta, it's not a proper release. It's blocked on getting the web UI finished (which has seen some activity since last week!)

You can grab the daily builds/tags as a "release"

PR #962, opened 2024-08-27
server: map email and display_name from attributes into user fields

nitnelave:


nitnelave:
@bojidar-bg that's the second piece of the puzzle. You shouldn't have to change the backend now.

server/src/infra/graphql/mutation.rs
```diff
@@ -120,6 +122,44 @@ impl Success {
...
+        .iter()
+        .find(|attr| attr.name == "mail")
+        .cloned()
+        .map(|attr| deserialize_attribute(&schema.get_schema().user_attributes, attr, is_admin))
+        .transpose()?
+        .map(|attr| attr.value.unwrap::<String>())
+        .map(Email::from);
+    let display_name = attributes
+        .iter()
+        .find(|attr| attr.name == "displayName")
```
bojidar-bg, line 146: Mm... excuse the question, but isn't the attribute called "display_name"? https://github.com/lldap/lldap/blob/6cf0f6df0639425f792af243b4345acc9b737fa8/server/src/domain/schema.rs#L75-L76

nitnelave: Yup, it is... Sending a fix.

PR #978, opened 2024-09-16
Update the documentation for PAM integration

rkj:
Remove `is-unix-user` mentions, add note about using nightly, fix typo for passwd gid entry.


nitnelave (approved):
Thanks!

PR #979, opened 2024-09-22
example_configs: Update Radicale guide with configuration for Radicale 3.2.4

poperigby:
3.2.4 hasn't actually released yet, so maybe wait for this to be merged.

### DIFF
--- a/example_configs/radicale.md
+++ b/example_configs/radicale.md
@@ -1,20 +1,14 @@
-# Configuration of RADICALE authentification with lldap.
+# Configuration of Radicale authentication with LLDAP.
 
-# Fork of the radicale LDAP plugin to work with LLDAP : https://github.com/shroomify-it/radicale-auth-ldap-plugin
+Requires Radicale 3.2.4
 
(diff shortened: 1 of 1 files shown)


nitnelave (changes requested):
You could maybe merge it as another paragraph: 
- config with plugin
- config with 2.3.4

That way if someone doesn't want to upgrade, they can still see the config for the plugin. And we can merge right now without waiting for the release.

example_configs/radicale.md
```diff
@@ -1,20 +1,14 @@
...
-
-# Radicale config file v0.3 (inside docker container /etc/radicale/config https://radicale.org/v3.html#configuration)
-
-```toml
+```ini
 [auth]
-type = radicale_auth_ldap
-ldap_url = ldap://lldap:3890  
+type = ldap
+ldap_uri = ldap://localhost:3890
```
nitnelave, line 8: Let's keep it as ldap://lldap:3890

poperigby:
@nitnelave Should be good now.

nitnelave (approved):
Thanks a lot!

poperigby:
No problem 🙂

PR #985, opened 2024-10-05
app: Clean up code, don't error on admin empty email

nitnelave:


nitnelave:
FYI @bojidar-bg

app/src/infra/form_utils.rs
```diff
@@ -0,0 +1,59 @@
...
+                attr.name
+            );
+            Ok(AttributeValue {
+                name: attr.name.clone(),
+                values: val,
+            })
+        })
+        .collect::<Result<Vec<_>>>()?;
+    let maybe_email_values = all_values.iter().find(|a| a.name == "mail");
+    if email_is_required || maybe_email_values.is_some() {
```
bojidar-bg, line 51: Having this in the function feels like breaking single-responsibility principle; maybe have it in an a separate function? (:

nitnelave: I like that you can't forget to validate. I broke out the validation in a separate function (called from here) though.

PR #991, opened 2024-10-09
Add support for bootstrapping schemas

ggodlewski:
* Moved default bootstrap dirs into single /bootstrap parent dir in order to have single docker volume bind
* Added default values for LDAP url and credentials - same as default lldap values
* Added support for bootstrapping schemas

Place schema files under /bootstrap/(user|group)-schemas/*.json

Sample content:
```
[
  {
    "name" : "test_attrib",
    "attributeType" : "STRING",
    "isEditable" : true,
    "isList" : false,
    "isVisible" : true
  }
]
```

### DIFF
--- a/example_configs/bootstrap/bootstrap.md
+++ b/example_configs/bootstrap/bootstrap.md
@@ -1,4 +1,4 @@
-# Bootstrapping lldap using [bootstrap.sh](bootstrap.sh) script
+# Bootstrapping lldap using [bootstrap.sh](/scripts/bootstrap.sh) script
 
 bootstrap.sh allows managing your lldap in a git-ops, declarative way using JSON config files.
 
@@ -12,7 +12,7 @@ The script can:
 * create groups
 * delete redundant users and groups (when `DO_CLEANUP` env var is true)
 * maintain the desired state described in JSON config files
-
+* create user/group user-defined attributes
 
 ![](bootstrap-example-log-1.jpeg)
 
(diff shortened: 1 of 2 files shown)


nitnelave (approved):

nitnelave:
I'm not a user of this file, but: is this change going to break for anyone? You mentioned something about changing the directory structure, if someone pulls the new version, is it going to break for them?

example_configs/bootstrap/bootstrap.md
```diff
@@ -129,10 +179,17 @@ Let's suppose you have the next file structure:
    │  ├─ ...
    │  └─ user-n.json
    └─ group-configs
-      ├─ group-1.json
+   |  ├─ group-1.json
+   |  ├─ ...
+   |  └─ group-n.json
+   └─ user-schames
```
nitnelave, line 185: What a schame... :)

example_configs/bootstrap/bootstrap.md
```diff
@@ -205,14 +264,15 @@ spec:
           volumeMounts:
             - name: bootstrap
```
nitnelave, line 265: Why not mount the whole bootstrap directory? That was the point of the change, no?

nitnelave (changes requested):
Much better! And thanks for updating the docs as well. Just a couple more nits and it's good to merg.

example_configs/bootstrap/bootstrap.md
ggodlewski: I was thinking about it, but I've never setup kubernetes. Below there are `volumes.projected.sources` . Someone who is able to test this config should have a look.

nitnelave (approved):
Thanks!

breakingflower:
hi

nice work!

I notice the do_cleanup is not cleaning up old user schema's or group schema's. 

Example: 
`user-schemas/test.json` -> 
```
[
  {
    "name": "tes321t",
    "attributeType": "STRING",
    "isEditable": false,
    "isList": false,
    "isVisible": true
  },
  {
    "name": "te21312312s321t",
    "attributeType": "STRING",
    "isEditable": false,
    "isList": false,
    "isVisible": true
  }
]
```
gets generated. 
if you remove the field `te21312312s321t` and rerun the script, the attribute is still visible. 
```
[
  {
    "name": "tes321t",
    "attributeType": "STRING",
    "isEditable": false,
    "isList": false,
    "isVisible": true
  }
]
```
The docs also show that this is intended. 

Second run: 
```
--- user schemas ---
User property "tes321t" already exists
--- user schemas ---
--- redundant groups ---
There are no redundant groups
--- redundant groups ---
--- redundant users ---
There are no redundant users
--- redundant users ---
``` 
It seems optimal to include it in the `bootstrap.sh` DO_CLEANUP method.

Issue #997, opened 2024-10-13
[BUG] Unable to search using multi-value attributes


ggodlewski:
**Describe the bug**

I created custom user attribute `mail_alias` with multiple values

```
{
  "name": "mail-alias",
  "attributeType": "STRING",
  "isEditable": false,
  "isList": true,
  "isVisible": true
}
```

Then I assigned multiple email addresses into the users.

```
info@example.com
admin@example.com
```

**To Reproduce**

I want to select an user containing one of the values:

```
ldapsearch -x -b "ou=people,dc=example,dc=com" -H ldap://localhost:3890 -D cn=admin,ou=people,dc=example,dc=com -W "mail-alias=info@example.com"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <ou=people,dc=example,dc=com> with scope subtree
# filter: mail-alias=info@example.com
# requesting: ALL
#

# search result
search: 2
result: 0 Success
control: 1.2.840.113556.1.4.319 false MAUCAQAEAA==
pagedresults: cookie=

# numResponses: 1
```

After I remove 1 of the values and list contains only 1 element it works. However the `mail-alias` value is not in the results.

```
ldapsearch -x -b "ou=people,dc=example,dc=com" -H ldap://localhost:3890 -D cn=admin,ou=people,dc=example,dc=com -W "mail-alias=info@example.com"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <ou=people,dc=example,dc=com> with scope subtree
# filter: mail-alias=info@example.com
# requesting: ALL
#

# test, people, example.com
dn: uid=test,ou=people,dc=example,dc=com
objectclass: inetOrgPerson
objectclass: posixAccount
objectclass: mailAccount
objectclass: person
uid: test
mail: test@example.com
createtimestamp: 2024-10-13T11:04:32.598982545+00:00
entryuuid: e5ed598e-57da-3c71-adbd-fe4261bdf4a4

# search result
search: 2
result: 0 Success
control: 1.2.840.113556.1.4.319 false MAUCAQEEAA==
pagedresults: estimate=1 cookie=

# numResponses: 2
# numEntries: 1
```

**Expected behavior**

I want to search for user containing email in `mail-alias` list.

nitnelave:
That's a known limitation of LLDAP.
The common workaround is to have `mail-alias1`, `mail-alias2` and so on as single value attributes

nitnelave:
[...]

PR #1000, opened 2024-10-13
Improve searching using attributes

ggodlewski:
* Add ability to search over attributes using wildcards

```
ldapsearch -x -b "ou=people,dc=example,dc=com" -H ldap://localhost:3890 -D cn=admin,ou=people,dc=example,dc=com -W "mailbox=*@example.com"
```

* Add ability to search using multi-value attributes: https://github.com/lldap/lldap/issues/997

```
ldapsearch -x -b "ou=people,dc=example,dc=com" -H ldap://localhost:3890 -D cn=admin,ou=people,dc=example,dc=com -W "mail-alias=info@example.com"
```

### DIFF
--- a/server/src/domain/sql_migrations.rs
+++ b/server/src/domain/sql_migrations.rs
@@ -68,6 +68,15 @@ pub enum UserAttributes {
     UserAttributeValue,
 }
 
+#[derive(DeriveIden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Copy)]
+pub enum UserAttributesSearch {
+    Table,
+    UserAttributeSearchUserId,
+    UserAttributeSearchName,
+    UserAttributeSearchNo,
+    UserAttributeSearchText,
+}
+
 #[allow(clippy::enum_variant_names)] // The table names are generated from the enum.
 #[derive(DeriveIden, PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum GroupAttributeSchema {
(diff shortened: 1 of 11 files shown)


server/src/domain/sql_migrations.rs
```diff
@@ -1131,6 +1140,91 @@ async fn migrate_to_v10(transaction: DatabaseTransaction) -> Result<DatabaseTran
...
+                        .columns([
+                            UserAttributeSchema::UserAttributeSchemaName,
+                            UserAttributeSchema::UserAttributeSchemaType,
+                            UserAttributeSchema::UserAttributeSchemaIsList,
+                            UserAttributeSchema::UserAttributeSchemaIsUserVisible,
+                            UserAttributeSchema::UserAttributeSchemaIsUserEditable,
+                            UserAttributeSchema::UserAttributeSchemaIsHardcoded,
+                        ])
+                        .values_panic([
+                            "mail".into(),
```
nitnelave, line 1214: Why are you adding the "mail" attribute there?

So far, I didn't convert it to a "custom" attribute because of the special requirements (it's used for password recovery, it needs to be non-empty, and up to now custom attributes were not searchable by substring, though you seem to have solved that one)

nitnelave (changes requested):
Before I go deeper into the details, I'd like to say something about the process:
- This would have been better discussed in advance, so we could discuss about the design and I could have given you a green light on principle, before you invested time into a substantial PR
- Similarly, the PR description is... terse, for something that heavily touches a core functionality, adding a new table and a non-trivial migration, changing the search process and hooking into the user creation utilities

I'll document here what I understood of the design from the PR, but that would have been better provided by you in the first place.

- There is a new table, UserAttributesSearch, that duplicates the values of the (text) attributes in the UserAttributes table, each accompanied by a number
- I could see no purpose for the number, aside from being unique for a given attribute+user and thus giving a primary key (do we need a primary key? why not just an index?)
- The user creation/update logic is hooked to populate and update that table.
- The search logic (now with an extra possibility for searching for substrings in attributes) is updated to search in the new table for matches.

I have some concerns about this:
- First, there is the duplication of information between the 2 tables. If we're storing the text attributes separately, there's no need to have them be in the same table at all: wouldn't it make more sense to remove them from the "general" attribute table and only store them in that "search" table (which would be renamed to UserTextAttributes, for instance)?
- Second, there is the genericity/extensibility: we have here a special case for text, arguably a reasonable one given the substring filter. However, it seems a natural extension to be able to compare numbers and dates with <, or to search for one (for instance, unix group membership could be encoded as an integer list attribute). The proposed design makes no affordances for this, making their addition later uncomfortable.

Lastly, there are some implementation details, but ones that can be resolved by iterating on the code. I'll list a few while I have them in mind, but there might be more:
- The sql migration doesn't migrate the existing attribute values, so only new values would become searchable
- The code relies on `Serialized` knowing whether it can deserialize into a `String` or `Vec<String>`. I don't think it actually has type information embedded, we need to couple it with the schema to know how to decode it.
- The attribute search doesn't (at first glance) handle gracefully non-String attributes, calling `vec.first().unwrap()` on a potentially empty vector
- I would use a newtype for wrapping `Vec<String>` for the conversion, since it might have an actual meaning (the `convert_to` method could actually be a trait implementation of `TryInto`)
- There are not enough tests. As a reviewer, I have no confidence that it works reliably except in the happiest of paths :)

I don't want to discourage you, since there are precious few contributors actually submitting Rust PRs, especially to implement features that I said the design didn't allow for 😅 . I like the practicality of the approach, and the fact that it _does_ solve the immediate problem. However, I'm the maintainer and I'll be the one keeping the code in good shape, so I'd like to only merge a design I'm comfortable with.

Thanks for the PR, and I do hope we can get it working! I think the next step is to have a good design conversation, either live (with a call), on Discord, or asynchronously on GitHub (though that might take longer)

server/src/domain/sql_migrations.rs
ggodlewski: I was getting foreign keys violation error because of inserting mail into this table https://github.com/lldap/lldap/pull/1000/files/8e2d89f2330fd7b89b3a0af81b3b3d12f3bfb244#diff-218546186ebdece3cd3f337cf80711f352fb77467f6e97c0cce99032e01aab47R466

ggodlewski:
I've done more testing and another issue occurred so unfortunately I had to switch to OpenLDAP.
This PR is a bit hacky because I needed working solution done changing least things possible.
I didn't want to do much refactoring and change function signatures.
Also I'm not familiar with SeaORM a lots of other things. So that's what I was able to in the Sunday afternoon.

Attributes are duplicated because any more advanced search is not possible because of Serialized.
Personally I would do the conversion just before storing in the DB.

The primary key necessary because compiler told me: Entity must have a primary key column. See <https://github.com/SeaQL/sea-orm/issues/485> for details. 

I didn't even consider searching among other types than String. My use case are query_filters from: https://nrdmnn.net/resources/2-LDAP-managed-mail-server-with-Postfix-and-Dovecot-for-multiple-domains

Last's week installation of email server broke my mind. I need to recover a bit.
After that I can work on it a bit. However, I need help with the design.

nitnelave:
Sorry for the delay, I got caught up with things.

I had thought about the design earlier, and the problem is that pretty much all the solutions are fairly intense in conceptual load and code complexity:
- One possibility is to create a table for each attribute, dynamically. That means that the queries to read/write users will be a lot more complex, touching a variable number of tables and stitching the results back together.
- Another possibility is to create a table per attribute _type_, i.e. string vs integer vs ... Still stitching queries together, but that can be hardcoded and avoid creating tables dynamically.
- In both cases, we can have some special implementations per type:
  - Strings support substring filter
  - integer, date and string support comparison
  - Strings can have the "display" value and the "lowercased for search" value, indexed
- In both cases, we would get rid of the existing user/group attributes table.

This should not require touching any other files than the SQL migration files, and the SQL backend (user/group) files. However, I expect these changes to be substantial, and there's no good way to make smaller steps, it has to be a big bang change.

WDYT?

nitnelave:
Yeah, I looked a bit more for the performance/storage tradeoffs for SQLite and PostgreSQL, and it seems a table per type makes more sense.

PR #1001, opened 2024-10-14
Dockerfile: add jq/jo/curl, required by bootstrap.sh (fix #921)

### LINKED ISSUE #921: [BUG] bootstrap.sh in rootless image
**Describe the bug**

bootstrap.sh does not work in rootless image as apk must be run as root to install curl/jq/jo

**To Reproduce**

1. Run container with latest-alpine-rootless tag
2. Execute bootstrap.sh inside container
3. See error about executable not found

```
Command not found "curl"
Command not found "jq"
Command not found "jo"
```

**Expected behavior**

bootstrap.sh should work with rootless image.

**Logs**

N/A

**Additional context**

One solution is to bundle required packages in the image though this will increase the image size for people not using bootstrap.sh.

johanneskastl:
fix #921 

Not sure if there is a separate Dockerfile for the rootless variant?

### DIFF
--- a/.github/workflows/Dockerfile.ci.alpine-base
+++ b/.github/workflows/Dockerfile.ci.alpine-base
@@ -64,7 +64,7 @@ WORKDIR /app
 ENV UID=1000
 ENV GID=1000
 ENV USER=lldap
-RUN apk add --no-cache tini ca-certificates bash tzdata && \
+RUN apk add --no-cache tini ca-certificates bash tzdata jq curl jo && \
     addgroup -g $GID $USER && \
     adduser \
     --disabled-password \
--- a/.github/workflows/Dockerfile.ci.debian-base
+++ b/.github/workflows/Dockerfile.ci.debian-base
@@ -65,7 +65,7 @@ ENV UID=1000
 ENV GID=1000
 ENV USER=lldap
 RUN apt update && \
-    apt install -y --no-install-recommends tini openssl ca-certificates tzdata && \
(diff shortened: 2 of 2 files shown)


johanneskastl:
> That won't do what you think (though it's a good change): that's the "developer" docker image, for people who don't want to install rust on their machine. The actual dockerfile(s) used for building the release images are in .github/workflows/

Thanks for the quick feedback. I reworked the branch and made the change to `.github/workflows/Dockerfile.ci.alpine-rootless`, is that the right one? Or should it go to the alpine-base one?

johanneskastl:
> > That won't do what you think (though it's a good change): that's the "developer" docker image, for people who don't want to install rust on their machine. The actual dockerfile(s) used for building the release images are in .github/workflows/
> 
> Thanks for the quick feedback. I reworked the branch and made the change to `.github/workflows/Dockerfile.ci.alpine-rootless`, is that the right one? Or should it go to the alpine-base one?

I could not test building, as the `FROM` references an image that comes from localhost...

nitnelave:
The build will be tested by the CI, don't worry about that.

Can you add the packages to the alpine-base image? That way both rootless and non-rootless get it.
Maybe also in debian-base.

johanneskastl:
> Can you add the packages to the alpine-base image? That way both rootless and non-rootless get it. Maybe also in debian-base.

Done. The packages should be called the same in Debian, as far as I can see.

nitnelave (approved):

PR #1004, opened 2024-10-18
fix: typo for first name attribute

danielsreichenbach:
It should be `givenName` instead of `givenname`. Using the later one, will result in Keycloak bugging out during the sync process, and henceforth displaying an empty user list.

### DIFF
--- a/example_configs/keycloak.md
+++ b/example_configs/keycloak.md
@@ -69,4 +69,4 @@ Since Keycloak and LLDAP use different attributes for different parts of a user'
 
 Go back to "User Federation", edit your LDAP integration and click on the "Mappers" tab.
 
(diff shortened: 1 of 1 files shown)


nitnelave (approved):

danielsreichenbach:
> I'm surprised this matters: LDAP is a case insensitive protocol, by and large

Surprisingly yes, at least in recent Keycloak Quarkus releases.

I did experiment a bit more though, and now I am a little confused. It may be the actual problem is that LLDAP only reacts to `givenName` and should also react to `givenname`.

nitnelave:
I am fairly confident that LLDAP reacts to both. Maybe if you have an ancient version, but that's been handled for a long time

danielsreichenbach:
Running 0.5.0, seems like anything newer has no release or package. Even made an Ansible role for it (https://github.com/danielsreichenbach/ansible-role-lldap). If there is a newer release I must have missed that.

nitnelave:
Yeah, I'm a *bit* behind on releases :)
The "latest"/nightly is far more recent

PR #1011, opened 2024-10-28
update to vaultwarden_ldap_sync:2.0.2

Hobbabobba:
added ssl-setting


nitnelave (approved):
Maybe a comment like "Set to true if using LDAPS" ?

Hobbabobba:
I added a comment regarding LDAPS

nitnelave (approved):

Issue #783, opened 2024-01-03
Password Expiration Policy


mbay-ODW:
**Motivation**
Currently there is no implementation of setting the non editable password expired attribute. However in using lldap it might be useful to define a certain password expiration policy such that passwords need to be rotated. This makes it challenging to enforce password rotation policies for users. This feature is essential for enhancing security and ensuring that users regularly update their passwords.

**Describe the solution you'd like**
OpenLdap uses the attribute: pwdMaxAge and pwdChangedTime that indicates when a password was last changed and need to be changed again based on the configured policy.

1. Populate the attributes that stores the expiration date of the user's password. This attribute should be set when the user changes their password or when it's initially set by the administrator.
2. Provide a mechanism to configure password expiration policies. 
3. (Optional) Implement a mechanism to notify users when their password is about to expire. 

**Describe alternatives you've considered**
Currently, administrators may have to rely on external scripts or tools to manage password expiration policies that would basically change the password thus that users have to reset them.

**Additional context**
Implementing password expiration in LDAP using the "lldap" library would greatly enhance security and ease the management of user accounts by providing a native and integrated solution. It would benefit organizations that rely on LDAP for authentication and user management, ensuring that password policies are enforced consistently and efficiently. I´m using it together with Authelia which reads the attributes of

Firstyear:
Password expiration is a security anti-pattern. It has been known for more than a decade to reduce password quality and cause more security issues than it resolves.

My advice to @nitnelave would be to never implement this "antifeature".

mbay-ODW:
As much as I agree with @Firstyear, some data privacy officers heavily jump on that topic.

nomandera:
Whilst I strongly agree with @Firstyear I should point out artificial password expiration unfortunately remains a common requirements of almost all accreditation and cyber security insurance polices.

nitnelave:
The main target of the LLDAP project is self-hosted servers, by individuals or small organizations who don't need accreditation or privacy officers. Regarding privacy/security, though I can vouch for the design of LLDAP, it hasn't been audited for use in a sensitive environment, so I'm not sure it'd be the darling of all privacy officers :)

[...]

Issue #514, opened 2023-03-29
Pass configuration to frontend


pixelrazor:
Currently, the frontend has no knowledge of the config. This leads to awkward things such as probing the backend to determine if password reset is enabled. Frontend configuration would also help solve #400, #247, and maybe #352

A proposed idea is to use the main.js file as a template, fill it in with a static json config object, and use that in the frontend (either passing it in to the main function, or using wasm_bindgen). [A basic template](https://docs.rs/text-template/latest/text_template/) should be good

nitnelave:
That or query a /settings endpoint returning a json. It might be simpler.

PR #1024, opened 2024-11-01
Adding regex to support basic complexity requirements.

gplubeck:
There appear to be a number of requests for volunteer organizations or small businesses ([521](https://github.com/lldap/lldap/issues/521), [783](https://github.com/lldap/lldap/issues/783), etc) that want basic password complexity.  This pull request would add the ability to to add pseudo complex password.  I know length is better indicator for initial password and entropy for password changes, but government organizations have their rules.

I am also tracking this doesn't enforce it on the back-end, nor with creation of an account.  This was purposeful as being an admin and creating an account with password or something easy to remember can be useful at times.  Additionally if you are changing the password via the cli, you have enough access to do whatever you want and should be responsible for your actions.

What this pull request is missing:
1) Not sure how lldap wants to support configuration.  Right now this merely replaces the empty_or_long validation.  Obviously this is not a good approach as it would be better implemented as a turn on when needed configuration.  Open to suggestions for how lldap usually does this.

2) Might be useful to add a database field for when passwords are changed.  This could be used to ensure password have been changed after a breach.  Additionally, for people like issue 783 they could update their clients to request the attribute if they really want. I am open to implementing this, but don't ever rust so would appreciate a push in the right direction.


app/src/components/change_password.rs
```diff
@@ -53,6 +60,20 @@ fn empty_or_long(value: &str) -> Result<(), validator::ValidationError> {
...
+    let has_uppercase = Regex::new(r"[A-Z]").unwrap().is_match(value);
+    let has_lowercase = Regex::new(r"[a-z]").unwrap().is_match(value);
+    let has_symbol = Regex::new(r"[^A-Za-z0-9]").unwrap().is_match(value);
+    let has_number = Regex::new(r"\d").unwrap().is_match(value);
+    let has_length = value.is_empty() || value.len() >= 8;
+
+    if has_uppercase && has_lowercase && has_symbol && has_number && has_length {
+        Ok(())
+    } else {
+        Err(validator::ValidationError::new(""))
```
nitnelave, line 73: Can we surface which constraint was not fulfilled? Maybe plural.

nitnelave (changes requested):
The change on its own looks not bad. However, as you noted, it should be togglable through configuration. There's no _easy_ way to pass config to the frontend for now, and it _has_ been a problem: for instance, to detect if password resets are enabled, we do an awkward query for an unusual login name and check the return value, which has broken before.

In short, I would like to address the configuration problem first (https://github.com/lldap/lldap/issues/514), before merging this.
There are some ideas in the ticket, I think the simplest is to query a config in json. For now we can copy paste the config struct in client and server code, but eventually we'll probably need to extract that to a separate, shared crate.

Would you like to tackle that?

nitnelave:
For point number 2 about when the password was changed, I'd file that as a separate issue (which can later have a PR)

app/src/components/change_password.rs
gplubeck: Seems reasonable enough. Stylistically would you rather them each be their own validation function?  E.g. password would get several validators (empty_or_long, has_uppercase, has_lowercase, has_symbol, etc.).  Or do you prefer it to all be in a single function?

gplubeck:
> The change on its own looks not bad. However, as you noted, it should be togglable through configuration. There's no _easy_ way to pass config to the frontend for now, and it _has_ been a problem: for instance, to detect if password resets are enabled, we do an awkward query for an unusual login name and check the return value, which has broken before.
> 
> In short, I would like to address the configuration problem first (#514), before merging this. There are some ideas in the ticket, I think the simplest is to query a config in json. For now we can copy paste the config struct in client and server code, but eventually we'll probably need to extract that to a separate, shared crate.
> 
> Would you like to tackle that?

This may be pushing my rust bounds, but I might be able to look at it this weekend.

nitnelave:
For implementation details, we'll need to:
- make a new struct in configuration.rs for the UI options
- pass that into auth.rs
- have a new route that serves the UI options as json (different struct
since we also want to add the password reset enabled or not)
That's the first PR.

The second pr is adding the fetch in app.rs on initial load, saving the
options, and passing the correct ones to login.rs

tecosaur:
I do wonder if this is well advised. I'm reminded of https://github.com/authelia/authelia/issues/7799 where another ill-advised but popular feature (password expiry) is asked for. My feeling (FWIW) is that despite their popularity, since we know by now is that these requirements actually *decrease* security by making users choose worse passwords and recent NIST gudelines etc. reflect this by making the absense of these reuirments mandatory in recent standards.

nitnelave:
Many words have already been spilled on this topic (see the issues mentioned in the description).

The outcome is: it's a bad feature, but one that is often required by security audits (for better or for worse), and the lack of it makes LLDAP a no-go. In the end, I won't personally implement it, but I'm open to pull requests and any docs will have a warning and a link to the NIST recommendations.

tecosaur:
Cool, with "this is a bad idea"-type warnings that sounds like a practical compromise.

nitnelave:
Closing this PR due to inactivity. If anyone is interested, they can create a new one or re-open this one.

gplubeck:
Work and life have got the best of me for the past couple months.  I apologize.  Has the front end retrieval of the configuration been solved to your satisfaction yet? If not, I might create a new branch and have a go at it. If yes, I might start tackling the original password complexity request.

nitnelave:
@gplubeck yeah, I finally found the time to do a bit of refactoring :)

Just to be clear, what do you plan on implementing? I'd rather talk about it first than have to tell you that I won't merge what you wrote :D

gplubeck:
Oh wow yeah the source has changed a ton!  Congrats on the refactor and thanks for continuing to put in so much work.

Looking at the current version of the code base and without playing with it yet, it seems like there are a few options.  If we only want to enforce password complexity on users without command line access (I am okay with this), I think we could add it to the front-end-options.  Inside /crates/frontend-options/src/lib.rs I can either add a struct of several booleans (e.g. password length, uppercase, lower case, etc).  Or if we want to keep it very simple it can be a single bool that requires all of the listed.  Then using that options config, I can enforce password complexity inside app/src/change_password.rs

Or if you want it to be enforced globally I can add it to the server/src/configuation.rs Configuration struct and maybe place the logic in set-password/src/main.rs to enforce complexity options.

nitnelave:
I think expressing all the requirements as a regex should cover most bases (with "special characters" defined as "anything that's not a letter or digit" for instance), though it might be the most readable for end users. 

I still _really_ don't like anything other than password length, leak check and mayyyybe entropy checks, but I think that's what the crowd wants. Looking at the Bitwarden password generator, I think we'd need:
- length
- uppercase required?
- lowercase required?
- digits required?
- special characters required: list of characters (empty for no requirements)

And we'll need it both on the server (for LDAP password changes) and on the client. Come to think of it, the lldap_set_user_password should ideally fetch those options and apply them as well, unless the given (new) flag to ignore the restrictions is passed.

WDYT?

gplubeck:
Certainly agree with regex. I will likely implement it similar to my last approach, but hopefully a bit more elegant.  You implemented length that can be bypassed for users.  I think uppercase, lowercase, digits, and special characters covers most generic policy demands.

In terms of not liking it, I am also in agreement.  Which is why if I were in your position I would require it be a config option defaulted to off.  Maybe in the spirit good security practices we split them into two separate options?  For example, the current bypass_password_policy bool becomes bypass_recommended_password_policy which defaults to false.  Then a second, enforce_pseudo_password_complexity  that defaults to false enforces the above options?

I am not sure I love how those things contradict a bit.  i.e. bypass false means the options is on and enforce false means the options is off.  But I think it would be beneficial to distinguished between what is recommended very having to opt in to a worse practice to satisfy whatever requirements are being put on you.

Side note, I might be blind, but I don't see lldap_set_user_password in the code base.  Where should I be looking?

nitnelave:
I think a simple bypass option would be better.

As for the password setting code, it's in the top-level "set-password" folder

gplubeck:
Okay sounds good.  Just to be clear, one bypass option that wraps up length, uppercase, lower case, digits, and special characters?

I think I should be able to handle that.

PR #1037, opened 2024-11-13
Updated README.md

Teryl:
Added required environment variables to specify during the first run of the docker container

The curveball is how changing the config after youve messed up doesnt seem to edit the admin user in the user database once its created for the first time.

Perhaps this is an indication of a deeper issue in how the admin user is created in the first place, but this solution works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ services:
       - UID=####
       - GID=####
       - TZ=####/####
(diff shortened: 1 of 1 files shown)


nitnelave:
As mentioned in Discord, we're going to do something else to ensure a better user experience.

Issue #579, opened 2023-05-12
Get rid of git-versioned dependencies


nitnelave:
- [ ] lber: https://github.com/inejge/ldap3/issues/104
- [ ] opaque-ke: revert code to the published version, drop the type safety
- [ ] yew_form: get rid of it.

This blocks release of any crate.

nitnelave:
For yew-form, replacing it is complicated, there is no other good solution for forms with yew (apart from lots of hand-written boilerplate), and the crate is actually quite nice. I'm offering to step up to maintain it there: https://github.com/jfbilodeau/yew_form/issues/22

PR #1040, opened 2024-11-16
cargo,auth,server: update opaque-ke => 0.7

jduepmeier:
This changes the opaque-ke dependency back to the upstream and updates it to 0.7.

I think the changes from the fork were merged in 0.7.

See https://github.com/facebook/opaque-ke/releases/tag/v0.7.0

I don't know if a version higher than 0.7 does work.

This should fix the todo in #579.


nitnelave (approved):
Nice! Thanks :)

PR #1043, opened 2024-11-20
Update onedev.md

xeoneox:
fix capitalization and expound URL example

### DIFF
--- a/example_configs/onedev.md
+++ b/example_configs/onedev.md
@@ -1,8 +1,8 @@
-# Configuration for Onedev
+# Configuration for OneDev
 In Onedev, go to `Administration > Authentication Sources` and click `External Authentication`
 Select `Generic LDAP`
 
-* LDAP URL: Your lldap server's ip/hostname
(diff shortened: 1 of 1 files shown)


example_configs/onedev.md
```diff
@@ -1,8 +1,8 @@
-# Configuration for Onedev
+# Configuration for OneDev
 In Onedev, go to `Administration > Authentication Sources` and click `External Authentication`
 Select `Generic LDAP`
 
-* LDAP URL: Your lldap server's ip/hostname
+* LDAP URL: ldap:// or ldaps:// (Your lldap server's ip/hostname) :3890 or :6360
```
nitnelave, line 5: It might be clearer to write it as "`ldap://lldap_ip:3890` or `ldaps://lldap_ip:6360`"

nitnelave (changes requested):

example_configs/onedev.md
xeoneox: that works; done

nitnelave (approved):

PR #1050, opened 2024-11-24
Add Kimai config example

Elbullazul:


example_configs/kimai.yaml
```diff
@@ -0,0 +1,36 @@
+# See https://www.kimai.org/documentation/local-yaml.html
+# this file should be renamed local.yaml
+kimai:
+  ldap:
+    activate: true
+    connection:
+      host: lldap
+      port: 3890
+      useSsl: false
```
nitnelave, line 9: Maybe add a comment "true if using ldaps"?

example_configs/kimai.yaml
```diff
@@ -0,0 +1,36 @@
...
+      username: uid=admin,ou=people,dc=example,dc=com
+      password: <PASSWORD_HERE>
+      accountFilterFormat: (&(objectClass=person)(uid=%s))
+      bindRequiresDn: true
+      optReferrals: false
+
+    user:
+      baseDn: ou=people, dc=example, dc=com
+      usernameAttribute: uid
+      filter: (&(objectClass=person)(memberof=cn=kimai,ou=groups,dc=example,dc=com))
```
nitnelave, line 20: Add a top level note that you need to create the "kimai" group and add users to it

nitnelave (changes requested):
Looks good, thanks !

example_configs/kimai.yaml
Elbullazul: Would moving the group filtering condition to a comment do the trick? Kimai doesn't require a group for LDAP to work.

nitnelave: Yeah, you can move it to a comment, but still add a sentence (in the same comment) about the need to create the group.

nitnelave (approved):

PR #1071, opened 2025-01-08
server: Add ldap subschema entry

### LINKED ISSUE #893: [FEATURE REQUEST]  Return schema in RootDSE for better compatibility with Apache Directory Studio Browser
**Motivation**
I want to be able to use Apache Directory Studio Browser to browse LLDAP server

When creating the connection to LLDAP in Apache Directory Studio, you must enter the baseDN manually because LLDAP is "Missing schema location in RootDSE" and the baseDN(s) cannot be fetched. 

I can make this work half way by creating two connections and manually identifying one baseDN as `ou=people,dc=example,dc=com` or `ou=groups,dc=example,dc=com`.  However, they must be separate connections.  It would be nice if there was an OU one level up or some other way to identify the baseDN.  Even better would be for that baseDN to be fetchable by the Apache Directory Studio.

T0byV:
This PR aims to resolve #893. It adds a subschema entry to the rootDSE, which shows all attributes and objectclasses present on the LLDAP server, which is needed for some applications that need to index the LDAP server. The current implementation's goal is to have a bare minimum working subschema which follows the LDAP RFC. It also updates the GraphQL interface to follow the changes that have been made in actually seperating out objectclasses, instead of having them as an attribute.

~~Currently, this is work in progress and no tests have been implemented yet. 
I'm publishing this PR as a draft as it's not ready to be merged in the state it is currently in, but it might be a good start for others who would like to continue on this issue, as my progress has stalled for more than a month. Feel free to contribute / use this for inspiration.~~
Tests were added, and the PR is now in a presentable state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ service that seems definitely incompatible with LLDAP.
 - [Discord Integration](docs/faq.md#discord-integration)
 - [Migrating from SQLite](docs/faq.md#migrating-from-sqlite)
 - How does lldap compare [with OpenLDAP](docs/faq.md#how-does-lldap-compare-with-openldap)? [With FreeIPA](docs/faq.md#how-does-lldap-compare-with-freeipa)? [With Kanidm]?(docs/faq.md#how-does-lldap-compare-with-kanidm)
+- [Does lldap support vhosts?](docs/faq.md#does-lldap-support-vhosts)
 
 ## Contributions
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -4,6 +4,7 @@
 - [Discord Integration](#discord-integration)
 - [Migrating from SQLite](#migrating-from-sqlite)
 - How does LLDAP compare [with OpenLDAP](#how-does-lldap-compare-with-openldap)? [With FreeIPA](#how-does-lldap-compare-with-freeipa)? [With Kanidm](#how-does-lldap-compare-with-kanidm)?
+- [Does LLDAP support vhosts?](#does-lldap-support-vhosts)
 
 ## I can't log in!
 
@@ -81,3 +82,12 @@ read-only, and by having more moving parts it is inherently more complex. If
 you don't need to modify the users through LDAP and you're planning on
 installing something like [KeyCloak](https://www.keycloak.org) to provide
 modern identity protocols, check out Kanidm.
+
+## Does LLDAP support vhosts?
+
+LLDAP does not natively support virtualhosts, sometimes known as multi-tenancy:
+
+- users are all part of the same base DN (`ou=people,dc=example,dc=com`)
+- you may support multiple domains by using fully-qualified email addresses as usernames (eg. `user@example.com` and `user@example.org`); however, you can't have more fine-grained permissions than what is provided by default (which applies across all users)
+
+LLDAP is very lightweight (~15MiB RAM on startup) and it's therefore possible to run one instance per virtualhost if you need to properly support multiple domains.


server/src/infra/graphql/query.rs
```diff
@@ -530,9 +547,48 @@ impl<Handler: BackendHandler> AttributeList<Handler> {
...
+            "posixAccount".to_string(),
+            "mailAccount".to_string(),
+            "person".to_string(),
+        ]
+    }
+
+    fn object_classes(&self) -> Vec<ObjectClassGQL> {
+        let mut all_object_classes: Vec<ObjectClassGQL> = vec![
+            ObjectClassGQL {
+                object_class: "inetOrgPerson".to_string(),
```
broeng, line 566: These should probably just be generated based on the vector defined above (or the shared definition).

server/src/infra/ldap_handler.rs
```diff
@@ -204,10 +207,310 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+                    is_visible: true,
+                    is_editable: false,
+                    is_hardcoded: true,
+                    is_readonly: true,
+                },
+            ],
+        },
+        user_object_classes: ObjectClassList (
+            vec![
+            LdapObjectClass::from("inetOrgPerson"),
```
broeng, line 429: Maybe also worth generating from a shared definition of user object classes.

server/src/infra/ldap_handler.rs
```diff
@@ -204,10 +207,310 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+pub struct HardcodedSchema {
+    user_attributes_must: AttributeList,
+    user_attributes_may: AttributeList,
+    group_attributes_must: AttributeList,
+    group_attributes_may: AttributeList,
+    user_object_classes: ObjectClassList,
+    group_object_classes: ObjectClassList,
+}
+
+impl HardcodedSchema {
```
broeng, line 235: Just a naming suggestion, but maybe LldapSchema/LldapBuiltinSchema?

broeng (commented):
Just a few suggestions. I'm not really familiar enough with the codebase to give any in depth review at this point :-)

Think it looks good though!

server/src/infra/graphql/query.rs
```diff
@@ -530,9 +547,48 @@ impl<Handler: BackendHandler> AttributeList<Handler> {
...
 
-    fn extra_ldap_object_classes(&self) -> Vec<String> {
+    fn extra_object_classes(&self) -> Vec<String> {
         self.extra_classes.iter().map(|c| c.to_string()).collect()
     }
+
+    fn hardcoded_object_classes(&self) -> Vec<String> {
+        vec![
+            "inetOrgPerson".to_string(),
+            "posixAccount".to_string(),
```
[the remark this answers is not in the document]
T0byV: The list is now in a single place in the `'user.rs` and `group.rs` file where the `DEFAULT_USER_OBJECT_CLASSES` and `DEFAULT_GROUP_OBJECT_CLASSES` are now defined

server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +210,333 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+                vals: vec![b"cn=Subschema".to_vec()],
+            },
+        ],
+    })
+}
+
+pub struct ObjectClassList(Vec<LdapObjectClass>);
+
+impl ObjectClassList {
+    fn format_for_schema(&self) -> String {
```
nitnelave, line 224: format_for_ldap_schema_description maybe?
There was another instance of that function, let's also align.

For both, could you point to the RFC paragraph that describes the format?

server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +210,333 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+        self.0
+            .iter()
+            .map(|c| format!("'{}'", c))
+            .unique()
+            .collect::<Vec<_>>()
+            .join(" ")
+    }
+}
+
+pub struct LldapBuiltinSchema {
```
nitnelave, line 234: I'm not a big fan of this name. Lldap doesn't bring much value :)
And built-in sounds like hardcoded. Maybe LdapSchemaDescription?

And let's add a link to the RFC section, as well as a comment for what this struct represents.

server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +210,333 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+        self
+    }
+
+    fn formatted_attribute_list(&self) -> Vec<Vec<u8>> {
+        let mut formatted_list: Vec<Vec<u8>> = Vec::new();
+
+        for (index, attribute) in self.all_attributes().attributes.into_iter().enumerate() {
+            formatted_list.push(
+                format!(
+                    "( 2.{} NAME '{}' DESC 'LLDAP: {}' SUP {:?} )",
```
nitnelave, line 267: What's that LLDAP?

server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +210,333 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+    user_attributes_must: AttributeList,
+    user_attributes_may: AttributeList,
+    group_attributes_must: AttributeList,
+    group_attributes_may: AttributeList,
+    user_object_classes: ObjectClassList,
+    group_object_classes: ObjectClassList,
+}
+
+impl LldapBuiltinSchema {
+    fn extend_with_custom_schema(mut self, schema: &Schema) -> Self {
```
nitnelave, line 244: Instead of this function and `get_lldap_builtin_schema`, I would rather have 2 constructors for the struct: `hardcoded_schema() -> Self` and `from_schema(Schema) -> Self`

Or even better, use a default Schema and don't hardcode anything in this class, if possible (I'm a bit fuzzy on where those attributes are defined and how reusable the code is). Ideally we'd have a single source of truth for hardcode attributes

nitnelave: You'd just have to split them into may and must

server/src/domain/ldap/group.rs
```diff
@@ -20,6 +20,8 @@ use crate::domain::{
     types::{AttributeName, AttributeType, Group, GroupId, LdapObjectClass, UserId, Uuid},
 };
 
+pub const DEFAULT_GROUP_OBJECT_CLASSES: &[&str] = &["groupOfUniqueNames"];
```
nitnelave, line 23: Can we make that not pub, and instead expose a function that returns a `Vec<Vec<u8>>`? That'll simplify the rest of the code.
And maybe another one that returns a vec of LDAP object class

server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +210,333 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+            atype: "modifyTimestamp".to_string(),
+            vals: vec![current_time_utc.to_vec()],
+           },
+           LdapPartialAttribute {
+            atype: "ldapSyntaxes".to_string(),
+            vals: vec![
+                b"( 1.3.6.1.4.1.1466.115.121.1.15 DESC 'Directory String' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.24 DESC 'Generalized Time' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.27 DESC 'Integer' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.28 DESC 'JPEG' X-NOT-HUMAN-READABLE'TRUE' )".to_vec(),
```
nitnelave, line 505: Space before true?

server/src/infra/ldap_handler.rs
```diff
@@ -538,6 +871,14 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
                     ]);
                 }
             }
+        } else if request.base == "cn=Subschema" && request.scope == LdapSearchScope::Base {
+            debug!("Subschema request made");
```
nitnelave, line 875: Is there a separate schema request, or is that just the name for it? In which case, let's use "schema" in the debug message.

server/src/infra/ldap_handler.rs
```diff
@@ -2687,6 +3028,29 @@ mod tests {
...
+            aliases: LdapDerefAliases::Never,
+            sizelimit: 0,
+            timelimit: 0,
+            typesonly: false,
+            filter: LdapFilter::Present("objectClass".to_string()),
+            attrs: vec!["supportedExtension".to_string()],
+        };
+        assert_eq!(
+            ldap_handler.do_search_or_dse(&request).await,
+            Ok(vec![schema_response(&schema), make_search_success()])
```
nitnelave, line 3050: Here, the only thing that you're testing is essentially the query routing, not the code implementation.

Could you replace the call to schema_response with the actual response?
It also helps to see what is returned, maybe highlighting some errors.

nitnelave (commented):
The approach looks good! I have a bunch of small comments, but nothing big.

server/src/domain/handler.rs
```diff
@@ -173,6 +174,15 @@ impl AttributeList {
         self.get_attribute_schema(name)
             .map(|a| (a.attribute_type, a.is_list))
     }
+
+    pub fn format_for_schema(&self) -> String {
+        self.attributes
+            .iter()
+            .map(|a| a.name.as_str())
+            .unique()
```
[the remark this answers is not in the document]
T0byV: Well this is an interesting issue, if we look at the following ldapsearch for subschema snippet (no extra object classes):
```
objectClasses: ( 3.0 NAME ( 'inetOrgPerson' 'posixAccount' 'mailAccount' 'person' ) DESC 'LLDAP builtin: a person' STRUCTURAL MUST ( uid $ mail ) MAY ( objectclass $ givenname $ sn $ cn $ jpegphoto $ createtimestamp $ entryuuid $ avatar $ first_name $ last_name ) )
```
We can take the example of `posixAccount`, which is a default builtin objectclass. The issue here is that when I try to create an additional objectclass, it will happily do so since there is not actually a database entry with this name (since the default entry is in the list of constants). Now I don't know the default behavior of LDAP clients (and couldn't find it), but removing the duplicate `posixAccount` should fix this issue altogether if it comes up.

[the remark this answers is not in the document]
T0byV: Would it at that point not be better to add the `default` entries to the DB as well?

server/src/infra/ldap_handler.rs
T0byV: I'm not entirely sure if I understand you correctly, what do you mean by actual response?

server/src/infra/ldap_handler.rs
T0byV: I put it there as sort of a descriptor for the fact that the attribute is a LLDAP custom attribute and not a standard LDAP (RFC) attribute, do you think it's unnecessary/name should be changed?

server/src/infra/ldap_handler.rs
T0byV: I see what you mean, only thing is that you need the following variables:
```
                    name: "mail".into(),
                    attribute_type: AttributeType::String,
                    is_list: false,
                    is_visible: true,
                    is_editable: false,
                    is_hardcoded: true,
                    is_readonly: true,
```
These are not present in the other occurrence of these entries.
Maybe adding these 'hardcoded' entries to the DB with this information is the best course of action, what's your opinion?

server/src/infra/graphql/query.rs
```diff
@@ -515,10 +519,28 @@ impl<Handler: BackendHandler> From<DomainAttributeSchema> for AttributeSchema<Ha
...
 }
 
+#[derive(Clone)]
+pub struct ObjectClassGQL {
+    object_class: String,
+    is_hardcoded: bool,
+}
+
+#[graphql_object]
+impl ObjectClassGQL {
```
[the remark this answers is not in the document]
T0byV: I changed this to `ObjectClassInfo`, found it maybe a bit more fitting, do you agree?

server/src/infra/graphql/query.rs
```diff
@@ -530,15 +552,43 @@ impl<Handler: BackendHandler> AttributeList<Handler> {
...
+            .collect();
+
+        all_object_classes.extend(
+            self.extra_classes
+                .iter()
+                .map(|c| ObjectClassGQL {
+                    object_class: c.to_string(),
+                    is_hardcoded: false,
+                })
+                .collect::<Vec<_>>(),
```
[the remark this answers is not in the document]
T0byV: Good catch

server/src/domain/handler.rs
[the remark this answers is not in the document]
T0byV: So for now, what would you say we do with this?

server/src/infra/graphql/query.rs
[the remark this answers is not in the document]
nitnelave: SGTM!

server/src/infra/ldap_handler.rs
nitnelave: No, that's fine, I'm just not super familiar with the protocol

server/src/infra/ldap_handler.rs
nitnelave: I think you can use the already-implemented default schema as source, and just have a set of required parameter names.

In practice: build a default, empty `domain::handler::Schema`, create a `PublicSchema` from it (that'll populate the hardcode attributes), and then you can use that to create your default schema.

Something along these lines.

nitnelave:
Please tell me if it's ready for review again, I might have missed something

server/src/domain/handler.rs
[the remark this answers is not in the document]
T0byV: Shall we close this thread for now? I removed the deduplication in the subschema response for now, implementing something that checks for duplicates when creating objects on the database level is the way to go here, but better to create a seperate PR for that.

server/src/domain/ldap/group.rs
```diff
@@ -22,6 +22,31 @@ use lldap_domain::types::{
...
+const DEFAULT_GROUP_OBJECT_CLASSES: &[&str] = &["groupOfUniqueNames"];
+
+pub fn get_required_group_attributes() -> Vec<AttributeName> {
+    REQUIRED_GROUP_ATTRIBUTES
+        .iter()
+        .map(|a| AttributeName::from(a.to_string()))
+        .collect()
+}
+
+fn get_default_group_object_classes_vec_u8() -> Vec<Vec<u8>> {
```
nitnelave, line 36:
```suggestion
fn get_default_group_object_classes_as_bytes() -> Vec<Vec<u8>> {
```

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+    }
+}
+
+// See RFC4512 section 4.2.1 "objectClasses"
+impl ObjectClassList {
+    fn format_for_ldap_schema_description(&self) -> String {
+        self.0
+            .iter()
+            .map(|c| format!("'{}'", c))
+            .unique()
```
nitnelave, line 249: Please add a todo to remove this unique as well

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+// See RFC4512 section 4.2 "Subschema Subentries"
+// This struct holds all information on what attributes and objectclasses are present on the server.
+// It can be used to 'index' a server using a LDAP subschema call.
+pub struct LdapSchemaDescription {
+    base: Schema,
+    user_object_classes: ObjectClassList,
+    group_object_classes: ObjectClassList,
+}
+
+impl Deref for LdapSchemaDescription {
```
nitnelave, line 264: Implementing `Deref` is unusual, it's usually reserved for smart pointer-like objects. Why not use the `.base` field everywhere? Or a `.schema()` method.

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+
+impl LdapSchemaDescription {
+    fn from(schema: &Schema) -> Self {
+        let mut user_object_classes = get_default_user_object_classes();
+        user_object_classes.extend(schema.extra_user_object_classes.clone());
+        let mut group_object_classes = get_default_group_object_classes();
+        group_object_classes.extend(schema.extra_group_object_classes.clone());
+
+        Self {
+            base: PublicSchema::from(schema.clone()).get_schema().clone(),
```
nitnelave, line 280: I think we can add a `.into_schema` to `PublicSchema` to avoid that last `.clone()`

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+impl Deref for LdapSchemaDescription {
+    type Target = Schema;
+
+    fn deref(&self) -> &Self::Target {
+        &self.base
+    }
+}
+
+impl LdapSchemaDescription {
+    fn from(schema: &Schema) -> Self {
```
nitnelave, line 273: This should take the schema by value, to let the caller decide whether to clone it or not, since we need a value anyway.
Even better, this can take a `PublicSchema` by value, so we never forget to do the conversion. Semantically, it operates on the public schema anyway.

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+            group_object_classes: ObjectClassList(group_object_classes),
+        }
+    }
+
+    fn required_user_attributes(&self) -> AttributeList {
+        let attributes = self
+            .user_attributes
+            .attributes
+            .iter()
+            .filter(|&a| get_required_user_attributes().contains(&a.name))
```
nitnelave, line 291: Here, you're creating a `Vec` for each user attribute (the `get_required_user_attributes()`).
I think we can expose the underlying slice of names (the `&[&str]`) and directly use it here, so we don't have to create _any_ vector. At worst, create the vector once outside the loop.

Same for the other methods below.

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+
+    // See RFC4512 section 4.2.2 "attributeTypes"
+    fn formatted_attribute_list(&self) -> Vec<Vec<u8>> {
+        let mut formatted_list: Vec<Vec<u8>> = Vec::new();
+
+        for (index, attribute) in self.all_attributes().attributes.into_iter().enumerate() {
+            formatted_list.push(
+                format!(
+                    "( 2.{} NAME '{}' DESC 'LLDAP: {}' SUP {:?} )",
+                    (index + 4),
```
nitnelave, line 342: What's that value 4? Can you extract it to a named constant so the reader knows where it's from?

server/src/infra/ldap_handler.rs
```diff
@@ -568,6 +791,23 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
                     ]);
                 }
             }
+        } else if request.base == "cn=Subschema" && request.scope == LdapSearchScope::Base {
+            // See RFC4512 section 4.4 "Subschema discovery"
+            debug!("Schema request made");
```
nitnelave, line 796:
```suggestion
            debug!("Schema request");
```

server/src/infra/ldap_handler.rs
```diff
@@ -211,6 +224,216 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
...
+    pub fn all_attributes(&self) -> AttributeList {
+        let mut combined_attributes = self.user_attributes.attributes.clone();
+        combined_attributes.extend(self.group_attributes.attributes.clone());
+        AttributeList {
+            attributes: combined_attributes,
+        }
+    }
+}
+
+fn schema_response(schema: &Schema) -> LdapOp {
```
nitnelave, line 368: Same here, we can propagate the need for a value, not a reference. And then in the callsite, it becomes obvious that we didn't need to clone at all :)
Up to you whether this should take a `PublicSchema` or not, but I think it should.

Can you expand on the name a bit? `make_ldap_subschema_entry` maybe?

server/src/infra/ldap_handler.rs
```diff
@@ -2772,6 +3012,97 @@ mod tests {
...
+                    atype: "cn".to_owned(),
+                    vals: vec![b"Subschema".to_vec()]
+                },
+                LdapPartialAttribute {
+                    atype: "createTimestamp".to_owned(),
+                    vals: vec![current_time_utc.to_vec()]
+                },
+                LdapPartialAttribute {
+                    atype: "modifyTimestamp".to_owned(),
+                    vals: vec![current_time_utc.to_vec()]
```
nitnelave, line 3053: I hope no one is relying on that to do cache invalidation... Not actionable for now, I guess. Since we're lying either way, it's probably better to say that it always "changed" rather than always returning the epoch and have an old entry stay in a cache.

server/src/infra/ldap_handler.rs
```diff
@@ -2772,6 +3012,97 @@ mod tests {
...
+                LdapPartialAttribute {
+                    atype: "attributeTypes".to_owned(),
+                    vals: vec![
+                        b"( 2.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
+                        b"( 2.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
+                        b"( 2.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
+                        b"( 2.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                        b"( 2.4 NAME 'avatar' DESC 'LLDAP: builtin attribute' SUP JpegPhoto )".to_vec(),
+                        b"( 2.5 NAME 'creation_date' DESC 'LLDAP: builtin attribute' SUP DateTime )".to_vec(),
+                        b"( 2.6 NAME 'display_name' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
```
nitnelave, line 3073: How does it work if users and groups have the same attribute name, but with different types? Should that be forbidden?

nitnelave (changes requested):
Looks good overall! Just a few comments here and there, readability and performance, but nothing major.

broeng:
I have an interest in this PR as well, so I've taken the liberty of continuing the PR, and have implemented the last suggestions from @nitnelave, in the hope that we can get this into main. I'm thinking @T0byV can just pull my commits into this PR eventually, instead of me making a separate PR (think the credit for the work should go to him).

I've addressed all nitnelaves comments, and implemented the suggested changes, as I understand them. The only exception is the last comment, which I'm thinking might be a followup for someone, like the attribute name validation?

Unaddressed comment:
> How does it work if users and groups have the same attribute name, but with different types? Should that be forbidden?

My changes can be found in the last commit on this branch:
https://github.com/broeng/lldap/commits/add-ldap-subschema/

Specifically, this commit:
https://github.com/broeng/lldap/commit/3037ee7a3bc106455a973a84e51aa96fa900bf8b

Would you mind reviewing them @nitnelave? Then I'll have a look at getting it rebased with main.

broeng:
I have a rebased branch ready (with tests, fmt and clippy ran), that we can use here:
https://github.com/broeng/lldap/tree/add-ldap-subschema-rebased

Commits have been squashed to a single commit, to not make the conflict resolution worse than it had to be. @T0byV is still listed as author.

nitnelave:
I left a few comments on your commit, I hope you can see them.

crates/domain/src/schema.rs
```diff
@@ -36,4 +36,13 @@ impl AttributeList {
         self.get_attribute_schema(name)
             .map(|a| (a.attribute_type, a.is_list))
     }
+
+    pub fn format_for_ldap_schema_description(&self) -> String {
+        self.attributes
+            .iter()
+            .map(|a| a.name.as_str())
+            // .unique()
```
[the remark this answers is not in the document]
broeng: Commenting as I made that change; I removed the call to unique, and after feedback, removed the proposed TODO again as well.

server/src/domain/ldap/group.rs
broeng: Renamed as proposed.

server/src/infra/ldap_handler.rs
broeng: I removed the call to unique, and also the proposed todo after a review iteration, as it should be enforced by DB.

server/src/infra/ldap_handler.rs
broeng: Changed as proposed.

server/src/infra/ldap_handler.rs
broeng: Done

server/src/infra/ldap_handler.rs
broeng: Changed to take the PublicSchema by value.

server/src/infra/ldap_handler.rs
broeng: Changed to use the underlying slice of names directly, same for the methods below.

server/src/infra/ldap_handler.rs
broeng: The constant 4 was an offset needed as the builtin attributes had already been added earlier. Extracted it to a parameter to the function, and passed it from the call-site (with a comment) where the builtin attributes are added, so the connection is clear.

server/src/infra/ldap_handler.rs
broeng: Done

server/src/infra/ldap_handler.rs
broeng: Changed to take the PublicSchema, as proposed, and renamed to the suggested name instead.

server/src/infra/ldap_handler.rs
broeng: As it's deemed not actionable, I didn't make any changes to this. Suggest we just mark this as resolved.

broeng:
Dropped some replies to @nitnelave's review comments on what I changed, and what I didn't. The trivial changes I marked as resolved.

crates/domain/src/schema.rs
```diff
@@ -1,6 +1,7 @@
 use serde::{Deserialize, Serialize};
 
 use crate::types::{AttributeName, AttributeType, LdapObjectClass};
+use itertools::Itertools;
```
nitnelave: Why do we need this? Maybe scope the import to just the function that needs it.

crates/ldap/src/core/utils.rs
```diff
@@ -330,6 +337,148 @@ pub fn get_custom_attribute(
         })
 }
 
+pub struct ObjectClassList(Vec<LdapObjectClass>);
```
nitnelave, line 341:
```suggestion
#[derive(derive_more::From)]
pub struct ObjectClassList(Vec<LdapObjectClass>);
```
And you can skip the manual impl

crates/ldap/src/core/utils.rs
```diff
@@ -330,6 +337,148 @@ pub fn get_custom_attribute(
...
+    }
+}
+
+// See RFC4512 section 4.2.1 "objectClasses"
+impl ObjectClassList {
+    pub fn format_for_ldap_schema_description(&self) -> String {
+        self.0
+            .iter()
+            .map(|c| format!("'{}'", c))
+            .collect::<Vec<_>>()
```
nitnelave: We can use https://docs.rs/itertools/0.7.2/itertools/fn.join.html to avoid the intermediate `collect`

crates/ldap/src/core/utils.rs
```diff
@@ -330,6 +337,148 @@ pub fn get_custom_attribute(
...
+            .join(" ")
+    }
+}
+
+// See RFC4512 section 4.2 "Subschema Subentries"
+// This struct holds all information on what attributes and objectclasses are present on the server.
+// It can be used to 'index' a server using a LDAP subschema call.
+pub struct LdapSchemaDescription {
+    base: Schema,
```
nitnelave: I think this class should store a `PublicSchema`, and we can unwrap it in the `schema()` method.

That way we don't have to add the `PublicSchema::into_schema()` method either.

crates/ldap/src/core/utils.rs
```diff
@@ -330,6 +337,148 @@ pub fn get_custom_attribute(
...
+        &self.group_object_classes
+    }
+
+    pub fn required_user_attributes(&self) -> AttributeList {
+        let attributes = self
+            .schema()
+            .user_attributes
+            .attributes
+            .iter()
+            .filter(|&a| REQUIRED_USER_ATTRIBUTES.contains(&a.name.as_str()))
```
nitnelave: That's a bit too many `&` :)

```suggestion
            .filter(|a| REQUIRED_USER_ATTRIBUTES.contains(&a.name.as_str()))
```
Same below

crates/ldap/src/core/utils.rs
```diff
@@ -330,6 +337,148 @@ pub fn get_custom_attribute(
...
+                .to_vec(),
+            )
+        }
+
+        formatted_list
+    }
+
+    pub fn all_attributes(&self) -> AttributeList {
+        let mut combined_attributes = self.schema().user_attributes.attributes.clone();
+        combined_attributes.extend(self.schema().group_attributes.attributes.clone());
```
nitnelave:
```suggestion
        combined_attributes.extend_from_slice(self.schema().group_attributes.attributes);
```

crates/ldap/src/search.rs
```diff
@@ -164,6 +165,82 @@ pub(crate) fn root_dse_response(base_dn: &str) -> LdapOp {
...
+            vals: vec![
+                b"( 1.3.6.1.4.1.1466.115.121.1.15 DESC 'Directory String' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.24 DESC 'Generalized Time' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.27 DESC 'Integer' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.28 DESC 'JPEG' X-NOT-HUMAN-READABLE 'TRUE' )".to_vec(),
+                ],
+           },
+           LdapPartialAttribute {
+            atype: "attributeTypes".to_string(),
+            vals: vec![
```
nitnelave:
```suggestion
            vals: [
```

You have the `collect` making the final `Vec`, here it's just a fixed-size slice. We can even make the magic number 4 more explicit with:
```suggestion
            vals: {
              let hardcoded_attributes = [
                 b"( 2.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
                 b"( 2.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
                 b"( 2.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
                 b"( 2.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
               ];
               let num_hardcoded_attributes = hardcoded_attributes.size();
               hardcoded_attributes.into_iter().chain(
                   ldap_schema_description
                       .formatted_attribute_list(num_hardcoded_attributes)
                 ).collect()
             }
```

server/src/infra/ldap_handler.rs
nitnelave: This is a potential sticking point, let's not forget about it.

crates/ldap/src/core/group.rs
```diff
@@ -18,6 +18,24 @@ use lldap_domain::{
 use lldap_domain_handlers::handler::{GroupListerBackendHandler, GroupRequestFilter};
 use tracing::{debug, instrument, warn};
 
+pub const REQUIRED_GROUP_ATTRIBUTES: &[&str] = &["display_name"];
+
+const DEFAULT_GROUP_OBJECT_CLASSES: &[&str] = &["groupOfUniqueNames"];
```
nitnelave:
```suggestion
const DEFAULT_GROUP_OBJECT_CLASSES: &[&str] = &["groupOfUniqueNames", "groupOfNames"];
```

crates/ldap/src/core/user.rs
```diff
@@ -21,6 +21,25 @@ use lldap_domain_handlers::handler::{UserListerBackendHandler, UserRequestFilter
 use lldap_domain_model::model::UserColumn;
 use tracing::{debug, instrument, warn};
 
+pub const REQUIRED_USER_ATTRIBUTES: &[&str] = &["user_id", "mail"];
+
+const DEFAULT_USER_OBJECT_CLASSES: &[&str] =
```
nitnelave, line 26: Can you use that in `convert_user_filter` as well? Similarly for groups. There might be some casing to handle.

broeng:
> I think the only thing I'm really missing before approving is a good top-level test comparing the LDAP response with a hardcoded string. I thought there was one at some point? Unless I missed them, I didn't see any test in this PR.

I just looked through the old commits, and it looks like there's one there, so I guess it must have gotten lost when I rebased with the refactoring changes on main.

T0byV:
> > I think the only thing I'm really missing before approving is a good top-level test comparing the LDAP response with a hardcoded string. I thought there was one at some point? Unless I missed them, I didn't see any test in this PR.
> 
> I just looked through the old commits, and it looks like there's one there, so I guess it must have gotten lost when I rebased with the refactoring changes on main.

My bad, I thought they were moved somewhere else due to the rebase. I'll get them back into the PR in a bit.

crates/domain/src/schema.rs
T0byV: Removed it where possible and scoped it to `itertools::join`

crates/ldap/src/core/utils.rs
T0byV: I now implemented it like `join(self.0.iter().map(|c| format!("'{}'", c)), " ")`, is that what you meant?

crates/ldap/src/search.rs
T0byV: Implemented this with the `num_hardcoded_attributes` method

crates/ldap/src/core/user.rs
T0byV: Could you please have a look at how It's implemented now? I tested it and it works, but maybe you could quickly check if the way case (in)sensitivity is checked is correct.
(both in `user.rs` and `group.rs`)

T0byV:
Ready for review again, the git history was... interesting so I just copied the code for the test from the head of the old (non-rebased repo).

crates/ldap/src/search.rs
```diff
@@ -437,6 +441,97 @@ mod tests {
...
+            scope: LdapSearchScope::Base,
+            aliases: LdapDerefAliases::Never,
+            sizelimit: 0,
+            timelimit: 0,
+            typesonly: false,
+            filter: LdapFilter::Present("objectClass".to_string()),
+            attrs: vec!["supportedExtension".to_string()],
+        };
+
+        let current_time_utc = Utc::now().format("%Y%m%d%H%M%SZ").to_string().into_bytes();
```
nitnelave: Ooh, I like the attempt, and we might merge as is if you don't feel like fixing this, but this is a recipe for a flaky test: you're basically saying "the interval between generating this string and generating the LLDAP answer does not straddle a second boundary". No matter how small the interval is, that will sometimes be false (and that will more often be the case on overloaded machine, like the cloud test runners can be).
A better test (though more annoying) would be to test each expected element of the response separately, and check that the "createTimestamp" and "modifyTimestamp" are within 5 min of when we started this test.

crates/ldap/src/core/user.rs
nitnelave: LGTM!

crates/ldap/src/core/utils.rs
nitnelave: Perfect.

crates/ldap/src/search.rs
T0byV: Good catch, didn't think of that. I split the response out to multiple asserts now and implemented an assert function that takes a `time_margin` which for now I've set to ±5 minutes. I'm not entirely sure where functions belonging to tests should go, please correct me on that if there's a better way.

nitnelave (approved):

nitnelave:
Hey @T0byV ! Sorry, I didn't see that you sent the cargo format fix. I pulled the PR, refactored a bit the last test (I hope you don't mind!), and squashed everything. It's good to go now!

server/src/infra/ldap_handler.rs
nitnelave: https://github.com/lldap/lldap/issues/1202

Issue #1073, opened 2025-01-12
[BUG] LDAP search for attribute values will only match lowercased strings. Attribute values with any uppercase will never match.


broeng:
**Describe the bug**
When storing attribute values, the values are stored in the DB in the case they are given, serialized with serde. I think this is correct behaviour.

When an LDAP search for a specific attribute value is done, the value in the search filter will have been lowercased, before it is serialized with Serde, and used for lookup in the user_attributes table. This means the serialized value will never match a value that was not also stored in lowercase.

**To Reproduce**
1. Create a new custom user attribute, a string value ('sambaSID' in this example).
2. Assign a value with uppercase letters to this attribute for a user. ('S-1-5-21-2293260629-1257142175-1903762185-3202' in this example).
3. Search for this attribute and value with ldapsearch, with same casing. `ldapsearch ... '(&(objectclass=person)(sambaSID=S-1-5-21-2293260629-1257142175-1903762185-3202))'`
4. Observe no results returned.

**Expected behavior**
Expected to get user back, with the given attribute and value.

**Logs**
Query from LLDAP - observe how value is lowercased in filter expression:

> INFO     LDAP request [ 1.07ms | 5.46% / 100.00% ] session_id: 557bc751-8cee-46f8-ac2a-6db7b5921e4e
> DEBUG    ┝━ 🐛 [debug]:  | msg: LdapMsg { msgid: 2, op: SearchRequest(LdapSearchRequest { base: "dc=example,dc=com", scope: Subtree, aliases: Never, sizelimit: 0, timelimit: 0, typesonly: false, filter: And([Equality("objectclass", "person"), Equality("sambasid", "S-1-5-21-2293260629-1257142175-1903762185-3202")]), attrs: [] }), ctrl: [] }
> DEBUG    ┝━ do_search [ 1.01ms | 33.85% / 94.54% ]
> DEBUG    │  ┝━ 🐛 [debug]:  | request.base: "dc=apc,dc=dk" | scope: Global
> DEBUG    │  ┝━ get_user_list [ 336µs | 1.92% / 31.48% ]
> DEBUG    │  │  ┝━ 🐛 [debug]:  | filters: And([And([]), AttributeEquality(AttributeName(CaseInsensitiveString("sambasid")), Serialized("s-1-5-21-2293260629-1257142175-1903762185-3202"))])
[...]

PR #1074, opened 2025-01-12
server: include preserved case in user attribute value search (#1073)

broeng:
Extends the generated UserRequestFilter with an OR'ed clause for the attribute value in both it's original case and lowercased.

So, where the search was previously `(attr=LOWERCASE(value))` it is now basicly `(|(attr=LOWERCASE(value))(attr=value))`. This matters, because the DB search is case sensitive, so it was only possible to resolve attributes with lowercase values. Original behavior is still compatible.

The `value => value_lc` changes are just variable renaming. The `value` variable, was redefined as a lowercased version of the original `value` variable, and was therefore shadowing the original. As I need the original unmodified value, I thought it best to restore that, and avoid the shadowing.

Fixes issue #1073.

### DIFF
--- a/server/src/domain/ldap/group.rs
+++ b/server/src/domain/ldap/group.rs
@@ -154,12 +154,23 @@ fn get_group_attribute_equality_filter(
     is_list: bool,
     value: &str,
 ) -> GroupRequestFilter {
-    deserialize_attribute_value(&[value.to_owned()], typ, is_list)
-        .map(|v| GroupRequestFilter::AttributeEquality(field.clone(), v))
-        .unwrap_or_else(|e| {
+    let value_lc = value.to_ascii_lowercase();
+    let serialized_value = deserialize_attribute_value(&[value.to_owned()], typ, is_list);
+    let serialized_value_lc = deserialize_attribute_value(&[value_lc.to_owned()], typ, is_list);
+    match (serialized_value, serialized_value_lc) {
+        (Ok(v), Ok(v_lc)) => GroupRequestFilter::Or(vec![
+            GroupRequestFilter::AttributeEquality(field.clone(), v),
+            GroupRequestFilter::AttributeEquality(field.clone(), v_lc),
+        ]),
+        (Ok(_), Err(e)) => {
+            warn!("Invalid value for attribute {} (lowercased): {}", field, e);
+            GroupRequestFilter::from(false)
+        }
+        (Err(e), _) => {
             warn!("Invalid value for attribute {}: {}", field, e);
             GroupRequestFilter::from(false)
-        })
+        }
+    }
 }
 
 fn convert_group_filter(
@@ -171,24 +182,24 @@ fn convert_group_filter(
     match filter {
         LdapFilter::Equality(field, value) => {
             let field = AttributeName::from(field.as_str());
-            let value = value.to_ascii_lowercase();
+            let value_lc = value.to_ascii_lowercase();
             match map_group_field(&field, schema) {
-                GroupFieldType::GroupId => Ok(value
+                GroupFieldType::GroupId => Ok(value_lc
                     .parse::<i32>()
                     .map(|id| GroupRequestFilter::GroupId(GroupId(id)))
                     .unwrap_or_else(|_| {
-                        warn!("Given group id is not a valid integer: {}", value);
+                        warn!("Given group id is not a valid integer: {}", value_lc);
                         GroupRequestFilter::from(false)
                     })),
-                GroupFieldType::DisplayName => Ok(GroupRequestFilter::DisplayName(value.into())),
-                GroupFieldType::Uuid => Uuid::try_from(value.as_str())
+                GroupFieldType::DisplayName => Ok(GroupRequestFilter::DisplayName(value_lc.into())),
+                GroupFieldType::Uuid => Uuid::try_from(value_lc.as_str())
                     .map(GroupRequestFilter::Uuid)
                     .map_err(|e| LdapError {
                         code: LdapResultCode::Other,
                         message: format!("Invalid UUID: {:#}", e),
                     }),
                 GroupFieldType::Member => Ok(get_user_id_from_distinguished_name_or_plain_name(
-                    &value,
+                    &value_lc,
                     &ldap_info.base_dn,
                     &ldap_info.base_dn_str,
                 )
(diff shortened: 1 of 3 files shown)


broeng:
Formatting performed, and suggestion from clippy implemented.

nitnelave:
I'm not convinced this is the right way of doing things. Attribute names are in LDAP domain space, so it makes sense to have case insensitivity, but attribute values are in the  application domain space, and case sensitivity depends on the, well, case. E.g. if you store some base 64 string in there, a case insensitive match is not what you want. If you do want case insensitive matching, then you can store the lowercased version in the app code (or there are probably some unsupported extensions to support that, knowing LDAP).
I'm curious how other servers behave.

I agree that matching against the lowercase is wrong, but I disagree with the fix (if only because searching for the lowercased version will only help if the actual stored case is lowercase, so it's not a good case insensitive comparison). I propose to only match the case.

nitnelave (changes requested):
See above comment, just filing this as a proper review

broeng:
> I'm not convinced this is the right way of doing things. Attribute names are in LDAP domain space, so it makes sense to have case insensitivity, but attribute values are in the application domain space, and case sensitivity depends on the, well, case. E.g. if you store some base 64 string in there, a case insensitive match is not what you want. If you do want case insensitive matching, then you can store the lowercased version in the app code (or there are probably some unsupported extensions to support that, knowing LDAP). I'm curious how other servers behave.
> 
> I agree that matching against the lowercase is wrong, but I disagree with the fix (if only because searching for the lowercased version will only help if the actual stored case is lowercase, so it's not a good case insensitive comparison). I propose to only match the case.

I don't have a problem with that myself, so I'm perfectly fine to change that behaviour. But just to be absolutely certain, you are aware it's a breaking change? I only ask, because I suspect the renaming to `value_lc` to avoid variable shadowing, might give the opposite impression.

What I am adding in this PR, is case sensitive matching of the value -- not case insensitive matching. And the behavior prior to this was always searching for lowercase values.

nitnelave:
Right, it is technically a breaking change. Even your proposal is a breaking change itself, someone might be relying on the value not matching. 

There's one use case that can potentially break that I know of and would be uncomfortable with: dovecot/postfix search for mailAlias. They look up whether the target email address is in the mail alias. If the email was stored in the DB in lowercase (likely) and an email is sent to the uppercase address (possible), how does postfix search? Do they lowercase before searching? 
If they don't, then we'll go with your PR as formulated. Otherwise, we'll go with my proposal. 

It would be nice to know what OpenLDAP does, though.

broeng:
> Even your proposal is a breaking change itself, someone might be relying on the value not matching.

That's true, I find it a bit more unlikely though, but yes, someone might technically do that.

> There's one use case that can potentially break that I know of and would be uncomfortable with: dovecot/postfix search for mailAlias. They look up whether the target email address is in the mail alias. If the email was stored in the DB in lowercase (likely) and an email is sent to the uppercase address (possible), how does postfix search? Do they lowercase before searching? If they don't, then we'll go with your PR as formulated. Otherwise, we'll go with my proposal.
> 
> It would be nice to know what OpenLDAP does, though.

I can try to look up what postfix does.

But, when I tried to find an answer in the RFC's, initially, I had an impression that it's not uncommon to match case insensitively for values. Absent any actual matching rules, which make the desired matching more explicit in the query. I didn't find any definitive answer though.

Just to be sure I understand correctly, your proposal is to remove the old matching rule which lowercases the filter value before search, right?

broeng:
Postfix seems to be case sensitive, so no lowercasing prior to searching, based on these StackExchange posts:

[PostFix+Dovecot: Case Sensitive Email Address](https://serverfault.com/questions/969671/postfixdovecot-case-sensitive-email-address)

[Are email addresses case sensitive?](https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive)

broeng:
I rebased the PR with main, by the way.

nitnelave:
Okay. In that case, let's go with the PR as currently implemented, preparing the ground for a fully case-insensitive matching later on.

server/src/domain/ldap/user.rs
```diff
@@ -174,12 +174,23 @@ fn get_user_attribute_equality_filter(
...
+        (Ok(v), Ok(v_lc)) => UserRequestFilter::Or(vec![
+            UserRequestFilter::AttributeEquality(field.clone(), v),
+            UserRequestFilter::AttributeEquality(field.clone(), v_lc),
+        ]),
+        (Err(e), _) => {
             warn!("Invalid value for attribute {}: {}", field, e);
             UserRequestFilter::from(false)
-        })
+        }
+        _ => {
```
nitnelave, line 189: Maybe, to be more explicit, match against `(Ok(_), Err(e))`

nitnelave (changes requested):
Can you do the same for group attributes?

broeng:
> Can you do the same for group attributes?

Done now. Had completely missed, that the same would need to be done for groups.

server/src/domain/ldap/group.rs
```diff
@@ -154,12 +154,30 @@ fn get_group_attribute_equality_filter(
...
-        .unwrap_or_else(|e| {
+    let value_lc = value.to_ascii_lowercase();
+    let serialized_value = deserialize_attribute_value(&[value.to_owned()], typ, is_list);
+    let serialized_value_lc = deserialize_attribute_value(&[value_lc.to_owned()], typ, is_list);
+    match (serialized_value, serialized_value_lc) {
+        (Ok(v), Ok(v_lc)) => GroupRequestFilter::Or(vec![
+            GroupRequestFilter::AttributeEquality(field.clone(), v),
+            GroupRequestFilter::AttributeEquality(field.clone(), v_lc),
+        ]),
+        (Err(e), Ok(_)) => {
```
nitnelave, line 165: I think we can still collapse the `(Err(e), _)` case. I just wanted to be explicit about the ok, err

nitnelave (approved):

server/src/domain/ldap/group.rs
broeng: Bit unsure what you mean. Remove the `(Err(e), Ok(_))` case, and let it be caught, by making the `(Err(e), Err(_))` into `(Err(e), _)` ?

broeng: Great, pushed :)

nitnelave (approved):

Issue #1072, opened 2025-01-11
[FEATURE REQUEST] prevent the creation of attributes with underscores


nitnelave:
**Motivation**
It's not part of the RFC, and various clients refuse them. 

**Describe the solution you'd like**
Backend refuses the creation, and ideally the frontend explains why.

broeng:
Implemented with PR #1075

PR #1075, opened 2025-01-13
Add validation and error/warning messages for attribute names (#1072)

broeng:
This commit adds support for basic validation of attribute names at creation, and also in the schema overview. Both user and group attributes are validated with the same rules.

For now, attribute names will be considered valid, if they only contain alphanumeric characters.

Validation has been added the following places:

- In graphql API, for creation of both user and group attributes. Request will be rejected, if attribute name is invalid.

- In frontend, before submitting a request to create a new user or group attribute. Rejection here will show an error message including a list of the invalid characters used.

As this change adds stricter validation to attributes, and, since the rationale for this is partly compatibility with other LDAP systems, this change also adds a warning in the schema overviews to any attribute using invalid characters. As this is unfortunately quite a few of the hardcoded attributes, we should probably consider listing other, valid, aliases for the same attributes instead.

# Screenshots

Backend rejecting creation through GraphQL API (frontend validation disabled here for the screenshot):

backend_rejection

Frontend rejecting creation of invalid attribute pre-submit:

frontend_rejection

Schema attribute overview showing warnings for potentially problematic attributes:

overview_warnings

### DIFF
--- a/app/src/components/create_group_attribute.rs
+++ b/app/src/components/create_group_attribute.rs
@@ -12,6 +12,7 @@ use crate::{
 use anyhow::{bail, Result};
 use gloo_console::log;
 use graphql_client::GraphQLQuery;
+use lldap_validation::attributes::validate_attribute_name;
 use validator_derive::Validate;
 use yew::prelude::*;
 use yew_form_derive::Model;
@@ -62,6 +63,13 @@ impl CommonComponent<CreateGroupAttributeForm> for CreateGroupAttributeForm {
                     bail!("Check the form for errors");
                 }
                 let model = self.form.model();
+                validate_attribute_name(&model.attribute_name).or_else(|invalid_chars| {
+                    let invalid = String::from_iter(invalid_chars);
+                    bail!(
+                        "Attribute name contains one or more invalid characters: {}",
+                        invalid
+                    );
+                })?;
                 let attribute_type = model.attribute_type.parse::<AttributeType>().unwrap();
                 let req = create_group_attribute::Variables {
                     name: model.attribute_name,
--- a/app/src/components/group_schema_table.rs
+++ b/app/src/components/group_schema_table.rs
@@ -157,7 +157,7 @@ impl GroupSchemaTable {
         let desc = group::resolve_group_attribute_description_or_default(&attribute.name);
         html! {
             <tr key={attribute.name.clone()}>
-                <td>{render_attribute_name(hardcoded, &attribute.name, &desc)}</td>
+                <td>{render_attribute_name(hardcoded, &desc)}</td>
                 <td>{if attribute.is_list { format!("List<{attribute_type}>")} else {attribute_type.to_string()}}</td>
                 <td>{if attribute.is_visible {checkmark.clone()} else {html!{}}}</td>
                 {
--- a/app/src/components/user_schema_table.rs
+++ b/app/src/components/user_schema_table.rs
@@ -156,7 +156,7 @@ impl UserSchemaTable {
         let desc = user::resolve_user_attribute_description_or_default(&attribute.name);
         html! {
             <tr key={attribute.name.clone()}>
-                <td>{render_attribute_name(hardcoded, &attribute.name, &desc)}</td>
+                <td>{render_attribute_name(hardcoded, &desc)}</td>
                 <td>{if attribute.is_list { format!("List<{attribute_type}>")} else {attribute_type.to_string()}}</td>
                 <td>{if attribute.is_editable {checkmark.clone()} else {html!{}}}</td>
                 <td>{if attribute.is_visible {checkmark.clone()} else {html!{}}}</td>
--- a/server/src/infra/graphql/mutation.rs
+++ b/server/src/infra/graphql/mutation.rs
@@ -23,7 +23,8 @@ use crate::{
 };
 use anyhow::{anyhow, Context as AnyhowContext};
 use base64::Engine;
-use juniper::{graphql_object, FieldResult, GraphQLInputObject, GraphQLObject};
+use juniper::{graphql_object, FieldError, FieldResult, GraphQLInputObject, GraphQLObject};
+use lldap_validation::attributes::{validate_attribute_name, ALLOWED_CHARACTERS_DESCRIPTION};
 use tracing::{debug, debug_span, Instrument, Span};
 
 #[derive(PartialEq, Eq, Debug)]
@@ -440,6 +441,22 @@ impl<Handler: BackendHandler> Mutation<Handler> {
         span.in_scope(|| {
             debug!(?name, ?attribute_type, is_list, is_visible, is_editable);
         });
+        validate_attribute_name(&name).map_err(|invalid_chars: Vec<char>| -> FieldError {
+            let chars = String::from_iter(invalid_chars);
+            span.in_scope(|| {
+                debug!(
+                    "Cannot create attribute with invalid name. Valid characters: {}. Invalid chars found: {}",
+                    ALLOWED_CHARACTERS_DESCRIPTION,
+                    chars
+                )
+            });
+            anyhow!(
+                "Cannot create attribute with invalid name. Valid characters: {}. Invalid chars found: {}",
+                ALLOWED_CHARACTERS_DESCRIPTION,
+                chars
+            )
+            .into()
+        })?;
         let handler = context
             .get_admin_handler()
             .ok_or_else(field_error_callback(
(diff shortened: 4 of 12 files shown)


app/src/components/group_schema_table.rs
```diff
@@ -154,7 +155,27 @@ impl GroupSchemaTable {
...
+                            Ok(()) => {
+                                html!{}
+                            }
+                            Err(_invalid_chars) => {
+                                html!{
+                                    <>
+                                        <br/>
+                                        <small class="text-warning">
+                                            {"Warning: This attribute uses one or more invalid characters. "}
+                                            {"Not globally supported."}
```
nitnelave, line 171: Maybe "Some clients may not support it"?
Also, let's put the list of allowed characters there.

app/src/components/user_schema_table.rs
```diff
@@ -153,7 +154,27 @@ impl UserSchemaTable {
...
+                        match validate_attribute_name(&attribute.name) {
+                            Ok(()) => {
+                                html!{}
+                            }
+                            Err(_invalid_chars) => {
+                                html!{
+                                    <>
+                                        <br/>
+                                        <small class="text-warning">
+                                            {"Warning: This attribute uses one or more invalid characters. "}
```
nitnelave, line 169: Can we refactor this code? If not a component, at least put the error message in a constant somewhere

server/src/infra/graphql/mutation.rs
```diff
@@ -440,6 +441,19 @@ impl<Handler: BackendHandler> Mutation<Handler> {
...
+            Ok(()) => {}
+            Err(invalid_chars) => {
+                let chars = String::from_iter(invalid_chars);
+                span.in_scope(|| {
+                    debug!(
+                        "Cannot create attribute with invalid name. Invalid chars found: {}",
+                        chars
+                    )
+                });
+                return Err("Cannot create attribute with invalid name".into());
```
nitnelave, line 454: Let's put the list of allowed characters there too. Ideally exposed by the common crate

validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+authors = ["Valentin Tolmer <valentin@tolmer.fr>"]
```
nitnelave, line 2: You can add yourself in there :)

validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[features]
+default = []
+js = []
+
+[dependencies.derive_more]
+features = ["debug", "display"]
+default-features = false
+version = "1"
+
+# For WASM targets, use the JS getrandom.
```
nitnelave, line 20: Is that necessary? Because it's a dependency somewhere? Can we disable that dependency via features maybe?

validation/src/attributes.rs
```diff
@@ -0,0 +1,43 @@
+pub fn validate_attribute_name(attribute_name: &str) -> Result<(), Vec<char>> {
+    let invalid_chars: Vec<char> = attribute_name
+        .chars()
+        .filter(|c| !c.is_alphanumeric())
```
nitnelave, line 4: Alphanumeric or dash (not underscore)

validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+authors = ["Valentin Tolmer <valentin@tolmer.fr>"]
+description = "Validation logic for LLDAP"
+edition = "2021"
+homepage = "https://github.com/lldap/lldap"
+license = "GPL-3.0-only"
+name = "lldap_validation"
+repository = "https://github.com/lldap/lldap"
+version = "0.6.0"
+
+[features]
```
nitnelave, line 11: Do we really need features for this crate?

nitnelave (commented):
I'd like to avoid populating the frontend with non-actionable warnings. We can either change the alias we return for hardcoded attributes (in a separate PR, if only to validate that the test changes are expected) or silence the warning for hardcoded attributes. Ideally the first one, and it'd be merged before this PR.

Otherwise, it looks good! Thanks a lot, and sorry for the slow review

app/src/components/create_group_attribute.rs
```diff
@@ -62,6 +63,16 @@ impl CommonComponent<CreateGroupAttributeForm> for CreateGroupAttributeForm {
                     bail!("Check the form for errors");
                 }
                 let model = self.form.model();
+                match validate_attribute_name(&model.attribute_name) {
```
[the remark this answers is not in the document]
broeng: I'll look it up, but that sounds good, if I'm guessing correctly. :)

app/src/components/group_schema_table.rs
broeng: Check. I'll see if I can find a good place for it.

server/src/infra/graphql/mutation.rs
broeng: Do you want the full list, or is some "a-zA-Z" abbreviation preferred?

validation/Cargo.toml
broeng: Will do - didn't want to intrude :)

validation/Cargo.toml
broeng: I'm assuming you mean the WASM stuff? Yeah, since it's also validating in the frontend. Any suggestions?

validation/src/attributes.rs
broeng: Ah, right, will get that fixed. Thanks.

broeng:
> I'd like to avoid populating the frontend with non-actionable warnings. We can either change the alias we return for hardcoded attributes (in a separate PR, if only to validate that the test changes are expected) or silence the warning for hardcoded attributes. Ideally the first one, and it'd be merged before this PR.

Changing the suggested list was also what I had in mind, just thought I'd do it after, so we could have the conversation with this as a backdrop.

I'm not sure there's aliases for all -- I'll check -- if not, isn't it okay it warns for a hardcoded attribute? Might save some debugging time knowing they may be troublesome. 

I'll try to push an update a bit later today.

Thanks for the review!

server/src/infra/graphql/mutation.rs
nitnelave: Whatever is most readable. I think regex is acceptable, though it might not be clear with the dash. Maybe plain English "alphanumeric (a-z, A-Z and 0-9) and dash (-)"

validation/Cargo.toml
nitnelave: This is only necessary if we depend on random, which we don't use here. Does it actually break if you remove it?

app/src/components/create_group_attribute.rs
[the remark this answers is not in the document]
broeng: Done

app/src/components/group_schema_table.rs
broeng: Done

app/src/components/user_schema_table.rs
broeng: Done.

server/src/infra/graphql/mutation.rs
broeng: Done.

validation/Cargo.toml
broeng: They could all be removed. Done.

validation/src/attributes.rs
broeng: Fixed.

broeng:
Comments addressed. Rebased with main as well.

server/src/infra/graphql/mutation.rs
```diff
@@ -440,6 +441,20 @@ impl<Handler: BackendHandler> Mutation<Handler> {
...
         });
+        validate_attribute_name(&name).map_err(|invalid_chars: Vec<char>| -> FieldError {
+            let chars = String::from_iter(invalid_chars);
+            span.in_scope(|| {
+                debug!(
+                    "Cannot create attribute with invalid name. Invalid chars found: {}",
+                    chars
+                )
+            });
+            anyhow!(
```
nitnelave, line 452: Let's report both the invalid characters found and the valid character list. Let's not make the user guess.

validation/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#![forbid(non_ascii_idents)]
+#![allow(clippy::nonstandard_macro_braces)]
+
+pub mod attributes;
```
nitnelave, line 4: I like that it doesn't have any dependencies right now :)

nitnelave (approved):

nitnelave (approved):

kosssi:
It's a problem with Vikunja to not authorize underscore. It's possible to configure it on Authentik and Keycloak. You can see [docs here](https://vikunja.io/docs/openid/#automatically-assign-users-to-teams)

Reference https://github.com/go-vikunja/vikunja/issues/1650

nitnelave:
@kosssi it seems you found a solution?

kosssi:
yes, sorry, it's not LLDAP problem finally, it's the OIDC to transform name group ;)

Issue #1077, opened 2025-01-16
[FEATURE REQUEST] Allow the user to specify the path to front-end assets in the config file


tyami94:
**Motivation**
This would help us package the software properly. I am currently packaging this for Alpine, and I have to place the assets in ```/var/lib/lldap/``` instead of the proper ```/usr/share/webapps/lldap/```. The PKGBUILD for Arch also mentions this being an issue.

**Describe the solution you'd like**
A config option and/or environment variable to override the hard-coded ```./app/``` paths for front-end assets.

**Describe alternatives you've considered**
Patching the source to replace the hard-coded paths in ```server/src/infra/tcp_server.rs``` before building, but this seems hacky and fragile. The AUR packaging just places it in ```/var/lib/lldap/app```.

nitnelave:
How big of an issue is it? Given that you are already writing files in that folder, why is it important to move the webapp files to another folder?

tyami94:
Storing the static assets in ```/var/lib``` violates the FHS, as that directory is intended for variable state information. ```/var/lib``` is intended more for things like sqlite databases or other "live" files. ```/usr/share``` is intended for architecture-independent static data, like documentation, static web assets, configuration examples, etc.

Additionally, on both Alpine and Arch, ```/var/lib/$pkgname```  is typically used as the home directory for the user that we run the service as.

An incomplete list of problems that the hard-coded paths cause are as follows (from least to most important):
- Savvy users and sysadmins are expecting static assets to be in ```/usr/share```, and this violates that expectation.
- Users cannot easily override the assets provided by the package manager if they wish only to use our binaries, but not our assets. The package manager will replace any changes they make to the files.
- Alternative front-ends cannot exist without conflicting with the default assets.
- Special consideration has to be taken for backups so the assets aren't needlessly backed up when we really just need state information, as well as so the known-good assets provided by the package manager aren't clobbered upon restoring a backup.
- If the service is somehow compromised, it's possible to maliciously edit, replace, or remove the assets because they are owned by ```lldap:lldap``` as is everything else in that directory. When they are in ```/usr/share/```, they are owned by ```root:root``` and cannot be modified in any way, even if the service itself is compromised. This eliminates a possible method of achieving persistence.

I understand that importance is relative and some of these things may not sound like a very big deal, but adding a knob for us to tweak here is a trivial way to resolve all of these problems, and will generally just make our lives easier as package maintainers. Additionally, this is relatively standard practice for other projects, such as the following few examples:
[...]

PR #1079, opened 2025-01-17
server: Allow custom path to front-end assets (#1077)

tyami94:
This commit allows the user to specify a custom path to the front-end assets with either the LLDAP_ASSETS_PATH environment variable or the assets_path config option. This should make lldap more pleasant for maintainers to package properly.

Addresses issue #1077


server/src/infra/configuration.rs
```diff
@@ -131,6 +131,8 @@ pub struct Configuration {
     pub ldaps_options: LdapsOptions,
     #[builder(default = r#"HttpUrl(Url::parse("http://localhost").unwrap())"#)]
     pub http_url: HttpUrl,
+    #[builder(default = r#"String::from("./app")"#)]
+    pub assets_path: String,
```
nitnelave, line 135: A few things here: 
- can you move it above the option groups? (Ldaps, SMTP, ...)
- can we parse it as a path instead of leaving it as a string? It's a bit too loose of a type for my taste. 
- maybe add a warning on startup if, let's say, index.html can't be found

server/src/infra/tcp_server.rs
```diff
@@ -25,7 +25,7 @@ use std::sync::RwLock;
 use tracing::info;
 
 async fn index<Backend>(data: web::Data<AppState<Backend>>) -> actix_web::Result<impl Responder> {
-    let mut file = std::fs::read_to_string(r"./app/index.html")?;
+    let mut file = std::fs::read_to_string(format!("{}/index.html", data.assets_path))?;
```
nitnelave, line 28: I'd like to use path methods rather than string concatenation here

lldap_config.docker_template.toml
```diff
@@ -33,6 +33,9 @@
 ## The public URL of the server, for password reset links.
 #http_url = "http://localhost"
 
+## The path to the front-end assets.
```
nitnelave, line 36: Mention that relative paths are interpreted from the working directory, maybe.

nitnelave (commented):
Thanks! It looks nice, a couple of comments but overall it's good

tyami94:
I think I've addressed your comments. Is this closer to what you wanted?

server/src/infra/configuration.rs
```diff
@@ -89,6 +90,8 @@ pub struct HttpUrl(pub Url);
 #[derive(Clone, Deserialize, Serialize, derive_builder::Builder, derive_more::Debug)]
 #[builder(pattern = "owned", build_fn(name = "private_build"))]
 pub struct Configuration {
+    #[builder(default = r#"PathBuf::from("./app")"#)]
+    pub assets_path: PathBuf,
```
nitnelave, line 94: Now that's a bit too high :D
Let's put it right after the key seed. It affects the order in which the debug info is printed, and I want to highlight the most important/common config values at the top. This is more niche.

nitnelave (approved):
Much better, thanks! One last nit and we're good to go.

I assume you tested it with a non-default value?

tyami94:
Yes, I tested it with a custom path as well as the default and it works as expected. Thanks!

tyami94:
Hi! Is there anything else I need to do to get this merged?

nitnelave:
Sorry, I meant to merge it earlier, let me try to do this tonight.

nitnelave:
Or, you know, right now

PR #1080, opened 2025-01-19
Update Authelia config example

fa-sharp:
This updates the Authelia config example to use the built-in `lldap` implementation. I tried to keep only the settings that are specific to LLDAP, and also clarify which settings are optional.

Note I haven't tried the TLS/LDAPS option myself, so I'm assuming that Authelia's TLS settings will work with LLDAP.


nitnelave (approved):
Nice! Thanks :)

PR #1081, opened 2025-01-20
Add support for querying GroupId with LDAP filters

broeng:
Adds support for querying groups by their Id's in LDAP search filters. 

Support is added for both group_id and groupid, the former being shown on the group schema overview, and the latter being supported by openldap's C library, and thus more clients.

### DIFF
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -1500,6 +1500,39 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_search_groups_by_groupid() {
+        let mut mock = MockTestBackendHandler::new();
+        mock.expect_list_groups()
+            .with(eq(Some(GroupRequestFilter::GroupId(GroupId(1)))))
+            .times(1)
+            .return_once(|_| {
+                Ok(vec![Group {
+                    id: GroupId(1),
+                    display_name: "group_1".into(),
+                    creation_date: chrono::Utc.timestamp_opt(42, 42).unwrap().naive_utc(),
+                    users: vec![UserId::new("bob"), UserId::new("john")],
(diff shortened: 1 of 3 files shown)


server/src/infra/ldap_handler.rs
```diff
@@ -1499,6 +1499,70 @@ mod tests {
...
+                    users: vec![UserId::new("bob"), UserId::new("john")],
+                    uuid: uuid!("04ac75e0-2900-3e21-926c-2f732c26b3fc"),
+                    attributes: Vec::new(),
+                }])
+            });
+        let mut ldap_handler = setup_bound_admin_handler(mock).await;
+        let request = make_group_search_request(
+            LdapFilter::Equality("groupid".to_string(), "1".to_string()),
+            vec![
+                "objectClass",
```
nitnelave, line 1522: Let's reduce the queried fields. Maybe just DN is enough

nitnelave (approved):

server/src/infra/ldap_handler.rs
broeng: Done.

nitnelave (approved):
Thanks!

nitnelave (commented):
Just realized: can we query groupId as an LDAP attribute? Not as a filter, but get the value.

broeng:
> Just realized: can we query groupId as an LDAP attribute? Not as a filter, but get the value.

Just tested. No, it's actually not included.

Should I make a followup PR with that?

nitnelave:
> > > Just realized: can we query groupId as an LDAP attribute? Not as a filter, but get the value.
> > 
> > Just tested. No, it's actually not included.
> > 
> > Should I make a followup PR with that, or add to this?
> 
> Scratch that, it **is** there, if the attribute is requested explicitly, just not by default.

Okay, that works for me.

PR #1082, opened 2025-01-20
Change default alias for User & Group schema attributes

broeng:
A number of the hardcoded attributes displayed in the User and Group schema overviews were using aliases which contain underscores, which is not always completely supported by clients. Therefore, this commit changes the primary alias for each attribute to be one without underscores.

To reduce confusion with this change, and also improve the visibility of available aliases, this commit also includes a list of each alias, for each hardcoded attribute. This list will also include the old primary aliases.

This PR has been marked as a draft, as it depends on somewhat on [PR #1081](https://github.com/lldap/lldap/pull/1081)

### DIFF
--- a/app/src/components/group_schema_table.rs
+++ b/app/src/components/group_schema_table.rs
@@ -1,10 +1,12 @@
 use crate::{
     components::{
         delete_group_attribute::DeleteGroupAttribute,
+        fragments::attribute_schema::render_attribute_name,
         router::{AppRoute, Link},
     },
     convert_attribute_type,
     infra::{
+        attributes::group,
         common_component::{CommonComponent, CommonComponentParts},
         schema::AttributeType,
     },
@@ -152,9 +154,10 @@ impl GroupSchemaTable {
         </svg>
                 };
         let hardcoded = ctx.props().hardcoded;
+        let desc = group::resolve_group_attribute_description_or_default(&attribute.name);
         html! {
             <tr key={attribute.name.clone()}>
-                <td>{&attribute.name}</td>
+                <td>{render_attribute_name(hardcoded, &attribute.name, &desc)}</td>
                 <td>{if attribute.is_list { format!("List<{attribute_type}>")} else {attribute_type.to_string()}}</td>
                 <td>{if attribute.is_visible {checkmark.clone()} else {html!{}}}</td>
                 {
--- a/app/src/infra/attributes.rs
+++ b/app/src/infra/attributes.rs
@@ -1,0 +1,109 @@
+pub struct AttributeDescription<'a> {
+    pub attribute_identifier: &'a str,
+    pub attribute_name: &'a str,
+    pub aliases: Vec<&'a str>,
+}
+
+pub mod group {
+
+    use super::AttributeDescription;
+
+    pub fn resolve_group_attribute_description(name: &str) -> Option<AttributeDescription> {
+        match name {
+            "creation_date" => Some(AttributeDescription {
+                attribute_identifier: name,
+                attribute_name: "creationdate",
+                aliases: vec![name, "createtimestamp", "modifytimestamp"],
+            }),
+            "display_name" => Some(AttributeDescription {
+                attribute_identifier: name,
+                attribute_name: "displayname",
+                aliases: vec![name, "cn", "uid", "id"],
+            }),
+            "group_id" => Some(AttributeDescription {
+                attribute_identifier: name,
+                attribute_name: "groupid",
+                aliases: vec![name],
+            }),
+            "uuid" => Some(AttributeDescription {
+                attribute_identifier: name,
+                attribute_name: name,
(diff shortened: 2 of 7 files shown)


app/src/components/group_schema_table.rs
```diff
@@ -154,7 +155,35 @@ impl GroupSchemaTable {
         let hardcoded = ctx.props().hardcoded;
         html! {
             <tr key={attribute.name.clone()}>
-                <td>{&attribute.name}</td>
+                {
+                    if hardcoded {
```
nitnelave, line 159: Nit: I usually try to keep the indentation at 2 spaces for html macros, since they get pretty deep. It's not enforced by cargo fmt, sadly.

app/src/components/group_schema_table.rs
```diff
@@ -154,7 +155,35 @@ impl GroupSchemaTable {
...
+                                {desc.attribute_name.clone()}
+                                {
+                                    if desc.aliases.is_empty() {
+                                        html!{}
+                                    } else {
+                                        html!{
+                                            <>
+                                                <br/>
+                                                <small class="text-muted">
+                                                    {if desc.aliases.len() == 1 {"Alias: "} else {"Aliases: "}}
```
nitnelave, line 172: Just "Aliases". Let's not add complexity for the sake of proper grammar

app/src/components/group_schema_table.rs
```diff
@@ -154,7 +155,35 @@ impl GroupSchemaTable {
         let hardcoded = ctx.props().hardcoded;
         html! {
             <tr key={attribute.name.clone()}>
-                <td>{&attribute.name}</td>
+                {
+                    if hardcoded {
+                        let desc = group::resolve_group_attribute_description_or_default(&attribute.name);
+                        html!{
+                            <td>
+                                {desc.attribute_name.clone()}
```
nitnelave, line 163: Do you need the clone? I would expect that a reference should be emough

app/src/infra/attributes.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub attribute_identifier: String,
+    pub attribute_name: String,
+    pub aliases: Vec<String>,
+}
+
+pub mod group {
+
+    use super::AttributeDescription;
+
+    pub fn resolve_group_attribute_description(name: &str) -> Option<AttributeDescription> {
```
nitnelave, line 11: I think we can have a single function for both user and group attributes together, there's no need for the split. And ideally also refactor the display code above for code sharing.

nitnelave (approved):

app/src/infra/attributes.rs
broeng: I kind of think that might be asking for trouble, displayName doesn't have the same aliases for users and groups, and one of those aliases is "id", which means something else for users. Personally, I think it'd be best to just keep them separate.

They still use the same struct, so I can easily make some shared display code anyway. 

By the way, I didn't do that, because I didn't see a good place for it, and thought it small enough to not matter too much. Any suggestions of where? Maybe a "fragments" folder under "components"?

app/src/components/group_schema_table.rs
broeng: Probably not, I'll fix it.

app/src/components/group_schema_table.rs
broeng: Check.

app/src/components/group_schema_table.rs
broeng: Check, I'll get it converted.

app/src/infra/attributes.rs
broeng: Pushed an update with shared display code (will probably update that a bit when revisiting attribute validation PR). Added a "fragments" folder for the shared display code.

For now, the user and group attributes are still split.

nitnelave: That makes sense, I didn't consider "displayName". If the names had been non-overlapping I would have pushed for it (YAGNI) but in this case you're right.

I guess "fragment" is as good a name as any, although it stinks of "utils". I don't have a better name.

broeng: Agree with the "utils"-stench. Should at least be fairly simple to rename, if we find some inspiration at some point.

I'll mark this as resolved then.

app/src/infra/attributes.rs
```diff
@@ -0,0 +1,122 @@
+pub struct AttributeDescription {
```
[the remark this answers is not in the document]
broeng: Sounds like a fine idea, and a occasion to get a bit familiar with lifetime parameters :)

I pushed an update with it now, is it what you were thinking?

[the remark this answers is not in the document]
nitnelave: Precisely! That's perfect :)

nitnelave (approved):

PR #1086, opened 2025-01-24
Move domain::types to separate lldap_domain module

broeng:
Preparation for using basic type definitions in other upcoming modules, in particular for plugins. By moving these fundamental domain types to a separate module, it is possible to reuse them in other modules, that the primary server module will need to depend on.

Specific use case in mind is the plugin functionality, where it would be very convenient to be able to share the same domain types with the backend API offered to plugins, so they get the same types for Users, Groups, Attributes, etc., as everything else, thus avoiding a need for duplicate types and constant conversion from and to.

Took the liberty to create a modules folder for these. I think the root of the project is starting to get a bit cluttered, and with plugins, for instance, it can easily get a bit complicated to find where actual bundled plugins are located, the plugin implementation, etc.


nitnelave:
I need to have a better look at it a bit later, but I'm all for separating in more modules. Btw, the rust name is crate ;)
Let's move lldap_auth to the crates folder as well, keep only the binaries at the top level. Maybe even rename app to frontend.

Another thing I wanted to split was putting the entire "domain" into its own crate, to enforce that it doesn't depend on anything infra.
We can even have a separate crate for LDAP and one for GraphQL

broeng:
Right, crates... ;)

Great, then I'll get it updated one of the next days, with rename and making sure it passes the checks, and anything you might have otherwise found before.

I've tried to just keep it to moving the existing types, and updating imports. So there's hopefully no surprises.

Was also thinking to move the other modules, like you mention, but I prefer to do it one at a time. So if you don't object to that, I'll do those right after :)

nitnelave:
Of course, one at a time. A small PR is a happy PR!

broeng:
Had to replace some `cfg(test)` directives to use a test-feature instead, as apparently that directive does not pass over between crates.

Also fixed a couple of old clippy issues in separate commits.

server/Cargo.toml
nitnelave, line 1: Can we remove some dependencies from here?

server/src/domain/handler.rs
```diff
@@ -1,12 +1,14 @@
...
-    error::Result,
-    types::{
-        AttributeName, AttributeType, AttributeValue, Email, Group, GroupDetails, GroupId,
-        GroupName, JpegPhoto, LdapObjectClass, Serialized, User, UserAndGroups, UserColumn, UserId,
-        Uuid,
-    },
-};
+use crate::domain::{error::Result, model::UserColumn};
 use async_trait::async_trait;
+use lldap_domain::requests::{
```
nitnelave, line 3: Group all the lines using lldap_domain

server/src/domain/model/group_attribute_schema.rs
```diff
@@ -1,10 +1,8 @@
 use sea_orm::entity::prelude::*;
 use serde::{Deserialize, Serialize};
 
-use crate::domain::{
-    handler::AttributeSchema,
-    types::{AttributeName, AttributeType},
-};
+use lldap_domain::schema::AttributeSchema;
```
nitnelave, line 4: Group

server/src/domain/model/user_attribute_schema.rs
```diff
@@ -1,10 +1,8 @@
 use sea_orm::entity::prelude::*;
 use serde::{Deserialize, Serialize};
 
-use crate::domain::{
-    handler::AttributeSchema,
-    types::{AttributeName, AttributeType},
-};
+use lldap_domain::schema::AttributeSchema;
```
nitnelave, line 4: Group

server/src/domain/sql_backend_handler.rs
```diff
@@ -23,15 +23,16 @@ pub mod tests {
...
+                GroupBackendHandler, UserBackendHandler, UserListerBackendHandler,
+                UserRequestFilter,
             },
             sql_tables::init_table,
-            types::{GroupId, UserId},
         },
         infra::configuration::ConfigurationBuilder,
     };
     use lldap_auth::{opaque, registration};
+    use lldap_domain::requests::{CreateGroupRequest, CreateUserRequest};
```
nitnelave, line 34: Group

server/src/domain/sql_group_backend_handler.rs
```diff
@@ -1,14 +1,14 @@
...
-        CreateGroupRequest, GroupBackendHandler, GroupListerBackendHandler, GroupRequestFilter,
-        UpdateGroupRequest,
-    },
+    handler::{GroupBackendHandler, GroupListerBackendHandler, GroupRequestFilter},
     model::{self, GroupColumn, MembershipColumn},
     sql_backend_handler::SqlBackendHandler,
-    types::{AttributeName, AttributeValue, Group, GroupDetails, GroupId, Serialized, Uuid},
 };
 use async_trait::async_trait;
+use lldap_domain::{
```
nitnelave, line 8: Yeah, just like that! 😁

server/src/domain/sql_user_backend_handler.rs
```diff
@@ -1,17 +1,15 @@
...
+    handler::{UserBackendHandler, UserListerBackendHandler, UserRequestFilter},
     model::{self, GroupColumn, UserColumn},
     sql_backend_handler::SqlBackendHandler,
-    types::{
-        AttributeName, AttributeValue, GroupDetails, GroupId, Serialized, User, UserAndGroups,
-        UserId, Uuid,
-    },
 };
 use async_trait::async_trait;
+use lldap_domain::requests::{CreateUserRequest, UpdateUserRequest};
```
nitnelave, line 8: Group

server/src/infra/access_control.rs
```diff
@@ -6,16 +6,20 @@ use tracing::info;
...
+        ReadSchemaBackendHandler, SchemaBackendHandler, UserBackendHandler,
+        UserListerBackendHandler, UserRequestFilter,
     },
     schema::PublicSchema,
-    types::{
-        AttributeName, Group, GroupDetails, GroupId, GroupName, LdapObjectClass, User,
-        UserAndGroups, UserId,
-    },
+};
+use lldap_domain::requests::{
```
nitnelave, line 15: Group

server/src/infra/auth_service.rs
```diff
@@ -21,14 +21,15 @@ use std::{
 use time::ext::NumericalDuration;
 use tracing::{debug, info, instrument, warn};
 
+use crate::domain::model::UserColumn;
```
nitnelave, line 24: Group with below

server/src/infra/graphql/api.rs
```diff
@@ -1,16 +1,15 @@
...
-    infra::{
-        access_control::{
-            AccessControlledBackendHandler, AdminBackendHandler, ReadonlyBackendHandler,
-            UserReadableBackendHandler, UserWriteableBackendHandler, ValidationResults,
-        },
-        auth_service::check_if_token_is_valid,
-        cli::ExportGraphQLSchemaOpts,
-        graphql::{mutation::Mutation, query::Query},
-        tcp_server::AppState,
+use crate::domain::handler::BackendHandler;
```
nitnelave, line 1: Group

server/src/infra/graphql/mutation.rs
```diff
@@ -24,6 +15,15 @@ use crate::{
 use anyhow::{anyhow, Context as AnyhowContext};
 use base64::Engine;
 use juniper::{graphql_object, FieldError, FieldResult, GraphQLInputObject, GraphQLObject};
+use lldap_domain::requests::{
+    CreateAttributeRequest, CreateGroupRequest, CreateUserRequest, UpdateGroupRequest,
```
nitnelave, line 19: Group

server/src/infra/ldap_handler.rs
```diff
@@ -28,6 +25,8 @@ use ldap3_proto::proto::{
     LdapResult as LdapResultOp, LdapResultCode, LdapSearchRequest, LdapSearchResultEntry,
     LdapSearchScope,
 };
+use lldap_domain::requests::CreateUserRequest;
```
nitnelave, line 28: Group

server/src/infra/test_utils.rs
```diff
@@ -1,4 +1,12 @@
-use crate::domain::{error::Result, handler::*, opaque_handler::*, types::*};
+use crate::domain::{error::Result, handler::*, opaque_handler::*};
+use lldap_domain::requests::{
```
nitnelave, line 2: Group

nitnelave (commented):
Looks good! Just a lot of the same comment on import style. I tried to flag all the instances so you wouldn't have to look for them.

crates/domain/Cargo.toml
```diff
@@ -0,0 +1,55 @@
+[package]
+name = "lldap_domain"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
broeng: Added.

server/Cargo.toml
broeng: Managed to find a couple, yes.

broeng:
All your comments should be addressed now, and the imports have been combined -- thanks for flagging them :-)

nitnelave (approved):

PR #1091, opened 2025-01-29
Clippy fixes

broeng:
Some legacy issues raised sporadically by clippy.


nitnelave (approved):
Thanks! There's probably a few more places where we can use that operator, but it's already cool

PR #1094, opened 2025-02-03
server: Adds support for whoamiOID

eyJhb:
Adds support for the OID whoami extension(?). I've tested this with `ldapwhoami -H ldap://127.0.0.1:3890 -x -D "uid=admin,ou=people,dc=example,dc=com" -w password`, and works without any issues.
This has also been tested in combination with Stalwart mail server :)

### DIFF
--- a/server/src/infra/ldap_handler.rs
+++ b/server/src/infra/ldap_handler.rs
@@ -23,7 +23,7 @@ use ldap3_proto::proto::{
     LdapDerefAliases, LdapExtendedRequest, LdapExtendedResponse, LdapFilter, LdapModify,
     LdapModifyRequest, LdapModifyType, LdapOp, LdapPartialAttribute, LdapPasswordModifyRequest,
     LdapResult as LdapResultOp, LdapResultCode, LdapSearchRequest, LdapSearchResultEntry,
-    LdapSearchScope,
+    LdapSearchScope, OID_PASSWORD_MODIFY, OID_WHOAMI,
 };
 use lldap_domain::{
     requests::CreateUserRequest,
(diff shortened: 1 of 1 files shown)


server/src/infra/ldap_handler.rs
```diff
@@ -204,6 +204,11 @@ fn root_dse_response(base_dn: &str) -> LdapOp {
                 atype: "isGlobalCatalogReady".to_string(),
                 vals: vec![b"false".to_vec()],
             },
+            LdapPartialAttribute {
+                atype: "supportedExtension".to_string(),
+                // whoami extension.
```
nitnelave, line 209: With the constant names I think we can get rid of both comments.

server/src/infra/ldap_handler.rs
```diff
@@ -413,16 +418,33 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
...
         }
+
+        if request.name == OID_WHOAMI {
+            let dn = self
+                .user_info
+                .as_ref()
+                .map(|user_info| {
+                    format!(
+                        "dn:uid={},ou=people,{}",
+                        user_info.user.clone().into_string(),
```
nitnelave, line 435: `user_info.user.as_str()` should work

server/src/infra/ldap_handler.rs
```diff
@@ -413,16 +418,33 @@ impl<Backend: BackendHandler + LoginHandler + OpaqueHandler> LdapHandler<Backend
 
     #[instrument(skip_all, level = "debug")]
     async fn do_extended_request(&mut self, request: &LdapExtendedRequest) -> Vec<LdapOp> {
-        match LdapPasswordModifyRequest::try_from(request) {
-            Ok(password_request) => self
+        if let Ok(password_request) = LdapPasswordModifyRequest::try_from(request) {
```
nitnelave, line 421: I think I understand better why there's no try_from for the who am I request: I think the correct pattern is to match on the name (OID), then try to parse the matching request. I understand if you don't want to do this refactoring (you already did more than I expected!) but I think the correct pattern to follow is:

```rust
match (request.name) {
  OID_PASSWORD_MODIFY => try to parse the request, and error otherwise,
  OID_WHOAMI => parsing is easy :D
  _ => unwilling to perform
}
```

nitnelave (approved):

nitnelave:
Closing in favor of https://github.com/lldap/lldap/pull/1098, I hope you don't mind.

I got a rare chance to work on LLDAP, I just wanted to clean this up.

PR #1096, opened 2025-02-03
Cleanup log messages for ignored attribute warnings

maystar:
When debugging lldap logs I recognized that the output looks strange using syslog and grafana.
There are two outputs for each warnings, the first one contains the visible `\n\` sequence.
Additional the user message contains the string ` group attribute` .

Be aware: I've no clue about rust development!


nitnelave (approved):
Sure 😁

PR #1099, opened 2025-02-04
server: remove deprecated fields from CreateUserRequest

broeng:
The fields first_name, last_name, and avatar have all been moved to regular attributes in the database, and are available through the GraphQL API as such as well. This commit removes the legacy fields for each on the internal CreateUserRequest type, leaving these to only be updateable through attributes.

The fields are still available in the GraphQL CreateUserInput type, preserving backwards compatiblity, and if set, they will be used for the corresponding attribute values. If both fields and attributes are set, the values given through attributes will superceed the fields, and be used. This change also fixes a bug, where creation of a user would fail if either of these attributes were set as both attribute and field, as it would attempt to insert the attribute twice, violating a unique constraint in the database.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -65,10 +65,19 @@ input CreateUserInput {
   id: String!
   email: String
   displayName: String
-  firstName: String
-  lastName: String
-  "Base64 encoded JpegPhoto." avatar: String
-  "User-defined attributes." attributes: [AttributeValueInput!]
+  """
+    First name of user. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ firstName: String
+  """
+    Last name of user. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ lastName: String
+  """
+    Base64 encoded JpegPhoto. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ avatar: String
+  "Attributes." attributes: [AttributeValueInput!]
 }
 
 type AttributeSchema {
@@ -86,9 +95,18 @@ input UpdateUserInput {
   id: String!
   email: String
   displayName: String
-  firstName: String
-  lastName: String
-  "Base64 encoded JpegPhoto." avatar: String
+  """
+    First name of user. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ firstName: String
+  """
+    Last name of user. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ lastName: String
+  """
+    Base64 encoded JpegPhoto. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ avatar: String
   """
     Attribute names to remove.
     They are processed before insertions.
--- a/server/src/domain/sql_backend_handler.rs
+++ b/server/src/domain/sql_backend_handler.rs
@@ -33,7 +33,7 @@ pub mod tests {
     use lldap_auth::{opaque, registration};
     use lldap_domain::{
         requests::{CreateGroupRequest, CreateUserRequest},
-        types::{GroupId, UserId},
+        types::{AttributeValue as DomainAttributeValue, GroupId, Serialized, UserId},
     };
     use pretty_assertions::assert_eq;
     use sea_orm::Database;
@@ -92,9 +92,16 @@ pub mod tests {
                 user_id: UserId::new(name),
                 email: format!("{}@bob.bob", name).into(),
                 display_name: Some("display ".to_string() + name),
-                first_name: Some("first ".to_string() + name),
-                last_name: Some("last ".to_string() + name),
-                ..Default::default()
+                attributes: vec![
+                    DomainAttributeValue {
+                        name: "first_name".into(),
+                        value: Serialized::from(("first ".to_string() + name).as_str()),
+                    },
+                    DomainAttributeValue {
+                        name: "last_name".into(),
+                        value: Serialized::from(("last ".to_string() + name).as_str()),
+                    },
+                ],
             })
             .await
             .unwrap();
--- a/server/src/domain/sql_user_backend_handler.rs
+++ b/server/src/domain/sql_user_backend_handler.rs
@@ -16,8 +16,8 @@ use sea_orm::{
     sea_query::{
         query::OnConflict, Alias, Cond, Expr, Func, IntoColumnRef, IntoCondition, SimpleExpr,
     },
-    ActiveModelTrait, ActiveValue, ColumnTrait, DatabaseTransaction, EntityTrait, IntoActiveValue,
-    ModelTrait, QueryFilter, QueryOrder, QuerySelect, QueryTrait, Set, TransactionTrait,
+    ActiveModelTrait, ActiveValue, ColumnTrait, DatabaseTransaction, EntityTrait, ModelTrait,
+    QueryFilter, QueryOrder, QuerySelect, QueryTrait, Set, TransactionTrait,
 };
 use std::collections::HashSet;
 use tracing::instrument;
@@ -185,11 +185,6 @@ impl SqlBackendHandler {
             display_name: to_value(&request.display_name),
             ..Default::default()
         };
-        let to_serialized_value = |s: &Option<String>| match s.as_ref().map(|s| s.as_str()) {
-            None => None,
-            Some("") => Some(ActiveValue::NotSet),
-            Some(s) => Some(ActiveValue::Set(Serialized::from(s))),
-        };
         let mut update_user_attributes = Vec::new();
         let mut remove_user_attributes = Vec::new();
         let mut process_serialized =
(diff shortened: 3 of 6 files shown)


server/src/domain/sql_user_backend_handler.rs
```diff
@@ -318,27 +318,6 @@ impl UserBackendHandler for SqlBackendHandler {
             ..Default::default()
         };
         let mut new_user_attributes = Vec::new();
-        if let Some(first_name) = request.first_name {
```
nitnelave, line 321: Nice 🙂

nitnelave (commented):
Looks good, but it's a bit too complex to review on my phone. I'll actually have to bring out the laptop for that.

server/src/domain/sql_backend_handler.rs
```diff
@@ -92,9 +94,16 @@ pub mod tests {
                 user_id: UserId::new(name),
                 email: format!("{}@bob.bob", name).into(),
                 display_name: Some("display ".to_string() + name),
-                first_name: Some("first ".to_string() + name),
-                last_name: Some("last ".to_string() + name),
-                ..Default::default()
+                attributes: vec![
+                    DomainAttributeValue {
+                        name: AttributeName::from("first_name"),
```
nitnelave, line 99: Why the different style to construct the attribute name? If .into() is enough, let's use that

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +163,48 @@ fn unpack_attributes(
     })
 }
 
+fn consolidate_attributes(
```
nitnelave, line 166: Can you add a comment describing the purpose of the function? It's clear in the PR context, but not necessarily in general. And maybe pull up the comment about the relative priority, both here and in the GraphQL schema (by adding it to the correct struct)

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +163,48 @@ fn unpack_attributes(
...
+    attributes: Vec<AttributeValue>,
+    first_name: Option<String>,
+    last_name: Option<String>,
+    avatar: Option<String>,
+) -> Vec<AttributeValue> {
+    // Prepare map with the attributes from fields, which will have lower
+    // priority than a value from a corresponding attribute already given.
+    let mut fallback_attrs: BTreeMap<String, AttributeValue> = BTreeMap::new();
+    if let Some(first_name) = first_name {
```
nitnelave, line 175: It feels like you could have a for loop over `(first_name, "first_name"), (last_name, "last_name"), ...`

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +163,48 @@ fn unpack_attributes(
...
+    }
+    if let Some(avatar) = avatar {
+        let attr_value = AttributeValue {
+            name: "avatar".into(),
+            value: vec![avatar],
+        };
+        fallback_attrs.insert(attr_value.name.clone(), attr_value);
+    }
+    // Prepare map of the client provided attributes
+    let mut provided_attributes: BTreeMap<String, AttributeValue> = attributes
```
nitnelave, line 197: Maybe simpler: start with this map, and   insert the value of it's not already there in the loop. The logic will be clearer as well

server/src/infra/graphql/mutation.rs
```diff
@@ -962,4 +999,104 @@ mod tests {
...
+                    value: vec!["expected-last".to_string()],
+                },
+            ]
+        );
+    }
+
+    #[tokio::test]
+    async fn test_attribute_consolidation_field_fallback_2() {
+        let attributes = vec![AttributeValue {
+            name: "first_name".to_string(),
```
nitnelave, line 1075: Let's vary the casing of the attributes as well, to make sure we catch that case (which I don't think we do at the moment)

nitnelave (commented):
LGTM! I'm not approving yet because of the casing issue, but the design is good. Thanks!

server/src/infra/graphql/mutation.rs
broeng: Good point. Was assuming they had been normalized earlier.

server/src/domain/sql_backend_handler.rs
broeng: Yeah, that's a bit weird, into() it is.

server/src/infra/graphql/mutation.rs
broeng: Added a comment, and to the fields in the GraphQL schema.

server/src/infra/graphql/mutation.rs
broeng: Changed to that.

server/src/infra/graphql/mutation.rs
broeng: Done.

server/src/infra/graphql/mutation.rs
broeng: Fixed now. Also lowercases the attribute names, as we discussed on discord.

broeng:
I realized that I hadn't done the same for user updates -- that should be included now. It's pretty much the same, though with one addition; setting fields to the empty string would unset that field before, I thought it might be best to preserve that behaviour, and also enforce it for attributes, so after consolidation, the result is inspected and partitioned into either a list of attributes to delete (extra) or insert. Might be worth giving that a good look over. :-)

schema.graphql
```diff
@@ -65,9 +65,18 @@ input CreateUserInput {
...
+  """ firstName: String
+  """
+    Last name of user. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ lastName: String
+  """
+    Base64 encoded JpegPhoto. Deprecated: use attribute instead.
+    If both field and corresponding attribute is supplied, the attribute will take precedence.
+  """ avatar: String
   "User-defined attributes." attributes: [AttributeValueInput!]
```
nitnelave, line 80: It's not really just user-defined anymore. Can you update the comment?

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +172,52 @@ fn unpack_attributes(
...
+                x.name.clone().into(),
+                AttributeValue {
+                    name: x.name.clone().to_lowercase(),
+                    value: x.value,
+                },
+            )
+        })
+        .collect::<BTreeMap<_, _>>();
+    // Prepare list of fallback attribute values
+    let field_attrs: Vec<(&str, Option<String>)> = vec![
```
nitnelave, line 201: I don't think it needs to be a vec, just an array/slice

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +172,52 @@ fn unpack_attributes(
...
+        ("first_name", first_name),
+        ("last_name", last_name),
+        ("avatar", avatar),
+    ];
+    for a in field_attrs.iter() {
```
nitnelave, line 206: I think with ` into_iter` we can move the values out of the array, and then we avoid another copy of the attribute value (the `val.to_owned()`)

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +172,52 @@ fn unpack_attributes(
...
+    avatar: Option<String>,
+) -> Vec<AttributeValue> {
+    // Prepare map of the client provided attributes
+    let mut provided_attributes: BTreeMap<AttributeName, AttributeValue> = attributes
+        .into_iter()
+        .map(|x| {
+            (
+                x.name.clone().into(),
+                AttributeValue {
+                    name: x.name.clone().to_lowercase(),
```
nitnelave, line 194: `x.name.to_ascii_lowercase()` should be enough (avoiding an extra copy)

nitnelave (approved):

schema.graphql
broeng: Yes, certainly.

server/src/infra/graphql/mutation.rs
broeng: Check.

server/src/infra/graphql/mutation.rs
broeng: Hm, not sure I can do that. `val` comes from the `if let Some(val) ...`, and it won't let me move that out of the `AttributeValue` anyway.

server/src/infra/graphql/mutation.rs
broeng: Done

broeng:
Ah, sorry, the amend+push habit is strong.

server/src/infra/graphql/mutation.rs
nitnelave: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a41b6de40cb81ccddd0695a9db02cf35

nitnelave: Note: the field is an array, not a slice
Into_iter
The pattern matching doesn't borrow

broeng: Ah, right, didn't know I could unpack it in the for loop.

server/src/infra/graphql/mutation.rs
```diff
@@ -163,6 +172,52 @@ fn unpack_attributes(
...
+    avatar: Option<String>,
+) -> Vec<AttributeValue> {
+    // Prepare map of the client provided attributes
+    let mut provided_attributes: BTreeMap<AttributeName, AttributeValue> = attributes
+        .into_iter()
+        .map(|x| {
+            (
+                x.name.clone().into(),
+                AttributeValue {
+                    name: x.name.clone().to_ascii_lowercase(),
```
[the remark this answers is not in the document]
broeng: No, not needed. Removed.

server/src/infra/graphql/mutation.rs
broeng: Should be good now.

broeng:
I'd like to do the squash before merge, like we talked about. :)

nitnelave (approved):

nitnelave:
Sorry, not sure if you got notified. Go ahead. In the future, when I approve the changes, feel free to squash

broeng:
> Sorry, not sure if you got notified. Go ahead. In the future, when I approve the changes, feel free to squash

Sounds good, and done!

PR #1105, opened 2025-02-15
Add Project Quay example config

geragio:
This PR introduces a basic LDAP configuration for Project Quay deployments.

### DIFF
--- a/example_configs/quay.md
+++ b/example_configs/quay.md
@@ -1,0 +1,26 @@
+# Configuration for Quay
+
+Quay is a container image registry that enables you to build, organize, distribute, and deploy containers.
+
+The official LDAP configuration documentation is located [here](https://docs.projectquay.io/config_quay.html#config-fields-ldap).
+
+For standalone deployments of Project Quay, the core configuration is primarily set through the `config.yaml` file.
+
+If you install Project Quay on OpenShift Container Platform / OKD using the Project Quay Operator, the configuration resides in the `config bundle secret`. 
+
+This example assists with a basic LDAP configuration.
+
+```yaml
+AUTHENTICATION_TYPE: LDAP
+LDAP_ADMIN_DN: cn=admin,ou=people,dc=example,dc=com
+LDAP_ADMIN_PASSWD: password
+LDAP_ALLOW_INSECURE_FALLBACK: false
+LDAP_BASE_DN:
+    - dc=example
+    - dc=com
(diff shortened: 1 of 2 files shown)


example_configs/quay.md
```diff
@@ -0,0 +1,26 @@
...
+
+For standalone deployments of Project Quay, the core configuration is primarily set through the `config.yaml` file.
+
+If you install Project Quay on OpenShift Container Platform / OKD using the Project Quay Operator, the configuration resides in the `config bundle secret`. 
+
+This example assists with a basic LDAP configuration.
+
+```yaml
+AUTHENTICATION_TYPE: LDAP
+LDAP_ADMIN_DN: cn=admin,ou=people,dc=example,dc=com
```
nitnelave, line 15: What role do you need for the bind user? Admin, read-only, allow password changes?

example_configs/quay.md
```diff
@@ -0,0 +1,26 @@
...
+LDAP_ADMIN_PASSWD: password
+LDAP_ALLOW_INSECURE_FALLBACK: false
```
nitnelave, line 17: Is that about LDAPS?

nitnelave (commented):
Thanks! Can you also add it in the readme?

example_configs/quay.md
geragio: The bind user must be able to login and view the records for all user accounts

example_configs/quay.md
geragio: Quay supports both LDAP and LDAPS, the protocol should be defined inside `LDAP_URI` config parameter:
```
# LDAP example 
LDAP_URI: ldap://<example_url>
# LDAPS example
LDAP_URI: ldaps://<example_url>
```

The `LDAP_ALLOW_INSECURE_FALLBACK` parameter can be used to allow/disallow SSL/TLS insecure fallback for LDAP authentication.

example_configs/quay.md
nitnelave: Sounds good! Can you use a different bind user instead of admin, and document that it should be in `lldap_atrict_readonly`?

geragio: I modified the example configuration, used a different bind user and mentioned the user should be part of `lldap_strict_readonly` group.

nitnelave (approved):

nitnelave:
Thanks!

PR #1109, opened 2025-02-20
Create stalwart.md example_config

xeoneox:
Failed at getting Docker Mailserver and Maddy working, so I wrote a config for Stalwart Mailserver instead

### DIFF
--- a/example_configs/stalwart.md
+++ b/example_configs/stalwart.md
@@ -6,6 +6,9 @@ To integrate with LLDAP, ensure you correctly add these ldap setting to your `co
 
 ## Config.toml File Sample - (only the ldap portion)
 ```toml
+[storage]
+  directory = "ldap"
+
 [directory]
   [directory.ldap]
     base-dn = "dc=example,dc=org"
--- a/example_configs/stalwart.md
+++ b/example_configs/stalwart.md
@@ -2,7 +2,13 @@
 
 [Stalwart-mailserver](https://github.com/stalwartlabs/mail-server) is a Production-ready full-stack but simple mail server (SMTP, IMAP, LDAP, Antispam, Antivirus, etc.) written in Rust.
 
-To integrate with LLDAP, ensure you correctly add these ldap setting to your `config.toml`.
+To integrate with LLDAP, 
+
+1. Add manager user, & make sure to add to lldap_admin group for read/write permission
+   
+2. Create mail group, add users requiring email access 
+
+3. Ensure you correctly add the following ldap settings to your Stalwart `config.toml`.
 
 ## Config.toml File Sample - (only the ldap portion)
 ```toml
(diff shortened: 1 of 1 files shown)


example_configs/stalwart.md
```diff
@@ -0,0 +1,44 @@
...
+        0 = "displayName"
+    [directory.ldap.bind]
+      dn = "uid=admin,ou=people,dc=example,dc=org"
+      secret = "<YOUR_SECRET>"
+      [directory.ldap.bind.auth]
+        dn = "uid=?,ou=people,dc=example,dc=org"
+        enable = true
+        search = true
+      [directory.ldap.bind.filter]
+        email = "(&(|(objectClass=person)((memberof=cn=mail,ou=groups,dc=example,dc=org))(|(mail=?)(mailAlias=?)(mailList=?)))"
```
nitnelave, line 34: Ah, thanks, we finally have someone contributing this config! It was much requested.
A regular piece of feedback/bug report is that having a list attribute doesn't work if it contains more than one value (for mailAlias or mailList).
The recommended workaround (until we fix the bug, but it's a big design change so it'll take a while) is to have `mailAlias1`, `mailAlias2`, ...

example_configs/stalwart.md
```diff
@@ -0,0 +1,44 @@
...
+    base-dn = "dc=example,dc=org"
+    timeout = "30s"
+    type = "ldap"
+    url = "ldap://ldap.domain.example.org:3890"
+    [directory.ldap.attributes]
+      class = "objectClass"
+      email = "mail"
+      groups = "member"
+      name = "uid"
+      secret = "userPassword"
```
nitnelave, line 23: This value, IIRC, is not used at all (but needs to be specified, and maybe the attribute needs to exist). I would rather name it "dummyStalwartSecret" rather than userPassword, because that's what it is.

example_configs/stalwart.md
```diff
@@ -0,0 +1,44 @@
...
+      [directory.ldap.attributes.description]
+        0 = "displayName"
+    [directory.ldap.bind]
+      dn = "uid=admin,ou=people,dc=example,dc=org"
```
nitnelave, line 27: What permission does the bind user require? Read only, change passwords, create/modify users?

Usually we recommend against binding directly as the admin, principle of least privilege.

example_configs/stalwart.md
```diff
@@ -0,0 +1,44 @@
+# Stalwart Mailserver
+
+[Stalwart-mailserver](https://github.com/stalwartlabs/mail-server) is a Production-ready full-stack but simple mail server (SMTP, IMAP, LDAP, Antispam, Antivirus, etc.) written in Rust.
+
+To integrate with LLDAP, ensure you correctly add these ldap setting to your `config.toml`.
+
+## Config.toml File Sample - (only the ldap portion)
```
nitnelave, line 7: I think there are some setup steps that need to be documented on the LLDAP side:
- creating the attributes (which ones?)
- creating the mail group

nitnelave (changes requested):
Thanks a lot! People have been asking for this config a few times already :)
It's nice to finally have it.

example_configs/stalwart.md
xeoneox: will add instructions for a mail group

example_configs/stalwart.md
xeoneox: I think I'll remove alias for now and test. I was adapting someone else's attempt at Stalwart auth

example_configs/stalwart.md
xeoneox: I'm going to wait on this one until I confirm the other fixes

xeoneox: appears to be working

xeoneox:
changes complete

nitnelave (approved):

PR #1114, opened 2025-02-21
Update stalwart.md instructions w/ read-only bind

xeoneox:
Update instructions for read_only bind


nitnelave (approved):
Thanks for trying it!

PR #1115, opened 2025-02-21
domain: move Permission and ValidationResults to domain crate

broeng:
Moves the Permission enum and ValidationResults struct to the domain crate.

### DIFF
--- a/crates/domain/src/access_control.rs
+++ b/crates/domain/src/access_control.rs
@@ -1,0 +1,57 @@
+use lldap_auth::types::UserId;
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Permission {
+    Admin,
+    PasswordManager,
+    Readonly,
+    Regular,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ValidationResults {
+    pub user: UserId,
+    pub permission: Permission,
+}
+
+impl ValidationResults {
+    #[cfg(feature = "test")]
+    pub fn admin() -> Self {
+        Self {
+            user: UserId::new("admin"),
+            permission: Permission::Admin,
+        }
+    }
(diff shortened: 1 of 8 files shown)


nitnelave:
Where will you need them? In the context of the plugins?

nitnelave:
Would it make sense to move them to the auth crate instead? There's the user id in there, so we have to see what we can do, but it feels like the correct place.

broeng:
> Where will you need them? In the context of the plugins?

Yeah, for one thing in the LLDAP->plugin API, but was also thinking of just using the types when integrating with Lua, by using the serde serialization.

> Would it make sense to move them to the auth crate instead? There's the user id in there, so we have to see what we can do, but it feels like the correct place.

That would probably be more fitting, yes. Maybe I should just move the UserId to the auth crate as well?

nitnelave:
Thinking a bit more about it, having your concept of user id come from the auth layer... actually makes sense 😁

broeng:
> Thinking a bit more about it, having your concept of user id come from the auth layer... actually makes sense 😁

Yeah, agreed 😄

nitnelave:
It might even make sense to extract the case insensitive strings to a standalone crate out of this repository. That's something I can do later on.

broeng:
> It might even make sense to extract the case insensitive strings to a standalone crate out of this repository. That's something I can do later on.

Sounds good. I'll just move the current one to auth as well. I'll probably push an update tonight. Let me know if you end up making that crate before, I can do the switch then.

broeng:
> Thinking a bit more about it, having your concept of user id come from the auth layer... actually makes sense 😁

It makes so much sense, that it's in fact already there... :joy:

broeng:
Should be ready for another round.

nitnelave (approved):

PR #1119, opened 2025-02-28
DRAFT: Initial plugin support

broeng:
Initial work on providing support for plugins written in Lua for LLDAP.

Feel free to leave comments or suggestions. Would appreciate if any bigger structural refactorings could be done before I dig deeper in the current direction. I would suggest going on a per-commit basis when reviewing.

State:
Currently in a working state, but still very much a work in progress. Plugins can be included in the configuration, they will be initialized on start up, and all implemented listeners should be functional at this point.

Non-exhaustive TODO list:

- [x] Implement event handlers for get_users, get_groups, and other remaining BackendHandler methods.
- [ ] Handling of failing plugin handlers. Currently fails silently. Configurable block-on-failure/on-error-resume-next?
- [x] Change filters for list_users/list_groups to an AST instead of LDAP filter string.
- [x] Figure out if plugins should be triggered by actions caused by plugins.
- [ ] Support some kind of plugin api-version check.
- [x] Make chosen stdlib level explicit in Lua environment
- [x] Propagate a Result from Lua environment setup
- [x] Clean up sorting of registered event handlers a bit.
- [x] Finalize configuration format
- [x] Tests, many more tests...


server/src/domain/sql_backend_handler.rs
```diff
@@ -18,6 +18,12 @@ impl SqlBackendHandler {
 #[async_trait]
 impl BackendHandler for SqlBackendHandler {}
 
+impl WithContextHandler for SqlBackendHandler {
+    fn with_context(&self, _context: RequestContext) -> Self {
+        self.clone()
```
nitnelave, line 23: Note on cloning the handler: the DB connection is just a handle so can be cloned cheaply. It also takes a `Config`, which I guess is acceptable to clone (at least for now), but could be refactored as an `Arc` instead, or just the necessary fields extracted.

crates/domain-handlers/src/handler.rs
```diff
@@ -177,6 +196,10 @@ pub trait SchemaBackendHandler: ReadSchemaBackendHandler {
     async fn delete_group_object_class(&self, name: &LdapObjectClass) -> Result<()>;
 }
 
+pub trait WithContextHandler: Sized {
+    fn with_context(&self, context: RequestContext) -> Self;
+}
+
```
nitnelave, line 202: I'm not sure I'm a fan of this API. I see several problems with it:
- We're cloning the handler everywhere, potentially several times per query (getting a permission-restricted handler is not a zero-cost abstraction anymore).
- We can forget to initialize the context somewhere, if we never call `with_context`, since it returns the same type.

I was more thinking along the lines of:
```rust
struct PluginData{ ... }

struct PluginHandler <'a, BackendHandler> {
  plugin_data: &'a PluginData,
  backend_handler: &'a BackendHandler,
  request_context: RequestContext,
}

impl Handler for PluginHandler {
  // Forward the methods to the backend, and call the plugins
}
```

(Note that if we switch to the 2024 Rust edition first, I think we won't need the `'a` annotations)

server/src/infra/access_control.rs
```diff
@@ -183,41 +191,53 @@ impl<Handler: BackendHandler> AccessControlledBackendHandler<Handler> {
     pub fn get_admin_handler(
         &self,
         validation_result: &ValidationResults,
-    ) -> Option<&impl AdminBackendHandler> {
-        validation_result.is_admin().then_some(&self.handler)
```
nitnelave, line 187: Specifically, this line used to be a zero-cost forwarding of a reference, and now it comes with a clone.

crates/plugin-engine/plugins
broeng, line 1: Deleted. Don't know what that was about.

nitnelave (changes requested):
Reviewing one commit at a time, here's the review for the `WithContext` trait (which I think shouldn't be a trait)

crates/key-value-store/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "lldap_key_value_store"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+seaorm = ["dep:sea-orm"]
+
+[dependencies]
+async-trait = "0.1"
+serde = "*"
```
nitnelave, line 13: Major version

crates/key-value-store/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+thiserror = "1"
+
+[dependencies.tokio]
+version = "*"
```
nitnelave, line 17: Major version

crates/key-value-store/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+authors = ["Simon Broeng Jensen <sbj@cwconsult.dk>"]
+description = "Key-Value store for LLDAP"
+name = "lldap_key_value_store"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+seaorm = ["dep:sea-orm"]
```
nitnelave, line 9: This is, I guess, to be able to thread the error to the frontend without depending on sea-orm?

crates/key-value-store/src/api/store.rs
```diff
@@ -0,0 +1,29 @@
+use async_trait::async_trait;
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::api::error::KeyValueError;
+
+#[async_trait]
+pub trait KeyValueStore<K: Clone + Sync + Send>: Clone + Sync + Send {
+    //
```
nitnelave, line 8: Nit on style: remove the empty comment lines

crates/key-value-store/src/api/store.rs
```diff
@@ -0,0 +1,29 @@
+use async_trait::async_trait;
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::api::error::KeyValueError;
+
+#[async_trait]
+pub trait KeyValueStore<K: Clone + Sync + Send>: Clone + Sync + Send {
```
nitnelave, line 7: Why does it have to be Clone?

crates/plugin-kv-store/src/test.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    impl Drop for DbConn {
+        fn drop(&mut self) {
+            let filename: String = "kvtests".to_owned() + self.fileid + ".db";
+            let _ = std::fs::remove_file(filename);
+        }
+    }
+
+    async fn prepare_dbconn(id: &'static str) -> Result<DatabaseConnection, DbErr> {
+        let database_url: String = "sqlite://kvtests".to_owned() + id + ".db?mode=rwc";
```
nitnelave, line 24: You can have an in-memory sqlite DB.

crates/domain-handlers/src/handler.rs
broeng: I agree with the downsides, but I don't entirely understand how your proposal would work -- or what you're exactly proposing.

The proposed PluginHandler looks a lot like the PluginBackendHandler I've created in one of the later commits from here. My problem is, that the GraphQL layer, and LDAP layer for that matter, are only aware of the BackendHandler type, and none of the current traits comprising the BackendHandler type offers a good way to pass this information from those layers.

I kind of think it's correct, that those two layers only know of the BackendHandler trait, but that leaves me with two options, as I see it, for passing this request-context information into the BackendHandler -- which would specifically be a PluginBackendHandler, that triggers plugin, and forwards calls to SqlBackendHandler:

1. Create a WithContext-like trait like this, that provides a cloned version with the new information adopted.
2. Augment all method signatures of the traits BackendHandler is composed of, to include a RequestContext. I think that's a bit of a shame, to be honest.

I'm open to other suggestions for handling this, but I think those were the options we surfaced when we discussed it on discord.

crates/key-value-store/Cargo.toml
broeng: It was to get around, that I cant define From<> implementations in another crate than the one that defines the type, and I though it would be nice with some From-impls for sea-orms TransactionError and DbErr, to the error enum. So the are covered behind that feture.

crates/plugin-kv-store/src/test.rs
broeng: Oh, I wasn't aware of that. That also helps me with some tests, where I've created an InMemoryKeyValueStore :joy:

crates/key-value-store/src/api/store.rs
broeng: I'm not sure it does, I need to have a pass over some of these requirements -- not just here. I've had some "fun" with passing this type around the plugin engine with async, and to Lua, so there's been some iterations.

crates/plugin-engine/src/api/types.rs
```diff
@@ -0,0 +1,85 @@
...
+            PluginSource::ScriptSource(s) => {
+                write!(f, "<script: '{} [...]'>", &s.as_str()[0..30])
+            }
+        }
+    }
+}
+
+pub struct PluginConfig {
+    pub plugin_source: PluginSource,
+    pub kvscope: Option<String>,
```
nitnelave, line 54: Why optional? What does it represent?
I thought the scope was (basically) the plugin name, to partition the DB table?

crates/plugin-engine/src/internal/lualib/init.rs
```diff
@@ -0,0 +1,23 @@
...
+// TODO: return and propagate a Result type
+pub fn new_lua_environment() -> &'static Lua {
+    let lua = Box::leak(Box::new(Lua::new()));
+
+    let lldap_lib = lua.create_table().unwrap();
+    let _ = lldap_lib.set("encoding", LuaEncodingLib {});
+    let _ = lldap_lib.set("hashing", LuaHashingLib {});
+    let _ = lldap_lib.set("log", LuaLogger {});
+    let _ = lldap_lib.set("strings", LuaStringsLib {});
+    let _ = lldap_lib.set("tables", LuaTablesLib {});
```
nitnelave, line 17: This is where we'll add functionality for calling Http endpoints, I guess?

crates/plugin-engine/src/internal/lualib/tables.rs
```diff
@@ -0,0 +1,67 @@
...
+                } else {
+                    if value.equals(&bval)? {
+                        elem_result = true
+                    }
+                }
+            }
+        }
+        result = result && elem_result;
+    }
+    // meh
```
nitnelave, line 65: :D

crates/plugin-kv-store/src/store.rs
```diff
@@ -0,0 +1,137 @@
...
+                            .await?;
+                    // prepare model
+                    let model = model::plugin_key_values::ActiveModel {
+                        scope: ActiveValue::Set(key_scope.clone()),
+                        key: ActiveValue::Set(key.key.clone()),
+                        value: ActiveValue::Set(serialized_value),
+                        ..Default::default()
+                    };
+                    if existing.is_some() {
+                        model.update(transaction).await?;
```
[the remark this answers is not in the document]
broeng: Upsert is broadly supported now? How time flies.. :)

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+use super::arguments::ldap_bind_result::BindResult;
+
+#[async_trait]
+pub trait PluginHandlerEvents {
+    async fn initialize_plugins<A: BackendAPI>(
+        &self,
+        context: PluginContext<A>,
+    ) -> Result<(), String>;
+
+    async fn on_create_user<A: BackendAPI>(
```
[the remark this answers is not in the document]
broeng: Yeah, I also have a general point on my todo list on this draft about the error handling. I generally agree these options should be possible, but I just haven't really gotten around to it yet.

I general, these handlers that return a value, are basically doing a fold over the plugin handlers with the event, and the result should be used afterwards. I would also like the plugins to be able to abort the request, with for instance a permission denied reason, but I'm also considering if it should be possible to declare for a handler that failure during a plugin should either abort the request, or just be ignored, so it ties into that a bit.

It takes the request by value, because it will replace it, and the result should be the one going forward afterwards. I agree with optimize for no-op call, and I did actually return the original, but I can see i've forgotten that, when I moved the check into the exec_mutation_handler function. That should be fixed of course.

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+        backend::BackendAPI,
+        types::{PluginConfig, PluginContext},
+    },
+    internal::{plugins, types::plugins::PluginRegistry},
+};
+
+use super::arguments::ldap_bind_result::BindResult;
+
+#[async_trait]
+pub trait PluginHandlerEvents {
```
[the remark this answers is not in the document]
broeng: Yeah, plan to move it back. Was struggling with some lifetime stuff when there, but I think I figured that out with the KVStore.

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+use crate::{
+    api::{
+        arguments::ldap_search_result::SearchResult,
+        backend::BackendAPI,
+        types::{PluginConfig, PluginContext},
+    },
+    internal::{plugins, types::plugins::PluginRegistry},
+};
+
+use super::arguments::ldap_bind_result::BindResult;
```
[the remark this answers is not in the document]
broeng: Agreed. Generally replace them when I notice them (auto imports...)

crates/plugin-engine/src/api/types.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::api::backend::BackendAPI;
+use lldap_auth::access_control::ValidationResults;
+
+pub enum PluginSource {
+    ScriptFile(PathBuf),
+    ScriptSource(String),
+}
+
+impl PluginSource {
+    pub fn from_path(path: PathBuf) -> Result<Self, String> {
```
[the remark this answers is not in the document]
broeng: Yeah. Was still debating with myself if probing for what is pointed at should be inside or outside.

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+        search_request: LdapSearchRequest,
+        search_result: SearchResult,
+    ) -> Result<SearchResult, String>;
+
+    async fn on_ldap_root_dse<A: BackendAPI>(
+        &self,
+        context: PluginContext<A>,
+        search_result_entry: LdapSearchResultEntry,
+    ) -> Result<LdapSearchResultEntry, String>;
+}
```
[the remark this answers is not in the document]
broeng: That should already be covered by the other handlers. I have generally had that use case in mind. Though currently it's done by letting LLDAP do what it would do, and then change the result, for unsupported functionality that would usually be an error message that's replaced.

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+        group_id: GroupId,
+    ) -> Result<GroupId, String>;
+
+    async fn on_deleted_group<A: BackendAPI>(
+        &self,
+        context: PluginContext<A>,
+        group_id: GroupId,
+    ) -> Result<(), String>;
+
+    async fn on_added_user_to_group<A: BackendAPI>(
```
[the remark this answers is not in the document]
broeng: Yeah, missing quite a few actually, just didn't want to continue with those until you had a first look. :)

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+        context: PluginContext<A>,
+        bind_request: LdapBindRequest,
+        bind_result: BindResult,
+    ) -> Result<BindResult, String>;
+
+    async fn on_ldap_unbind<A: BackendAPI>(
+        &self,
+        context: PluginContext<A>,
+        user_id: UserId,
+    ) -> Result<(), String>;
```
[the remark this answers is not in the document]
broeng: They trigger after, allowing for overriding the result, and doing something else if the operation wasn't supported. Do plan to document these things though.

Forgot about the password here, actually, but I was thinking it would be nice, if different auth mechanisms could be implemented with plugins. Maybe we can circle back to this one, after talking about on_ldap_password_update :)

crates/plugin-engine/src/api/handler.rs
```diff
@@ -0,0 +1,184 @@
...
+        context: PluginContext<A>,
+        request: LdapExtendedRequest,
+        result: Vec<LdapOp>,
+    ) -> Result<Vec<LdapOp>, String>;
+
+    async fn on_ldap_password_update<A: BackendAPI>(
+        &self,
+        context: PluginContext<A>,
+        user_id: UserId,
+        password: String,
```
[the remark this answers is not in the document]
broeng: Absolutely not surprised by your comment :-)

And yeah, I would really like this functionality, and honestly also the cleartext version. I'd really like to be able to generate the NTLM hash for Samba, when the password is updated through this, and sadly, thats done by converting the password to UTF16-LE, and MD4 hashing it... So no matter what, if you as a user is going to store this hash in LLDAP for Samba integration, it's not going to be cryptographically secure, and you have to be okay with that.

But, I agree all plugins should not just have access to this, so for now, I've added a specific configuration parameter to the LLDAP config for the plugin, that needs to be set, before the plugin is allowed to register a listener for this handler (config is called allow_on_ldap_password_update, bool). So at the very least, you have to explicitly grant plugins access to listen to this handler.

I'd really like this functionality, but I agree, plugins should not be able to just tap into it without explicit permission.

This is actually partly why I thought a plugin system should be the solution to the Samba integration I'm striving for.

(re: on_bind, maybe something similar could be done there; though maybe allowing the handler, but only getting the password if allowed).

crates/plugin-engine/src/api/types.rs
broeng: This is basicly the config as given from the LLDAP config file, so the kvscope here is meant for explicitly defining a scope for this plugins storage. If it's None, which would be the general case, it will become the name of the plugin when the plugin is loaded. At this point I haven't accessed the registration info from the plugin yet.

crates/plugin-engine/src/internal/lualib/init.rs
broeng: Yes, that was what I was thinking.

crates/plugin-engine/src/internal/lualib/tables.rs
broeng: Ehm, yeah, I don't think that bit is done, yet (or working, for that matter). A part of my experimentation with matching on LDAP filters from the Samba plugin :-)

crates/key-value-store/src/api/store.rs
```diff
@@ -0,0 +1,29 @@
...
+    //
+    // Store an arbitrary, serializable value under the given key
+    //
+    async fn store<V: Serialize + Send>(&self, key: K, value: V) -> Result<(), KeyValueError>;
```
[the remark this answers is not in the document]
broeng: Yeah, it should be, I agree. Keep as-is for now?

crates/key-value-store/src/api/store.rs
broeng: Followup, should it actually be triple-slash I use? That's what would go to docs, right?

crates/key-value-store/src/api/store.rs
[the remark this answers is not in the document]
nitnelave: Yes

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
nitnelave: I don't know how it would work on the Lua side, but maybe it could be different types of errors? Processing failure (e.g. assert fail), fatal error (request noisily aborted), cancel error (request "successfully" aborted)

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
nitnelave: Kinda related, when we return a UnwillingToPerform or similar, plugins should ideally be able to override that

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
nitnelave: In that case you have to fail any attempt to change the password using the opaque API (so that should be exposed too).

And that's where I don't like it :/

But I understand the problem, and if LLDAP becomes just an implementation detail that users never see, it's probably okay.

But yeah, it should come with a warning/specific mode. Maybe within the context of this listener, we could implement a "safe sandbox", with no outside communication? But that would be easy to bypass, just set the info you want to exfiltrate as a user attribute and register another listener.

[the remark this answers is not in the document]
nitnelave: That leads me to plugin permissions. Maybe we should have capabilities in the config: access to network (get/post), access to hashed passwords, access to plaintext passwords, ...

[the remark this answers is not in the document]
broeng: > That leads me to plugin permissions. Maybe we should have capabilities in the config: access to network (get/post), access to hashed passwords, access to plaintext passwords, ...

Yeah, that's probably a good idea. Could definitely see that make sense for network access as well.

crates/key-value-store/Cargo.toml
broeng: Done.

crates/key-value-store/Cargo.toml
broeng: Done.

crates/key-value-store/src/api/store.rs
broeng: Done.

crates/key-value-store/src/api/store.rs
broeng: Done.

crates/plugin-kv-store/src/test.rs
broeng: Done

crates/plugin-kv-store/src/store.rs
```diff
@@ -0,0 +1,137 @@
...
+                    // prepare model
+                    let next_val: i64 = result_val + 1;
+                    let model = model::plugin_key_values::ActiveModel {
+                        scope: ActiveValue::Set(key_scope.clone()),
+                        key: ActiveValue::Set(key.key.clone()),
+                        value: ActiveValue::Set(Serialized::from(&next_val)),
+                        ..Default::default()
+                    };
+                    if existing.is_some() {
+                        model.update(transaction).await?;
```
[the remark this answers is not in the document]
broeng: Done.

crates/plugin-kv-store/src/test.rs
```diff
@@ -0,0 +1,196 @@
...
+    async fn test_insert_and_fetch_string() {
+        // Setup: Load a database and get a connection
+        let db_conn = load_fixture("01").await;
+        let scope = PluginKVScope("test".to_string());
+        let store = PluginKeyValueStore::new(db_conn.connection.clone());
+        let test_key: String = "test-key".to_string();
+        // Verify: ensure nothing is stored under the key beforehand
+        let pre_res = store.fetch::<String>(scope.key(test_key.clone())).await;
+        assert!(pre_res.is_ok());
+        assert!(pre_res.unwrap().is_none());
```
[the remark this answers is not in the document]
broeng: Done.

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
broeng: Done.

crates/plugin-engine/src/api/types.rs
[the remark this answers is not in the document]
broeng: Done.

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
broeng: > Kinda related, when we return a UnwillingToPerform or similar, plugins should ideally be able to override that

I think that should be covered for all the cases (I'll go over it to check). It's possible to handle each operation type, and the plugin listeners are all called with the result from LLDAP, so UnwillingToPerform, for instance.

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
broeng: > I guess it's mostly relevant to the `bind`, since we could want to cancel it.

Just realizing that self mutates as a result of this. So a bit more is needed to cancel this, than just overriding the result. Same for supporting auth methods from plugins.

I think I'll let do_bind return the ValidationResults together with ResultCode and message, and then assign self.user_info after delegating to plugins. That _should_ address both issues.

server/src/domain/ldap/utils.rs
```diff
@@ -262,13 +262,35 @@ pub fn map_group_field(field: &AttributeName, schema: &PublicSchema) -> GroupFie
...
+#[derive(Clone, Debug)]
 pub struct LdapInfo {
     pub base_dn: Vec<(String, String)>,
     pub base_dn_str: String,
     pub ignored_user_attributes: Vec<AttributeName>,
     pub ignored_group_attributes: Vec<AttributeName>,
 }
 
+impl LdapInfo {
+    pub fn from(
```
nitnelave, line 274: That would be just `new`, not `from`

server/src/domain/plugin/backend.rs
```diff
@@ -0,0 +1,254 @@
...
+
+fn parse_user_filter(
+    schema: &Schema,
+    ldap_info: &LdapInfo,
+    filter: String,
+) -> Result<UserRequestFilter, String> {
+    let ldap_filter: LdapFilter =
+        filter::parse_ldap_filter_str(filter.as_str()).map_err(|e| e.to_string())?;
+    //parse_ldap_user_filter(schema, ldap_info, ldap_filter)
+    let pub_schema = PublicSchema::from(schema.clone());
```
nitnelave, line 241: If you're going to move the schema, take it by value (and let the caller decide whether to clone it or not, in that case it's not necessary).
Or in this case, what you actually need is a PublicSchema, so take _that_ by reference and let the caller figure out how to pass it to you.

server/src/domain/plugin_backend_handler.rs
```diff
@@ -0,0 +1,510 @@
...
+    },
+    infra::{ldap_handler::InternalSearchResults, tcp_backend_handler::TcpBackendHandler},
+};
+
+use tracing::instrument;
+
+#[derive(Clone, Debug)]
+pub struct PluginBackendHandler {
+    pub(crate) backend_handler: SqlBackendHandler,
+    pub(crate) backend_api: &'static ServerBackendAPI<SqlBackendHandler>,
```
nitnelave, line 45: Why is it `&'static`? Why not keep the `Box`? Is it so that the `PluginContext` can keep a reference to it?
I know it's not a final review, but let's have a comment about the why, and let's be careful about what we put inside, since it doesn't get destroyed

server/src/domain/plugin_backend_handler.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+    pub fn new_plugin_context(&self) -> PluginContext<ServerBackendAPI<SqlBackendHandler>> {
+        PluginContext::new(
+            self.backend_api,
+            self.request_context.validation_results.clone(),
+        )
+    }
+    pub async fn initialize_plugins(&self) -> std::result::Result<(), String> {
+        let ctx = self.new_plugin_context();
+        self.plugin_handler.initialize_plugins(ctx).await
```
nitnelave, line 79: That reminds me that in the final PR, I'll probably bicker about what gets printed when initializing plugins (name, version, permission/capabilities, update URL, whether there's a new version available, ...)

server/src/domain/plugin_backend_handler.rs
```diff
@@ -0,0 +1,510 @@
...
+    async fn get_schema(&self) -> Result<Schema> {
+        self.backend_handler.get_schema().await
+    }
+}
+
+#[async_trait]
+impl SchemaBackendHandler for PluginBackendHandler {
+    #[instrument(skip(self), level = "debug", ret, err)]
+    async fn add_user_attribute(&self, request: CreateAttributeRequest) -> Result<()> {
+        self.backend_handler.add_user_attribute(request).await
```
nitnelave, line 109: I suppose this will actually have plugin calls later on?

server/src/domain/plugin_backend_handler.rs
```diff
@@ -0,0 +1,510 @@
...
+    async fn update_group(&self, request: UpdateGroupRequest) -> Result<()> {
+        let ctx = self.new_plugin_context();
+        let request = self
+            .plugin_handler
+            .on_update_group(ctx.clone(), request.clone())
+            .await
+            .unwrap_or(request);
+        self.backend_handler.update_group(request.clone()).await?;
+        let _ = self.plugin_handler.on_updated_group(ctx, request).await;
+        Ok(())
```
nitnelave, line 157: We can probably extract that to a function, or at least a macro

server/src/domain/sql_backend_handler.rs
```diff
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 
 use lldap_domain_handlers::handler::{BackendHandler, RequestContext, WithContextHandler};
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
nitnelave, line 6: Erm, what are you hoping to get from this debug implementation?

I'd rather have a custom debug impl that doesn't print anything, rather than a bunch of implementation details. Then there's the question of _why_ you need it to be Debug

server/src/infra/configuration.rs
```diff
@@ -136,6 +156,8 @@ pub struct Configuration {
     #[serde(skip)]
     #[builder(field(private), default = "None")]
     server_setup: Option<ServerSetupConfig>,
+    #[builder(default = r#"Vec::new()"#)]
+    pub plugin: Vec<PluginOptions>,
```
nitnelave, line 160: How does that work with env variables? Currently, every LLDAP setting is also available as an env variable.

I propose a different structure:
```toml
[[plugins.pam_plugin]]
enabled = true
plugin_path = "plugins/pam.lua"
[[plugins.pam_plugin.config]]
uid_offset = "10000"
```

With some checks that the loaded plugin declares the same name as the config.

server/src/infra/configuration.rs
```diff
@@ -81,6 +81,26 @@ impl std::default::Default for LdapsOptions {
...
 
+#[derive(Clone, Debug, Deserialize, Serialize, derive_builder::Builder)]
+#[builder(pattern = "owned")]
+pub struct PluginOptions {
+    #[builder(default = "true")]
+    pub enabled: bool,
+    #[builder(default = "false")]
+    pub allow_on_password_update: bool,
+    #[builder(default = r#"None"#)]
+    pub kvscope: Option<String>,
```
nitnelave, line 92: I know you talked about it, but I still don't see why we need this scope. Why _isn't_ this always the plugin name?

server/src/infra/configuration.rs
```diff
@@ -81,6 +81,26 @@ impl std::default::Default for LdapsOptions {
...
+    pub plugin_path: PathBuf,
+    #[builder(default = r#"None"#)]
+    pub config: Option<BTreeMap<String, String>>,
```
nitnelave, line 95: An empty map should be enough of a default value, no need to wrap it in an Option. Or does it have a different meaning?

server/src/domain/sql_backend_handler.rs
```diff
@@ -87,7 +93,7 @@ pub mod tests {
             .unwrap();
     }
 
-    pub async fn insert_user_no_password(handler: &SqlBackendHandler, name: &str) {
+    pub async fn insert_user_no_password<A: BackendHandler>(handler: &A, name: &str) {
```
nitnelave, line 96: I'm not going to review this commit, it feels like it'll change a lot after our conversation on the trait/handler wrapping.

server/src/domain/plugin/backend.rs
```diff
@@ -0,0 +1,254 @@
...
+    async fn get_schema(&self) -> Result<Schema, String> {
+        self.backend_handler
+            .get_schema()
+            .await
+            .map_err(|e| e.to_string())
+    }
+    //
+    // User Listing
+    //
+    // TODO: should probably just take an enum type instead
```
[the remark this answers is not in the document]
broeng: Right now it takes an ldap filter string. I thought that might be a nice convenience, but maybe it should just take an AST enum structure instead, for specifying the filter.

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+        lldap.log:debug("Created '" .. group_name .. "' group with id " .. tostring(id))
+        local group_gidnumber = get_config_gid_offset(context) + id
+        local res, err = context.api:update_group({
+            group_id = id,
+            insert_attributes = {
+                gidnumber = { int = group_gidnumber }
+            }
+        })
+        if err ~= nil then
+            error("Failed to set gidnumber attribute to '" .. group_name .. "' group", 1)
```
nitnelave, line 83: Just jotting down my thoughts: We need to be careful/intentional about logging levels. Again, this is something I don't want plugin authors to have to think about, but I don't want to flood the logs just because you installed a plugin. That means that unless you enable debug logs or the plugins log an error/warning, you don't see them (apart from initialization). If they _do_ log an error, but you're not in debug logs, you might not see the whole stacktrace leading to the error, so the message has to be prepended with "plugin pam_plugin:" or similar.

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+    else
+        group_id = groups[1].attributes.gidnumber.int
+    end
+    return group_id
+end
+
+local resolve_next_uid = function(context, uid_offset)
+    local nextuid, err = context.kvstore:fetch_and_increment(KVKEYS.NEXT_UID, uid_offset)
+    if err ~= nil then
+        error("Unable to resolve next-uid from plugin KV store. Error: " .. tostring(err), 1)
```
nitnelave, line 96: I guess that's a question of lua style, but I think in that case the idiomatic thing would be to return 2 values from the function, go-style, instead of logging. Here you'd be returning `nil, "Unable..."` and below just `return nextuid` (the second `nil` is implicit when unpacking the result at the callsite)

nitnelave: And actually, that error is not handled in the callsite, and will lead to some weirdness, no?

server/src/domain/plugin/backend.rs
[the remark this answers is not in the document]
nitnelave: We can provide both. Feel free to start with just one, maybe wrapped in a single-value enum for the sake of extensibility (like `LdapFilter(String)`)

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+    if err ~= nil then
+        -- Error
+        error("Failed to list_users: " .. tostring(err), 1)
+    end
+    if not lldap.tables:empty(users) then
+        -- obtain id of shared pam group
+        local group_id = resolve_gid_default_group(context)
+        lldap.log:debug("Resolved default group to ID: " .. tostring(group_id))
+        -- assign uid/gid to any users missing
+        for idx, user_and_group in pairs(users) do
```
nitnelave, line 116: That _screams_ for a lock.

I mean, the worst case is that you'll skip over a number with 2 queries setting the "next uid" to the user, but still it's a great showcase of a lock.

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+    end
+    return nextuid
+end
+
+local assign_user_attributes = function(context)
+    -- Ensure that the uidNumber for the created user is unique.
+    lldap.log:debug("Resolving current maximum uid and gid")
+    local uid_offset = get_config_uid_offset(context)
+    -- obtain list of all users in the system
+    local users, err = context.api:list_users({})
```
nitnelave, line 106: You can directly list the users with no uidnumber with the filter

server/src/domain/plugin_backend_handler.rs
broeng: Yes. I just wanted to get the structure ACK'ed before doing more. Seemed unnecessary, if they all needed to be changed, or something.

server/src/domain/plugin_backend_handler.rs
broeng: Not a bad idea with a macro. Were you thinking the whole block?

server/src/domain/sql_backend_handler.rs
broeng: It's just a leftover from some debugging, I won't need it, I'm pretty sure, and I'll also go over all of these additions to existing types to make sure they are actually needed in the final result.

plugins/samba.lua
```diff
@@ -0,0 +1,130 @@
+if lldap == nil then
+    lldap = {}
+end
+
```
nitnelave, line 4: What's that for?

server/src/domain/plugin_backend_handler.rs
nitnelave: Yeah, reduce boilerplate to enforce consistency. Something like
```rust
wrap_call_with_plugin_callbacks!(update_group, on_update_group, on_updated_group);
```

server/src/infra/configuration.rs
broeng: That's perfectly fine with me, honestly expected you to have a strong opinion here, so haven't given this too much attention. :-)

What about just "plugins.pam" though, instead of "plugins.pam_plugin"? Latter seems a bit redundant, and the former fits a bit more with what the plugin would declare as well?

plugins/samba.lua
```diff
@@ -0,0 +1,130 @@
...
+local ensure_user_attribute_exists = function(context, schema, attribute_name, attribute_type)
+    if schema.user_attributes.attributes[attribute_name] == nil then
+        local res = context.api:add_user_attribute({
```
nitnelave, line 7: It seems like you could just create the attribute regardless and swallow the error.

server/src/infra/configuration.rs
broeng: I think it's valuable for the user to be able to override what the plugin implementer has declared.

It would allow use cases like:

- Clearing the existing state at deployment, by adding an index to the name. Effectively providing a clean scope very time, without needing to do deletions on the DB.
- Debug if issues with a new version of a plugin is due to the existing state, by again giving a clean scope, without actually deleting the old.
- Maybe it's meaningful for some plugins to instantiate them more than once, configure them differently, and separate their state.
- Maybe it could be convenient to give a group of plugins the same scope.

I just think it's a convenient control for the user to have.

server/src/infra/configuration.rs
broeng: I agree, and I have been trying to bend figment to my will, but even with a default like BTreeMap::new(), and similar, I couldn't get it to accept a configuration where that key was missing, if it wasn't option. I know there are others that work fine without Option, by having a default, but I couldn't get it to work for this BTreeMap for some reason.

plugins/samba.lua
```diff
@@ -0,0 +1,130 @@
...
+            lldap.log:warn("Got error from creating 'sambaSamAccount' user object class")
+            return res
+        end
+    end
+end
+
+local on_password_update = function(context, args)
+    lldap.log:debug("New password being set for " .. args.user_id)
+    utf16_bytes = lldap.strings:to_utf16le(args.password)
+    hashed = lldap.hashing:md4_hash_bytes(utf16bytes)
```
nitnelave, line 60: I still think we could give the md4 of the (utf-16) password instead of the cleartext

server/src/infra/configuration.rs
nitnelave: Yeah, `pam_plugin` was my placeholder for the plugin name. It would be `plugins.pam` in your case.

server/src/infra/configuration.rs
nitnelave: SGTM

server/src/infra/configuration.rs
nitnelave: Have a new type wrapping the map, and that has a Default value

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+        return context.configuration.default_group
+    else
+        -- return default
+        return "pam_users"
+    end
+end
+
+local resolve_gid_default_group = function(context)
+    local group_name = get_config_default_group(context)
+    local groups = context.api:list_groups({ filter = "(cn=" .. group_name .. ")" })
```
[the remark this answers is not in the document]
broeng: That was actually the enum I was talking about here :-)

https://github.com/lldap/lldap/pull/1119#discussion_r1978272296

I'm thinking making a new enum, that mimics LdapFilter a bit, and make conversion to LdapFilter from it, from which it can be converted to both UserFilter and GroupFilter. Then the same enum can be used for both list_users and list_groups.

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+    local group_id = nil
+    if lldap.tables:empty(groups) then
+        lldap.log:debug("Creating group:" .. group_name)
+        local id, err = context.api:create_group({
+            display_name = group_name,
+            attributes = {}
+        })
+        if err ~= nil then
+            -- Error
+            error("Failed to create group: " .. group_name, 1)
```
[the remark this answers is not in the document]
broeng: It might tie in with general error handling... But, my thinking here was, that it's part of initialization, and if it can't even create the required group, then the plugin is not really going to be in a working state. Maybe it's then better to get the error up front, block start up, so the problem can either be resolved, or the plugin removed/fixed, instead of running in a non-working state?

plugins/pam.lua
broeng: We could make lldap.log:{debug,info,error} prepend the plugin name automatically? The log object should maybe be moved to context then, so it has the information...

Just to be sure; the error function here is not for logging, it aborts the script.

plugins/pam.lua
broeng: > And actually, that error is not handled in the callsite, and will lead to some weirdness, no?

Yeah, the intention was actually to abort the script -- not meaning it had been fully thought through though. Writing this was what made me make the todo on error return values from plugin :-)

plugins/pam.lua
broeng: You can search for "doesn't have the attribute"? Wasn't aware of that.

plugins/pam.lua
broeng: Don't disagree, but it is only for initialization. Once initialized they get the ID before creation with on_create_user, which only one node would proces.

But I agree, locks could be convenient. It seems like something that'd be suitable for a followup PR afterwards (I expect I'd still be up for doing it though).

plugins/samba.lua
broeng: Just my syntax highlighter, don't plan to keep it in :-)

plugins/samba.lua
broeng: I like the permission model idea, where you can choose a number of different types of hashed passwords, yes, but I also think it's convenient to allow one of the options to be cleartext. Maybe the plugin handles keeping some external, non-LDAP compatible system in sync with password -- or need it for implementing some alternative bind auth mechanism.

An MD4'ed UTF16LE value would be fine for my use, though, and what I'd use.

nitnelave: Let's start with just different flavors of hashing, leave room in the design for plaintext, and let the potential users come and argue for it.

server/src/domain/sql_backend_handler.rs
broeng: Marking this one as resolved, as we decided to pass the context, instead of using this trait.

crates/domain-handlers/src/handler.rs
broeng: We went with option two here, marking this as resolved.

server/src/domain/ldap/utils.rs
broeng: Done

server/src/domain/plugin/backend.rs
broeng: Done.

server/src/domain/sql_backend_handler.rs
broeng: Removed, including some others.

crates/plugin-engine/src/api/handler.rs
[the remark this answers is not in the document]
broeng: Done

server/src/infra/configuration.rs
broeng: Restructured as discussed.

Are we sure, we'd like to enforce that the same name is used as the plugin is declaring? It would deny the possibility of declaring a plugin twice, which might be interesting for a web-hook plugin (with different configs, I'd assume). It's still possible to reliably provide setup instructions using env vars without us forcing it.

plugins/pam.lua
```diff
@@ -0,0 +1,264 @@
...
+--
+-- Strives to maintain:
+--   Users: all have a unique uidnumber
+--   Users: all users by default have a gidnumber corresponding to the
+--          pam_users group created.
+--   Groups: all groups have a unique uidnumber
+--
+-- Group will only be created when a user needs the gidnumber for
+-- it, this is done to avoid it being created as group 1, before
+-- lldap_admin, in new setups.
```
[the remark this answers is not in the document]
broeng: Done.

crates/plugin-engine/src/internal/arguments/bind_request.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::api::arguments::ldap_bind_result::BindResult;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BindRequestArguments {
+    pub bind_result: BindResult,
+    pub bind_request: LdapBindRequest,
+}
+
+impl BindRequestArguments {
+    pub fn new(result: BindResult, request: LdapBindRequest) -> Self {
```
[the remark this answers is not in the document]
broeng: Removed

plugins/samba.lua
broeng: Implemented a version of the solution discussed on Discord. Avoids passing a cleartext password.

Marking this as resolved; there'll be review of the implemented solution later.

plugins/pam.lua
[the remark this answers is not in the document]
broeng: Introduced the enum, and started with an LdapFilter enum value alone for now, as suggested.

server/src/domain/plugin/backend.rs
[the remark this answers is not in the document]
broeng: Done.

server/src/domain/plugin_backend_handler.rs
broeng: Finished the last few handlers. Done.

PR #1134, opened 2025-03-29
example_configs: Fix nslcd group member mapping.

meetpatty:
If a user uses the template nslcd.conf as per the PAM configuration guide group members are not correctly mapped.

Noted when querying groups via `getent group` which was returning `error writing group entry: Invalid argument` for ldap groups with group members.


nitnelave (approved):
Thanks!

Issue #1145, opened 2025-04-01
[FEATURE REQUEST] Use SSSD instead of NSLCD for unix pam


vmd1:
**Motivation**
nslcd is no longer actively supported, but sssd is, as well as having a much larger variety of feautures

**Describe the solution you'd like**
A change of the guide for unix pam to allow it to use sssd

**Describe alternatives you've considered**
I'm going to try setting it up tomorrow and see if it works, if it does i'll see if i can make a guide + send a pr

**Additional context**

nitnelave:
If you can make it work, it'll be most welcome!

vmd1:
Just pinging in to say i managed to get it working, and I'm making a guide to add.

i'm also planning on adding a small section on how to do ssh public key sync from ldap, and how to sync permissions such as sudo or docker groups

vmd1:
I've sent a PR in with my changes: [Move from NSLCD to SSSD](https://github.com/lldap/lldap/pull/1146)

PR #1146, opened 2025-04-03
Move from NSLCD to SSSD

vmd1:
NSLCD is no longer actively supported, and its recommended to move to SSSD, which is significantly more feauture-rich.

I've added a guide here with an example config and an explanation on:
- How to setup SSSD
- How to setup SSH Key Sync with SSSD
- How to sync permissions for tools such as sudo

I've tested this out on a fresh install of Ubuntu 24.04.
Any suggestions for things to change?
Thanks.


vmd1:
#1145

example_configs/pam/NSLCD.md
```diff
@@ -0,0 +1,90 @@
...
+LDAPS certificate, which must be available on the client. This is used to
+verify the LDAPS server identity.
+
+You will need to add the `binddn` and `bindpw` settings.
+
+The provided implementation uses custom attributes to mark users and groups
+that should be included in the system (for instance, you don't want LDAP
+accounts of other services to have a matching unix user).
+
+For users, you need to add an (integer) `unix-uid` attribute to the schema, and
```
nitnelave, line 45: What about home directory, ssh keys and so on? Or is that the PAM config? How do they relate to each other?

example_configs/pam/README.md
```diff
@@ -1,90 +1,167 @@
...
+E.g. Using nano
 
-You will need to add the `binddn` and `bindpw` settings.
+```bash
+sudo nano /etc/sssd/sssd.conf
+```
+
+Insert the contents of the provided template (sssd.conf), and you will need to change the content on the following line numbers:
+
+- Line 3
```
nitnelave, line 63: Line numbers are not very robust: add a comment at the top and everything is off by one. Could you mention _what_ you need to edit instead?

nitnelave (approved):

example_configs/pam/README.md
vmd1: Would it be a better idea to add comments into the template config file of what to edit, and say in the README to just edit those things?

example_configs/pam/NSLCD.md
vmd1: nslcd doesn't really have native support for ssh keys, but I can add the home directory and how to do that. should I also mention that nslcd doesn't actually support ssh keys? I could also add a comparison between sssd and nslcd?

nitnelave: I don't know that much about these tools: what does pam vs nslcd vs Saad do? Who cares about the home directory, or the ssh keys?

example_configs/pam/README.md
nitnelave: That sounds like a good idea

vmd1: Alright, ill start on that in a bit.

example_configs/pam/NSLCD.md
nitnelave: Okay, so you're saying that for nslcd we don't care about home/ssh keys, but they should be in the Pam config?

Then that's okay for me

vmd1: Yes, thats about right, but PAM can't actually handle SSH keys, thats handled by the SSH server.

The main reason I kept this file was it was the original README, so I was just keeping it for those who wanted to see the old docs, or use NSLCD explicitly

vmd1:
How about now? 
I've added comments instead of line numbers to be more specific.
The nslcd config guide is more of a legacy guide in the sense it's just there for those who may have a specific setup/want to see the previous docs.

nitnelave (approved):

PR #1155, opened 2025-04-21
Include custom attributes when bootstrapping user and group configs

### LINKED ISSUE #1088: [FEATURE REQUEST] Boostrap user/group schema fields
**Motivation**

The bootstrap script can [create schemas](https://github.com/lldap/lldap/pull/991), but the creation of users and groups does not take into account the extra fields.

Being able to add the extra schema fields in the user and group files and have the bootstrap script initialize those values as well.

For instance:
```json
{
  "id": "username",
  "email": "username@example.com",
  "password": "changeme",
  "displayName": "Display Name",
  "firstName": "First",
  "lastName": "Last",
  "avatar_file": "/path/to/avatar.jpg",
  "avatar_url": "https://i.imgur.com/nbCxk3z.jpg",
  "gravatar_avatar": "false",
  "weserv_avatar": "false",
  "groups": [
    "group-1",
    "group-2"
  ],
  "unix-uid": 10000
}
```

Set via ldap ?

**Additional context**

avargaskun:
Fixes #1088 

Update the bootstrap script to include additional attributes for users and groups.

I've done light testing on these changes but they seem to work fine for all supported data types.

Currently I use this script by mapping it to a separate folder with docker (i.e. -v "<script_location>":/scripts) and executing it from there.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -375,16 +375,28 @@ update_group_attributes() {
   local attributes_json="$2"
   
   # shellcheck disable=SC2016
-  local query='{"query":"mutation UpdateGroup($group: UpdateGroupInput!) {updateGroup(group: $group) {ok}}","operationName":"UpdateGroup"}'
-
-  local variables="{\"group\":{\"id\":$group_id,\"insertAttributes\":$attributes_json}}"
+  local query
+  query=$(jq -n -c \
+    --argjson groupId "$group_id" \
+    --argjson attributes "$attributes_json" '
+    {
+      "query": "mutation UpdateGroup($group: UpdateGroupInput!) {updateGroup(group: $group) {ok}}",
+      "operationName": "UpdateGroup",
+      "variables": {
+        "group": {
+          "id": $groupId,
+          "insertAttributes": $attributes
+        }
+      }
+    }
+  ')
   
   local response='' error=''
   response="$(curl --silent --request POST \
     --url "$LLDAP_URL/api/graphql" \
     --header "Authorization: Bearer $TOKEN" \
     --header 'Content-Type: application/json' \
-    --data "{\"query\":\"mutation UpdateGroup(\$group: UpdateGroupInput!) {updateGroup(group: \$group) {ok}}\",\"operationName\":\"UpdateGroup\",\"variables\":$variables}")"
+    --data "$query")"
   
   error="$(printf '%s' "$response" | jq --raw-output '.errors | if . != null then .[].message else empty end')"
   if [[ -n "$error" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.sh
```diff
@@ -180,7 +180,7 @@ get_user_details() {
   local id="$1"
 
   # shellcheck disable=SC2016
-  local query='{"query":"query GetUserDetails($id: String!) {user(userId: $id) {id email displayName firstName lastName creationDate uuid groups {id displayName}}}","operationName":"GetUserDetails"}'
+  local query='{"query":"query GetUserDetails($id: String!) {user(userId: $id) {id email displayName firstName lastName creationDate uuid groups {id displayName} attributes {name value}}}","operationName":"GetUserDetails"}'
```
nitnelave, line 183: Do you think you can format it better with a heredoc?

scripts/bootstrap.sh
```diff
@@ -364,6 +364,104 @@ create_user_schema_property() {
   fi
 }
 
+update_group_attributes() {
+  local group_id="$1"
+  local attributes_json="$2"
+  
+  # shellcheck disable=SC2016
+  local query='{"query":"mutation UpdateGroup($group: UpdateGroupInput!) {updateGroup(group: $group) {ok}}","operationName":"UpdateGroup"}'
```
nitnelave, line 372: Are you using this variable? It seems repeated below

scripts/bootstrap.sh
```diff
@@ -364,6 +364,104 @@ create_user_schema_property() {
...
+    printf 'Custom attributes for group ID "%s" successfully updated\n' "$group_id"
+  fi
+}
+
+update_user_attributes() {
+  local user_id="$1"
+  local attributes_json="$2"
+  
+  # shellcheck disable=SC2016
+  local query='{"query":"mutation UpdateUser($user: UpdateUserInput!) {updateUser(user: $user) {ok}}","operationName":"UpdateUser"}'
```
nitnelave, line 396: Same here

scripts/bootstrap.sh
avargaskun: Sure! I've added a heredoc to make this more readable but I don't think I can split the graphql query into multiple lines because it's defined as a string in a JSON object?

nitnelave: Ah, damn, JSON is indeed pretty limiting. Feel free to revert to the previous version without the extra cat, it seems simpler

nitnelave (approved):
Thanks!

scripts/bootstrap.sh
avargaskun: Agreed. Reverted to original change.

scripts/bootstrap.sh
avargaskun: Good catch. I made a change to make the construction of the query more readable using jq.

scripts/bootstrap.sh
avargaskun: Ditto

PR #1157, opened 2025-04-23
Update mailserver.md

cogdavid:
Updated ACCOUNT_PROVISIONER variable in environment directive. Change from ACCOUNT_PROVISIONER=LDAP to ACCOUNT_PROVISIONER=FILE as LDAP provisioning is not fully supported in Dovecot.


nitnelave (approved):
Thanks!

PR #1178, opened 2025-05-19
server: add ldap_user_display_name config

karaolidis:


nitnelave:
What is the motivation for this change? The display name can easily be changed once the admin is created.

karaolidis:
I'd like to declare all `lldap` options with Nix on my deployment without having to manually change the display name afterwards :)

 I am fairly certain this change should be backwards-compatible, unless I've missed something?

nitnelave:
You can use the bootstrap script to declaratively manage your users.

Yes, the change is backwards compatible, but it adds (IMO unnecessary) complexity to the software, both in terms of maintenance but also (and that's what I'm mainly worried about) in terms of usage/configuration. Every new option makes the config/docs longer and should pull their own weight.

PR #1181, opened 2025-05-29
fix: nextcloud example allows all ldap users to login

ThorpeJosh:
Nextcloud has 2 sections in the LDAP/AD integration app called "Users" and "Login Attributes".

I don't fully understand the difference between the two, however I followed the current lldap nextcloud example and discovered the following security issue...

With the Login Attributes filter set to `(&(objectclass=person)(uid=%uid))` any ldap user (even those that are not in the `nextcloud_users` group) can login to the nextcloud instance and get a fully functioning nextcloud account provisioned.

Even more confusing is that any user that is not part of `nextcloud_users` group that logs into nextcloud, does not show up as a user in the nextcloud administrator portal, making these users almost an invisible user to the nextcloud administrator, but can still login, user the nextcloud services, etc.

The proposed changes add the same filter from the examples "Users" section, to the "Login" section so that only ldap users in the "nextcloud_users" group are allowed to login.


nitnelave (approved):
Nice catch!

PR #1188, opened 2025-06-06
Mailcow LLDAP Example

PHIDIAS0303:
A file for reference when using LLDAP in mailcow as LDAP.

### DIFF
--- a/example_configs/mailcow.md
+++ b/example_configs/mailcow.md
@@ -26,7 +26,7 @@ Set `Base DN` in the format like `dc=example,dc=org`.
 
 Set `Username Field` as `mail` or `uid`.
 
-Set `Filter` as `(objectclass=inetOrgPerson)` or `person`.
+Set `Filter` as `(objectclass=inetOrgPerson)` or `(objectclass=person)`.
 
 Set `Attribute Field` also as `mail`.
 
--- a/example_configs/mailcow.md
+++ b/example_configs/mailcow.md
@@ -14,11 +14,13 @@ Set `Host` as the LLDAP IP.
 
 Set `Port` as the LLDAP port, which should be `3890` by default.
 
-Set `Use SSL` (LDAPS, the port like 389 will become 636) to `False`.
+Set `Use SSL` to `False`.
 
 Set `Use StartTLS` (SSL port will not be used) to `False`.
 
 Set `Ignore SSL Errors` (validation bypass) to `False`.
+
+When using `Use SSL` (LDAPS), `Port` should be set to `6360` instead, the default port if unchanged.
 
 These three options should be `False`, unless there is a properly-signed non self-signed certificate for LLDAP installed.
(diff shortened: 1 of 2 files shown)


example_configs/mailcow.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Configuration
+
+Set `Identity Provider` as `LDAP`
+
+Set `Host` as the LLDAP IP.
+
+Set `Port` as the LLDAP port, which should be `3890` by default.
+
+Set `Use SSL`, `Use StartTLS`, `Ignore SSL Errors` to `False` if you did not enable SSL for LLDAP.
```
nitnelave: Maybe mention that otherwise it's SSL, not StartTLS, and that if you have a properly-signed certificate (not self-signed, or with the proper root installed) you shouldn't ignore errors.

example_configs/mailcow.md
```diff
@@ -0,0 +1,34 @@
...
+
+Set `Base DN` in the format like `dc=example,dc=org`.
+
+Set `Username Field`, `Attribute Field` as `mail`.
```
nitnelave: Not "uid" for username?

example_configs/mailcow.md
```diff
@@ -0,0 +1,34 @@
...
+
+Set `Filter` as `(objectclass=inetOrgPerson)`.
```
nitnelave: Or just "person" works

nitnelave (commented):
Thanks!

nitnelave:
Oh, and can you add a link in the readme?

example_configs/mailcow.md
```diff
@@ -0,0 +1,43 @@
...
+
+Set `Host` as the LLDAP IP.
+
+Set `Port` as the LLDAP port, which should be `3890` by default.
+
+Set `Use SSL` (LDAPS, the port like 389 will become 636) to `False`.
+
+Set `Use StartTLS` (SSL port will not be used) to `False`.
+
+Set `Ignore SSL Errors` (validation bypass) to `False`.
```
nitnelave: I think this part is not very clear. It might be best to say: 
- if using LDAPS, set these variables like so (maybe include the port there) 
- otherwise, set them like so

example_configs/mailcow.md
```diff
@@ -0,0 +1,43 @@
...
+
+These three options should be `False`, unless there is a properly-signed non self-signed certificate for LLDAP installed.
+
+Set `Base DN` in the format like `dc=example,dc=org`.
+
+Set `Username Field` as `mail` or `uid`.
+
+Set `Filter` as `(objectclass=inetOrgPerson)` or `(objectclass=person)`.
```
nitnelave: Let's not put both :) Person is probably the simplest.

nitnelave (approved):

Issue #1196, opened 2025-07-04
[DOCS] Virtualhost support ?


selfhoster1312:
I don't think it's clear from the docs whether lldap can support multiple vhosts and per-vhost admins. From what i've tested so far, i believe:

- it's possible to create users on different domains by using the same key for username and email fields (this way you can have 2 accounts contact@foo.com and contact@foo.net, whereas you could not have two accounts called `contact`)
- it's not possible to create multiple bases on LDAP side (`ou=people,dc=foo,dc=net` and `ou=people,dc=foo,dc=com`) unless running multiple instances of lldap
- it's not possible to have custom permissions for a group, for example allow a group to add/edit/delete accounts matching a regex (eg. `.*@foo.net$`)

This is personal belief based on quick testing. Please correct this information with actual knowledge of the way lldap operates 😄 . I understand that lldap tries to stay minimalist and that's what brought me here. I believe however the answer to these questions should be added to the docs/FAQ.

Why it's important to me: as a member of several hosting non-profits, we try to provide reliable service to different collectives, each having their own vhost. We've been investigating replacing different existing stacks based on slapd, moulinette, or PAM (yes, with /etc/passwd integration) with a more modern/maintainable alternative. It's not a problem to run one service per vhost (although having proper vhost support in lldap would be easier) but i could not find documentation about this in the lldap docs.

nitnelave:
Your assessment is correct: 
- one base DC per instance
  - the base DC is only in the config, it's not even stored in the DB
- only 2 hardcoded OUs (people and groups)
- the permissions are simple: only admin, password manager, or read-only, all at the global scope

Feel free to send a PR with some proposed documentation that you feel was missing!

selfhoster1312:
Thanks for the quick answer.

I will definitely send a PR. The README is really long: i personally would be tempted to break it up into smaller pages (or at least smaller collapsible sections), or is it on purpose that it's so long? In the meantime, i could submit it in the "Recommended architecture" section: would this place be appropriate?

About the vhosts support, two questions:

- can you see potential problems in using fully-qualified emails as username in the same base DC to mock proper vhosts (minus the fine-grained permissions per-domain)?
- would vhosts support be accepted as a contribution, or is it considered out of scope for the project?

nitnelave:
Vhosts are definitely out of scope.

[...]

PR #1198, opened 2025-07-06
docs: Split README into smaller files

selfhoster1312:
As discussed in #1196, paving the way for more FAQs (like #1196 and #1197).

I didn't fundamentally change the content, but editorialized a bit.

Feedback is welcome.

PR #1201, opened 2025-07-06
Add FAQ about virtualhosts

selfhoster1312:
This is my understanding from #1196.

Please merge #1998 first.

I have no idea how to document reverse proxying of several instances so i just mentioned it's theoretically possible.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -4,6 +4,7 @@
 - [Discord Integration](#discord-integration)
 - [Migrating from SQLite](#migrating-from-sqlite)
 - How does LLDAP compare [with OpenLDAP](#how-does-lldap-compare-with-openldap)? [With FreeIPA](#how-does-lldap-compare-with-freeipa)? [With Kanidm](#how-does-lldap-compare-with-kanidm)?
+- [Does LLDAP support vhosts?](#does-lldap-support-vhosts)
 
 ## I can't log in!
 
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -0,0 +1,96 @@
...
+read-only, and by having more moving parts it is inherently more complex. If
+you don't need to modify the users through LDAP and you're planning on
+installing something like [KeyCloak](https://www.keycloak.org) to provide
+modern identity protocols, check out Kanidm.
+
+## Does lldap support vhosts?
+
+lldap does not natively support virtualhosts, sometimes known as multi-tenancy:
+
+- users are all part of the same base DN (`ou=people,cn=example,cn=com`)
```
nitnelave:
```suggestion
- users are all part of the same base DN (`ou=people,dc=example,dc=com`)
```

docs/faq.md
```diff
@@ -0,0 +1,96 @@
...
+- you may support multiple domains by using fully-qualified email addresses as usernames (eg. `user@example.com` and `user@example.org`); however, you can't have more fine-grained permissions than what is provided by default (which applies across all users)
+
+lldap is very lightweight (~15MiB RAM on startup) and it's therefore possible to run one instance per virtualhost if you need to properly support multiple domains. However, reverse proxying those different instances over the same port/socket is not documented (contributions welcome).
```
nitnelave:
```suggestion
lldap is very lightweight (~15MiB RAM on startup) and it's therefore possible to run one instance per virtualhost if you need to properly support multiple domains.
```

I would leave it at that. I don't think you can realistically proxy different instances on the same ports, short of implementing a LDAP and graphql-aware reverse proxy with custom dispatching logic. Actually, even that would fall short since graphql queries don't mention the DC at all, so you wouldn't know where to send the query.

The obvious solution is to host them on different ports.

docs/faq.md
```diff
@@ -0,0 +1,96 @@
...
+platform, covering all your bases: OAuth, Linux accounts, SSH keys, Radius,
+WebAuthn. It comes with a (read-only) LDAPS server.
+
+It's fairly easy to install and does much more; but their LDAP server is
+read-only, and by having more moving parts it is inherently more complex. If
+you don't need to modify the users through LDAP and you're planning on
+installing something like [KeyCloak](https://www.keycloak.org) to provide
+modern identity protocols, check out Kanidm.
+
+## Does lldap support vhosts?
```
nitnelave:
```suggestion
## Does LLDAP support vhosts?
```

Same below

nitnelave (approved):

docs/faq.md
selfhoster1312: I had in mind a LDAP-aware reverse proxy (if that's a thing) because there's a standard port for the protocol. I was less concerned about the GraphQL API to be honest. I just removed that sentence to avoid possible confusion.

selfhoster1312:
Resolved the case, the confusing reverse proxying suggestion, and rebased on #1198

PR #1204, opened 2025-07-08
ldap: Add missing subschema entries

### LINKED ISSUE #1203: [BUG] Subschema missing attribute definitions used in LDAP responses
**Describe the bug**
The subschema response does not describe all attributes used in regular LDAP responses of e.g. user or groups searches. Some examples of this are: uid, cn, sn and ou as they do not have an `attributeTypes` entry in the subschema.

**To Reproduce**
Running the following query against the LLDAP server:
```bash
ldapsearch -H ldap://localhost:3890 -D uid=admin,ou=people,dc=example,dc=com -w password -o ldif-wrap=no -b cn=Subschema -s base +
```
Shows the attribute descriptions for attributes mentioned are not present.

**Expected behavior**
Although some clients might have internal attribute definitions ready, it seems to be general good practice to return all attributes used by the server. For this reason, the subschema response should (at least) return the definitions of the attributeTypes used in a regular search response.

T0byV:
Fixes #1203.

This PR implements all attributes used in a regular LLDAP search response:
```
ldapsearch -H ldap://localhost:3890 -D uid=admin,ou=people,dc=example,dc=com -w password -o ldif-wrap=no -b dc=example,dc=com
```

I walked through a response and comprised the following list of missing `attributeTypes` definitions and their `ldapSyntaxes` and `matchingRules`. **Please check if I didn't miss any**
We might technically not need the `matchingRules`, however I thought it might be good to be as complete as possible.

I also (maybe as a temporary solution) changed the start number of custom LLDAP attributes from 2.x to 10.x, as we're currently not using that range at all and we are now using the 2.x range with the RFC attribute defintions.

The list:
_Note that these entries were derived from OpenLDAP's implementation._
```
attributeTypes: ( 2.5.4.0 NAME 'objectClass' DESC 'RFC4512: object classes of the entity' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )
attributeTypes: ( 2.5.21.9 NAME 'structuralObjectClass' DESC 'RFC4512: structural object class of entry' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
attributeTypes: ( 2.5.4.3 NAME ( 'cn' 'commonName' ) DESC 'RFC4519: common name(s) for which the entity is known by' SUP name )
attributeTypes: ( 2.5.18.1 NAME 'createTimestamp' DESC 'RFC4512: time which object was created' EQUALITY generalizedTimeMatch ORDERING generalizedTimeOrderingMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
attributeTypes: ( 2.5.18.2 NAME 'modifyTimestamp' DESC 'RFC4512: time which object was last modified' EQUALITY generalizedTimeMatch ORDERING generalizedTimeOrderingMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
attributeTypes: ( 1.3.6.1.4.1.1466.101.120.16 NAME 'ldapSyntaxes' DESC 'RFC4512: LDAP syntaxes' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.54 USAGE directoryOperation )
attributeTypes: ( 2.5.21.5 NAME 'attributeTypes' DESC 'RFC4512: attribute types' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.3 USAGE directoryOperation )
attributeTypes: ( 2.5.21.6 NAME 'objectClasses' DESC 'RFC4512: object classes' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.37 USAGE directoryOperation )
attributeTypes: ( 0.9.2342.19200300.100.1.1 NAME ( 'uid' 'userid' ) DESC 'RFC4519: user identifier' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{256} )
attributeTypes: ( 2.5.4.11 NAME ( 'ou' 'organizationalUnitName' ) DESC 'RFC2256: organizational unit this object belongs to' SUP name )
attributeTypes: ( 1.3.6.1.1.16.4 NAME 'entryUUID' DESC 'UUID of the entry' EQUALITY UUIDMatch ORDERING UUIDOrderingMatch SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )
attributeTypes: ( 2.5.4.4 NAME ( 'sn' 'surname' ) DESC 'RFC2256: last (family) name(s) for which the entity is known by' SUP name )
attributeTypes: ( 2.5.4.50 NAME 'uniqueMember' DESC 'RFC2256: unique member of a group' EQUALITY uniqueMemberMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.34 )
attributeTypes: ( 2.5.4.31 NAME 'member' DESC 'RFC2256: member of a group' SUP distinguishedName )
attributeTypes: ( 2.5.4.41 NAME 'name' DESC 'RFC4519: common supertype of name attributes' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{32768} )
attributeTypes: ( 2.5.4.49 NAME 'distinguishedName' DESC 'RFC4519: common supertype of DN attributes' EQUALITY distinguishedNameMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 )

ldapSyntaxes: ( 1.3.6.1.4.1.1466.115.121.1.38 DESC 'OID' )
ldapSyntaxes: ( 1.3.6.1.4.1.1466.115.121.1.24 DESC 'Generalized Time' )
ldapSyntaxes: ( 1.3.6.1.4.1.1466.115.121.1.15 DESC 'Directory String' )
ldapSyntaxes: ( 1.3.6.1.1.16.1 DESC 'UUID' )
ldapSyntaxes: ( 1.3.6.1.4.1.1466.115.121.1.34 DESC 'Name And Optional UID' )
ldapSyntaxes: ( 1.3.6.1.4.1.1466.115.121.1.12 DESC 'Distinguished Name' )
# bellow ldapSyntaxes are not returned by an OpenLDAP server, I did implement them for sake of being complete.
( 1.3.6.1.4.1.1466.115.121.1.54 DESC 'LDAP Syntax Description' )
( 1.3.6.1.4.1.1466.115.121.1.3 DESC 'Attribute Type Description' )
( 1.3.6.1.4.1.1466.115.121.1.37 DESC 'Object Class Description' )
( 1.3.6.1.4.1.1466.115.121.1.58 DESC 'Substring Assertion' )

matchingRules: ( 2.5.13.0 NAME 'objectIdentifierMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )
matchingRules: ( 2.5.13.27 NAME 'generalizedTimeMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )
matchingRules: ( 2.5.13.28 NAME 'generalizedTimeOrderingMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )
matchingRules: ( 2.5.13.30 NAME 'objectIdentifierFirstComponentMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )
matchingRules: ( 2.5.13.2 NAME 'caseIgnoreMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
matchingRules: ( 2.5.13.4 NAME 'caseIgnoreSubstringsMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.58 )
matchingRules: ( 2.5.13.23 NAME 'uniqueMemberMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.34 )
matchingRules: ( 2.5.13.1 NAME 'distinguishedNameMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 )
matchingRules: ( 1.3.6.1.1.16.2 NAME 'UUIDMatch' SYNTAX 1.3.6.1.1.16.1 )
matchingRules: ( 1.3.6.1.1.16.3 NAME 'UUIDOrderingMatch' SYNTAX 1.3.6.1.1.16.1 )
```

### DIFF
--- a/crates/ldap/src/search.rs
+++ b/crates/ldap/src/search.rs
@@ -202,25 +202,66 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
            LdapPartialAttribute {
             atype: "ldapSyntaxes".to_string(),
             vals: vec![
+                b"( 1.3.6.1.1.16.1 DESC 'UUID' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.3 DESC 'Attribute Type Description' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.12 DESC 'Distinguished Name' )".to_vec(),
                 b"( 1.3.6.1.4.1.1466.115.121.1.15 DESC 'Directory String' )".to_vec(),
                 b"( 1.3.6.1.4.1.1466.115.121.1.24 DESC 'Generalized Time' )".to_vec(),
                 b"( 1.3.6.1.4.1.1466.115.121.1.27 DESC 'Integer' )".to_vec(),
                 b"( 1.3.6.1.4.1.1466.115.121.1.28 DESC 'JPEG' X-NOT-HUMAN-READABLE 'TRUE' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.34 DESC 'Name And Optional UID' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.37 DESC 'Object Class Description' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.38 DESC 'OID' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.54 DESC 'LDAP Syntax Description' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.58 DESC 'Substring Assertion' )".to_vec(),
+                ],
+           },
+           LdapPartialAttribute {
+            atype: "matchingRules".to_string(),
+            vals: vec![
+                b"( 1.3.6.1.1.16.2 NAME 'UUIDMatch' SYNTAX 1.3.6.1.1.16.1 )".to_vec(),
+                b"( 1.3.6.1.1.16.3 NAME 'UUIDOrderingMatch' SYNTAX 1.3.6.1.1.16.1 )".to_vec(),
+                b"( 2.5.13.0 NAME 'objectIdentifierMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )".to_vec(),
+                b"( 2.5.13.1 NAME 'distinguishedNameMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 )".to_vec(),
+                b"( 2.5.13.2 NAME 'caseIgnoreMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
+                b"( 2.5.13.4 NAME 'caseIgnoreSubstringsMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.58 )".to_vec(),
+                b"( 2.5.13.23 NAME 'uniqueMemberMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.34 )".to_vec(),
+                b"( 2.5.13.27 NAME 'generalizedTimeMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                b"( 2.5.13.28 NAME 'generalizedTimeOrderingMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                b"( 2.5.13.30 NAME 'objectIdentifierFirstComponentMatch' SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )".to_vec(),
                 ],
            },
            LdapPartialAttribute {
             atype: "attributeTypes".to_string(),
             vals: {
                 let hardcoded_attributes = [
-                    b"( 2.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
-                    b"( 2.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
-                    b"( 2.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
-                    b"( 2.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                    b"( 0.9.2342.19200300.100.1.1 NAME ( 'uid' 'userid' 'user_id' ) DESC 'RFC4519: user identifier' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{256} SINGLE-VALUE NO-USER-MODIFICATION )".to_vec(),
+                    b"( 1.3.6.1.1.16.4 NAME ( 'entryUUID' 'uuid' ) DESC 'UUID of the entry' EQUALITY UUIDMatch ORDERING UUIDOrderingMatch SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 1.3.6.1.4.1.1466.101.120.16 NAME 'ldapSyntaxes' DESC 'RFC4512: LDAP syntaxes' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.54 USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.4.0 NAME 'objectClass' DESC 'RFC4512: object classes of the entity' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )".to_vec(),
+                    b"( 2.5.4.3 NAME ( 'cn' 'commonName' 'display_name' ) DESC 'RFC4519: common name(s) for which the entity is known by' SUP name SINGLE-VALUE )".to_vec(),
+                    b"( 2.5.4.4 NAME ( 'sn' 'surname' 'last_name' ) DESC 'RFC2256: last (family) name(s) for which the entity is known by' SUP name SINGLE-VALUE )".to_vec(),
+                    b"( 2.5.4.11 NAME ( 'ou' 'organizationalUnitName' ) DESC 'RFC2256: organizational unit this object belongs to' SUP name )".to_vec(),
+                    b"( 2.5.4.41 NAME 'name' DESC 'RFC4519: common supertype of name attributes' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{32768} )".to_vec(),
+                    b"( 2.5.4.49 NAME 'distinguishedName' DESC 'RFC4519: common supertype of DN attributes' EQUALITY distinguishedNameMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 )".to_vec(),
+                    b"( 2.5.4.50 NAME ( 'uniqueMember' 'member' ) DESC 'RFC2256: unique member of a group' EQUALITY uniqueMemberMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.34 )".to_vec(),
+                    b"( 2.5.18.1 NAME ( 'createTimestamp' 'creation_date' ) DESC 'RFC4512: time which object was created' EQUALITY generalizedTimeMatch ORDERING generalizedTimeOrderingMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.18.2 NAME 'modifyTimestamp' DESC 'RFC4512: time which object was last modified' EQUALITY generalizedTimeMatch ORDERING generalizedTimeOrderingMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.21.5 NAME 'attributeTypes' DESC 'RFC4512: attribute types' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.3 USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.21.6 NAME 'objectClasses' DESC 'RFC4512: object classes' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.37 USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.21.9 NAME 'structuralObjectClass' DESC 'RFC4512: structural object class of entry' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 10.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
+                    b"( 10.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
+                    b"( 10.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
+                    b"( 10.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
                 ];
                 let num_hardcoded_attributes = hardcoded_attributes.len();
                 hardcoded_attributes.into_iter().chain(
                     ldap_schema_description
-                        .formatted_attribute_list(num_hardcoded_attributes)
+                        .formatted_attribute_list(
+                            num_hardcoded_attributes,
+                            vec!["creation_date", "display_name", "last_name", "user_id", "uuid"]
+                        )
                 ).collect()
             }
            },
@@ -485,7 +526,7 @@ mod tests {
         };
 
         let attrs = &search_result_entry.attributes;
-        assert_eq!(attrs.len(), 9);
+        assert_eq!(attrs.len(), 10);
         assert_eq!(search_result_entry.dn, "cn=Subschema".to_owned());
 
         assert_eq!(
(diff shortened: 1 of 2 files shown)


crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
+                b"( 1.3.6.1.4.1.1466.115.121.1.38 DESC 'OID' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.54 DESC 'LDAP Syntax Description' )".to_vec(),
+                b"( 1.3.6.1.4.1.1466.115.121.1.58 DESC 'Substring Assertion' )".to_vec(),
+                ],
+           },
+           LdapPartialAttribute {
+            atype: "matchingRules".to_string(),
+            vals: vec![
+                b"( 1.3.6.1.1.16.2 NAME 'UUIDMatch' SYNTAX 1.3.6.1.1.16.1 )".to_vec(),
+                b"( 1.3.6.1.1.16.3 NAME 'UUIDOrderingMatch' SYNTAX 1.3.6.1.1.16.1 )".to_vec(),
```
nitnelave, line 223: I don't think we support uuid ordering. Also, they're random.

crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
            },
            LdapPartialAttribute {
             atype: "attributeTypes".to_string(),
             vals: {
                 let hardcoded_attributes = [
-                    b"( 2.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
-                    b"( 2.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
-                    b"( 2.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
-                    b"( 2.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                    b"( 0.9.2342.19200300.100.1.1 NAME ( 'uid' 'userid' ) DESC 'RFC4519: user identifier' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{256} )".to_vec(),
```
nitnelave: Single value, no user modification?

crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
+                    b"( 1.3.6.1.1.16.4 NAME 'entryUUID' DESC 'UUID of the entry' EQUALITY UUIDMatch ORDERING UUIDOrderingMatch SYNTAX 1.3.6.1.1.16.1 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 1.3.6.1.4.1.1466.101.120.16 NAME 'ldapSyntaxes' DESC 'RFC4512: LDAP syntaxes' EQUALITY objectIdentifierFirstComponentMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.54 USAGE directoryOperation )".to_vec(),
+                    b"( 2.5.4.0 NAME 'objectClass' DESC 'RFC4512: object classes of the entity' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 )".to_vec(),
+                    b"( 2.5.4.3 NAME ( 'cn' 'commonName' ) DESC 'RFC4519: common name(s) for which the entity is known by' SUP name )".to_vec(),
```
nitnelave: Single value

crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
+                    b"( 2.5.4.4 NAME ( 'sn' 'surname' ) DESC 'RFC2256: last (family) name(s) for which the entity is known by' SUP name )".to_vec(),
```
nitnelave: Single value

crates/ldap/src/search.rs
nitnelave: I would be careful with the "SUP name": these are not substring searchable. Basically only uid and email are

Which means that mail and cn must be treated differently to declare that

crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
+                    b"( 2.5.4.11 NAME ( 'ou' 'organizationalUnitName' ) DESC 'RFC2256: organizational unit this object belongs to' SUP name )".to_vec(),
+                    b"( 2.5.4.31 NAME 'member' DESC 'RFC2256: member of a group' SUP distinguishedName )".to_vec(),
```
nitnelave: For us, member is effectively an alias for uniqueMember, I think we should use that as the default

crates/ldap/src/search.rs
```diff
@@ -202,20 +202,59 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
...
+                    b"( 2.5.4.41 NAME 'name' DESC 'RFC4519: common supertype of name attributes' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{32768} )".to_vec(),
+                    b"( 2.5.4.49 NAME 'distinguishedName' DESC 'RFC4519: common supertype of DN attributes' EQUALITY distinguishedNameMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 )".to_vec(),
+                    b"( 2.5.4.50 NAME 'uniqueMember' DESC 'RFC2256: unique member of a group' EQUALITY uniqueMemberMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.34 )".to_vec(),
+                    b"( 2.5.18.1 NAME 'createTimestamp' DESC 'RFC4512: time which object was created' EQUALITY generalizedTimeMatch ORDERING generalizedTimeOrderingMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
```
nitnelave: You can't use ordering operations on our timestamps (random LLDAP limitation)

nitnelave (commented):
So, LGTM, but most of my comments revolve around one question: is the attribute definition a declaration of what the server supports, or a hint for the client of how they should handle the values?

For instance, when we don't support ordering, does it make sense to declare that the values should be ordered like so?

T0byV:
> So, LGTM, but most of my comments revolve around one question: is the attribute definition a declaration of what the server supports, or a hint for the client of how they should handle the values?
> 
> For instance, when we don't support ordering, does it make sense to declare that the values should be ordered like so?

I'm not entirely certain regarding how different clients would handle the attribute definitions and I could not easily find if there is a general consensus on how attribute's are handled. However I do agree that it might be 'misleading' to specify that the server can e.g. do ordering while it can't. The issue here is that if we for example look at the RFC page for `entryUUID` (https://www.rfc-editor.org/rfc/rfc4530#section-2.4), it shows that the definition of `1.3.6.1.1.16.4` already includes the ORDERING in the standard.

I see a couple ways to go from here:
- We could remove all the 'extra' properties of an attribute like ordering, no user modification etc. but that does mean that we use the identifier for e.g. entryUUID:  1.3.6.1.1.16.4 without correctly implementing it according to RFC
- We could leave it as is and assume that clients will not care about ordering and such, and if it does seem to be an issue we wait for someone to report it.
- We could acquire our own OID range(https://www.iana.org/assignments/enterprise-numbers/assignment/apply/) so we can safely redefine attributes with properties we _do_ support, without violating any RFC definitions.

I've not worked much with clients that use the subschema and I'm not sure where I should look to acquire information regarding this, so maybe we could inquire someone regarding this matter?

crates/ldap/src/search.rs
T0byV: The SUP part is also straight from https://datatracker.ietf.org/doc/html/rfc4519#section-2.3, shall we just keep the `SUP name` for now?

crates/ldap/src/search.rs
T0byV: I've change the `uniqueMember` attribute to `( 'uniqueMember' 'member' )` so it's now aliased.

crates/ldap/src/search.rs
nitnelave: Yeah, since we want to declare generous client permissions, I don't mind keeping it. I was just careful about declaring commonName as searchable, but that's fine.

crates/ldap/src/search.rs
```diff
@@ -530,47 +568,89 @@ mod tests {
...
+                    b"( 2.5.21.9 NAME 'structuralObjectClass' DESC 'RFC4512: structural object class of entry' EQUALITY objectIdentifierMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.38 SINGLE-VALUE NO-USER-MODIFICATION USAGE directoryOperation )".to_vec(),
+                    b"( 10.0 NAME 'String' SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )".to_vec(),
+                    b"( 10.1 NAME 'Integer' SYNTAX 1.3.6.1.4.1.1466.115.121.1.27 )".to_vec(),
+                    b"( 10.2 NAME 'JpegPhoto' SYNTAX 1.3.6.1.4.1.1466.115.121.1.28 )".to_vec(),
+                    b"( 10.3 NAME 'DateTime' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )".to_vec(),
+                    b"( 10.19 NAME 'avatar' DESC 'LLDAP: builtin attribute' SUP JpegPhoto )".to_vec(),
+                    b"( 10.20 NAME 'creation_date' DESC 'LLDAP: builtin attribute' SUP DateTime )"
                         .to_vec(),
-                    b"( 2.6 NAME 'display_name' DESC 'LLDAP: builtin attribute' SUP String )"
+                    b"( 10.21 NAME 'display_name' DESC 'LLDAP: builtin attribute' SUP String )"
```
nitnelave: That's an alias for cn, we should declare it as such.

crates/ldap/src/search.rs
```diff
@@ -530,47 +568,89 @@ mod tests {
...
                         .to_vec(),
-                    b"( 2.7 NAME 'first_name' DESC 'LLDAP: builtin attribute' SUP String )"
+                    b"( 10.22 NAME 'first_name' DESC 'LLDAP: builtin attribute' SUP String )"
                         .to_vec(),
-                    b"( 2.8 NAME 'last_name' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.9 NAME 'mail' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.10 NAME 'user_id' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.11 NAME 'uuid' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.12 NAME 'creation_date' DESC 'LLDAP: builtin attribute' SUP DateTime )"
+                    b"( 10.23 NAME 'last_name' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
```
nitnelave: And that's sn

crates/ldap/src/search.rs
```diff
@@ -530,47 +568,89 @@ mod tests {
...
+                    b"( 10.24 NAME 'mail' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
+                    b"( 10.25 NAME 'user_id' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
+                    b"( 10.26 NAME 'uuid' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
```
nitnelave: that's entryuuid

crates/ldap/src/search.rs
```diff
@@ -530,47 +568,89 @@ mod tests {
...
+                    b"( 10.22 NAME 'first_name' DESC 'LLDAP: builtin attribute' SUP String )"
                         .to_vec(),
-                    b"( 2.8 NAME 'last_name' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.9 NAME 'mail' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.10 NAME 'user_id' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.11 NAME 'uuid' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
-                    b"( 2.12 NAME 'creation_date' DESC 'LLDAP: builtin attribute' SUP DateTime )"
+                    b"( 10.23 NAME 'last_name' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
+                    b"( 10.24 NAME 'mail' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
+                    b"( 10.25 NAME 'user_id' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
```
nitnelave: That's uid

crates/ldap/src/search.rs
```diff
@@ -530,47 +568,89 @@ mod tests {
...
+                    b"( 10.26 NAME 'uuid' DESC 'LLDAP: builtin attribute' SUP String )".to_vec(),
+                    b"( 10.27 NAME 'creation_date' DESC 'LLDAP: builtin attribute' SUP DateTime )"
```
nitnelave: That's createTimestamp

nitnelave (commented):
LGTM, but let's remove the attributes that are aliases of hardcoded attributes.

crates/ldap/src/search.rs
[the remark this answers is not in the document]
nitnelave: Disregard.

T0byV:
> LGTM, but let's remove the attributes that are aliases of hardcoded attributes.

This is not as trivial as it seems, the attributes that are to be aliased are defined in the `crates/domain/src/public_schema.rs` file and not all of the attributes that are defined there should be left out (like first_name, mail etc should stay in). Do you have anything in mind to 'filter out' the attributes that should be an alias? I think it's getting ugly quite quickly if we filter them out and just hardcode them as an alias of the RFC attribute, however I also can't think of a better way that does not require a complete revamp of the code:)

nitnelave:
I think as a first step it's fine to have a hardcoded list of attributes to remove/deduplicate. Eventually everything will be properly supported with aliases.

T0byV:
Dedup is implemented now. Just added an extra argument to the format function that can exclude a list of attributes.

nitnelave:
LGTM! Can you rebase and squash?
With "ldap" as the commit tag instead of "fix"

nitnelave:
> LGTM! Can you rebase and squash?
> With "ldap" as the commit tag instead of "fix"

@T0byV can you squash to a single commit?

PR #1205, opened 2025-07-08
docs: move authelia configuration to markdown file

hendrik1120:
Hi,

this is a bit of an overhaul of the existing authelia configuration for LLDAP.

I recently moved and split up the authelia ldap reference for specific ldap providers to the integration section.
During that, I noticed that the user filter already is configured to allow signing in with username and email.
This was really the main issue with the existing documentation on this side, the other things are mostly QOL.

I removed the configuration options which were already the authelia default to focus more on the LLDAP specific things, which are because of the implementation template quite minimal.

Before removing the `additional_users_dn` option, I saw that it was introduced by #1182. It is in my opinion highly likely that this was an issue with the authelia helm chart during that time. See [PR](https://github.com/authelia/chartrepo/pull/319/). The value that was set is the current default for the LLDAP implementation template.

Password change is a relatively new feature introduced in v4.39 so it should be disabled as well if users don't wish to use authelia as a password manager.

I also saw the new PR to disable user logins, at the moment I am not quite sure if that can be integrated in the authelia configuration template for LLDAP without a breaking change, the updated user filter might have to be documented separately.

Sorry for the long explainer, we can also chat on discord if you have any questions or feedback.


example_configs/authelia_config.yml
```diff
@@ -1,35 +0,0 @@
...
-    base_dn: dc=example,dc=com
-    # The username and password of the bind user.
-    # "bind_user" should be the username you created for authentication with the "lldap_strict_readonly" permission. It is not recommended to use an actual admin account here.
-    # If you are configuring Authelia to change user passwords, then the account used here needs the "lldap_password_manager" permission instead.
-    user: uid=bind_user,ou=people,dc=example,dc=com
-    additional_users_dn: ou=people
-    # Password can also be set using a secret: https://www.authelia.com/configuration/methods/secrets/
-    password: "REPLACE_ME"
-
-    # Optional: Setup TLS if you've enabled LDAPS
```
nitnelave, line 29: What about this section? Should we port it over?

hendrik1120: I think you mean the TLS config, the review somehow marks lines 26-29.
The authelia default is:
```yml
tls:
  skip_verify: false
  minimum_version: 'TLS1.2'
  maximum_version: 'TLS1.3'
```
If you have a proper and modern TLS setup, you shouldn't need to configure anything extra. However, in reality that's still pretty uncommon. What do you think of adding the `skip_verify` option back for self-signed certs? Or should it be more clear about the TLS1.2 minimum?

nitnelave: LLDAP is quite modern in terms of TLS, I think it doesn't even accept TLS 1.2

But yes, I was mostly thinking of the skip_verify for self-signed certs

hendrik1120: Yeah, ok I will add it, one sec. I am using LLDAPS with a self-signed cert at the moment and I haven't touched the minimum.

nitnelave (approved):
Thanks!

PR #1216, opened 2025-07-16
WIP: fix: "Creation date" in user forum is displayed as a date

selfhoster1312:
Previously, when clicking on user details as an admin, i would see `creation_date` with the complete stringy timestamp.

Capture d’écran du 2025-07-02 13-47-51

Now it's displayed as in the user list view:

image

This is my first time working with graphql/wasm/yew, so i'm having doubts about the proper way to do things.

- the `User` type is defined as a type alias to `get_user_details::GetUserDetailsUser` but i have no idea where this is defined??? is that somehow an alias to `lldap_domain::types::User`?
- it seems that `User::attributes` is stringy but i can access `User::creation_date`… for the typed version (i also didn't find a way to reparse an arbitrary Attribute/AttributeValue from the stringy representation
- i don't understand why/how we use the arbitrary `attributes` in this template when `user_table` uses typed attributes… is this because of custom extra attributes? or to simplify form/view logic (if so, it doesn't make it easier to read from the outside)

I would be happy to contribute documentation if i understand something finally. I will open a separate issue about frontend documentation. :smile:

### DIFF
--- a/app/src/components/user_details_form.rs
+++ b/app/src/components/user_details_form.rs
@@ -104,7 +104,7 @@ impl Component for UserDetailsForm {
             if can_edit(a) {
                 get_custom_attribute_input(a, &self.user.attributes)
             } else {
-                get_custom_attribute_static(a, &self.user.attributes)
+                get_custom_attribute_static(a, &self.user)
             }
         };
         html! {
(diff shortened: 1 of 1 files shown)


nitnelave:
I don't have that much time to review tonight, but I'll give you a couple of pointers to hopefully help you along.

Data from server replies is generated by proc macros based on the GraphQL schema at the root. That's where you'll find the definition of user. 

As for the attributes, I tried to consolidate as many as possible in the variable-size "attributes" field that was added after the structural attributes like creation date. As a result the data is duplicated. But you should be able to parse the date from the string, get the time and display it correctly. If the string only contains the date and no time, that would surprise me, and that would be a backend bug. It may just be a question of changing how we format generic attributes of datetime type.

nitnelave:
As for documentation, I hope you found the docs/ folder at the top. Sadly, the code architecture part is sorely out of date after splitting the crates apart.

nitnelave:
Closed in favor of https://github.com/lldap/lldap/pull/1224

PR #1217, opened 2025-07-16
meta: Fix cargo clippy failures (format strings)

selfhoster1312:
I noticed in #1216 clippy was failing, not only because of the PR changes.

In this PR i apply clippy suggestions to make it happy.

In the future we may want to pin clippy/fmt to a specific version to avoid breaking CI due to the Github Actions upgrading to a newer release.


nitnelave (approved):

nitnelave:
Thanks! I actually like the fact that it auto updates, prompting PRs like this. It's not a blocking check, so I don't mind it breaking spuriously from time to time.

PR #1218, opened 2025-07-16
bootstrap: allow to give password from a file

ibizaman:
In the bootstrap script, instead of embedding the password in the config file directly, allow to pass a path to a file which contains the password.

This was tested in a NixOS draft PR https://github.com/NixOS/nixpkgs/pull/425923.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -691,8 +691,8 @@ main() {
 
   local user_config=''
   while read -r user_config; do
-    local field='' id='' email='' displayName='' firstName='' lastName='' avatar_file='' avatar_url='' gravatar_avatar='' weserv_avatar='' password=''
-    for field in 'id' 'email' 'displayName' 'firstName' 'lastName' 'avatar_file' 'avatar_url' 'gravatar_avatar' 'weserv_avatar' 'password'; do
+    local field='' id='' email='' displayName='' firstName='' lastName='' avatar_file='' avatar_url='' gravatar_avatar='' weserv_avatar='' password='' password_file=''
+    for field in 'id' 'email' 'displayName' 'firstName' 'lastName' 'avatar_file' 'avatar_url' 'gravatar_avatar' 'weserv_avatar' 'password' 'password_file'; do
       declare "$field"="$(printf '%s' "$user_config" | jq --raw-output --arg field "$field" '.[$field]')"
     done
     printf -- '\n--- %s ---\n' "$id"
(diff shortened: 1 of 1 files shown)


nitnelave (approved):

ibizaman:
Thanks @nitnelave for the very fast review and merge :)

nitnelave:
@ibizaman you caught me at an extremely lucky moment, these days I usually only review PRs every few days. You just happened to send this while I was reviewing something else!

ibizaman:
Luck is always part of it. Thanks nonetheless :)

scripts/bootstrap.sh
```diff
@@ -691,16 +691,18 @@ main() {
...
       declare "$field"="$(printf '%s' "$user_config" | jq --raw-output --arg field "$field" '.[$field]')"
     done
     printf -- '\n--- %s ---\n' "$id"
 
     create_update_user "$id" "$email" "$displayName" "$firstName" "$lastName" "$avatar_file" "$avatar_url" "$gravatar_avatar" "$weserv_avatar"
     redundant_users="$(printf '%s' "$redundant_users" | jq --compact-output --arg id "$id" '. - [$id]')"
 
-    if [[ "$password" != 'null' ]] && [[ "$password" != '""' ]]; then
+    if [[ "$password_file" != 'null' ]] && [[ "$password_file" != '""' ]]; then
+      "$LLDAP_SET_PASSWORD_PATH" --base-url "$LLDAP_URL" --token "$TOKEN" --username "$id" --password "$(cat $password_file)"
```
bjornfor, line 704: This looks like it's leaking the password in process listings (which are visible to _all_ processes). Adding a --password-file option to $LLDAP_SET_PASSWORD_PATH, to read the password directly from the file, should fix that.

nitnelave: Why not put it in the env instead, then?

bjornfor: What feature does that bring? (I'm not familiar with the code, so not sure why you'd want that.) I would avoid secrets in env vars if possible, to reduce risk of leakage.

ibizaman: I added this because I wanted to avoid to embed the password in the config file directly. I wanted the password to be a reference to a file instead.

This comes from wanting to generate the config file and embedding the passwords at generation time leaks the passwords in other parts of the system.

I’m not sure I follow what your proposed fix would be, sorry.

nitnelave: If we have an env var, we can write `PASSWORD=$(cat password_file) set_password --user ...`

That's simpler and more flexible

ibizaman: Oh I see what you mean @bjornfor. You mean having the binary itself read the file. I’ll create a PR to do that. It makes much more sense indeed.

nitnelave: Of the two, I'd prefer the env solution, unless you have compelling arguments

ibizaman: No arguments. The only thing is the comment above about avoiding leakage. I can do either.

nitnelave: I'd need a more specific attack scenario than "avoid leakage". Sure, if you set it in your global env and start other processes they'll see it, but that's not what's proposed here

PR #1228, opened 2025-07-21
bootstrap: do not leak password in process list

ibizaman:
As discussed [in a comment](https://github.com/lldap/lldap/pull/1218#discussion_r2218854469), we should use a variable to hold the password to avoid leaking it in the process list.

This was tested in https://github.com/NixOS/nixpkgs/pull/425923


scripts/bootstrap.sh
```diff
@@ -707,7 +707,8 @@ main() {
     redundant_users="$(printf '%s' "$redundant_users" | jq --compact-output --arg id "$id" '. - [$id]')"
 
     if [[ "$password_file" != 'null' ]] && [[ "$password_file" != '""' ]]; then
-      "$LLDAP_SET_PASSWORD_PATH" --base-url "$LLDAP_URL" --token "$TOKEN" --username "$id" --password "$(cat $password_file)"
+      local PASSWORD="$(cat $password_file)"
+      "$LLDAP_SET_PASSWORD_PATH" --base-url "$LLDAP_URL" --token "$TOKEN" --username "$id" --password "$PASSWORD"
```
nitnelave: I don't think that solves the issue, the password is still visible when listing the processes. It'll require a code change to read the env variable.

ibizaman: Isn't the line 712 prone to this issue too then? The one with the `$password` variable.

nitnelave: It is indeed. It's a new concern.

I'm honestly not completely convinced it's a valid security concern, but if you want to implement it I won't stand in your way.

ibizaman: Then I think the best way is to introduce a new flag in the `lldap_set_password` app that gets a path to a file directly, like suggested before.

nitnelave: I'd argue for reading from an env variable, again

ibizaman: Something like this?

```bash
local LLDAP_USER_PASSWORD="$(cat $password_file)"
"$LLDAP_SET_PASSWORD_PATH" --base-url "$LLDAP_URL" --token "$TOKEN" --username "$id"
```

nitnelave:
```bash
LLDAP_USER_PASSWORD="$(cat $password_file)" "$LLDAP_SET_PASSWORD_PATH" --base-url "$LLDAP_URL" --token "$TOKEN" --username "$id"
```

ibizaman: Eh it seems I can't bash in my head. Anyway, I'm on it.

set-password/src/main.rs
```diff
@@ -27,7 +29,7 @@ pub struct CliOpts {
     #[clap(short, long)]
     pub username: String,
 
-    /// New password for the user.
+    /// New password for the user. Can also be passed as the environment variable LLDAP_USER_PASSWORD.
     #[clap(short, long)]
     pub password: String,
```
nitnelave:
```suggestion
    pub password: Option<String>,
```

Otherwise I think the argument is required. Have you tried it as is? I don't think it works.

ibizaman: No sorry I forgot to mention. I didn't test yet but needed to push to be able to test. I'll convert the PR to a draft in the meantime and ping you when it's ready.

nitnelave (approved):
Thanks!

PR #1230, opened 2025-07-22
meta: Minimum Supported Rust Version MSRV 1.85

selfhoster1312:
Make sure code still compiles on rust 1.85 on pull requests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -108,3 +108,24 @@ jobs:
           fail_ci_if_error: true
           codecov_yml_path: .github/codecov.yml
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  msrv:
+    name: cargo msrv verify
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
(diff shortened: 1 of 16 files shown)


.github/workflows/rust.yml
```diff
@@ -108,3 +108,24 @@ jobs:
           fail_ci_if_error: true
           codecov_yml_path: .github/codecov.yml
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  msrv:
```
nitnelave, line 112: I don't think we need this extra check, since we have https://github.com/lldap/lldap/blob/main/.github/workflows/Dockerfile.dev#L2

We're already compiling with 1.85.

I'd rather avoid having another random tool that potentially installs an entire toolchain and recompiles the entire project. If we want to check that the version is the same, I'd much rather have 95% of the benefits by running a simple shell script that greps both files for the same version, checking that they match.

nitnelave (changes requested):
LGTM apart from the extra check.

nitnelave:
Sorry for the late review!

PR #1232, opened 2025-07-24
bootstrap: make password_file a standard custom attribute

ibizaman:
Otherwise the bootstrap script tries to create the password_file as a custom attribute which fails since it's not in the schema. And anyway, it shouldn't be in the schema.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -440,7 +440,7 @@ extract_custom_group_attributes() {
 }
 
 extract_custom_user_attributes() {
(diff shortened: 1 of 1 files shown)


nitnelave:
Sorry for the long review delay!

nitnelave (approved):

ibizaman:
@nitnelave  no worries at all. Thanks for the easy reviews on the PRs :)

nitnelave:
Thanks for the short, easy to review PRs!

PR #1239, opened 2025-08-07
Add login_enabled database field for user enable/disable functionality

coolguy1771:
## Summary

  This PR introduces the database foundation for user enable/disable functionality by adding a `login_enabled` boolean field to the users table. This is the first in a series of PRs to implement the complete feature as discussed in #1193.

  ## Changes

  - **Database Migration**: Added migration v11 to create the `login_enabled` column
    - Defaults to `true` for backward compatibility with existing users
    - Includes a check for conflicting custom attributes with a warning message

  - **Model Updates**:
    - Updated the User model to include the `login_enabled` field
    - Ensured new users are created with `login_enabled: true` by default

  ## Migration Notes

  - The migration checks if a custom attribute named "login_enabled" already exists and warns users if found
  - Existing users will have `login_enabled` set to `true` by default, ensuring no disruption to current functionality
  - The field is a separate column (not a custom attribute) for better performance and type safety

  ## Testing

  - [x] Database migration runs successfully
  - [x] Code compiles without errors
  - [x] New users are created with `login_enabled: true`

  ## Related Issues

  Addresses part of #1193

### DIFF
--- a/.github/workflows/docker-build-static.yml
+++ b/.github/workflows/docker-build-static.yml
@@ -87,7 +87,7 @@ jobs:
       image: lldap/rust-dev:latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@v5.0.0
       - uses: actions/cache@v4
         with:
           path: |
@@ -132,7 +132,7 @@ jobs:
         CARGO_HOME: ${GITHUB_WORKSPACE}/.cargo
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@v5.0.0
       - uses: actions/cache@v4
         with:
           path: |
@@ -300,7 +300,7 @@ jobs:
 
     steps:
        - name: Checkout scripts
-         uses: actions/checkout@v4.2.2
+         uses: actions/checkout@v5.0.0
          with:
            sparse-checkout: 'scripts'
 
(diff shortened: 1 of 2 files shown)


crates/sql-backend-handler/src/sql_migrations.rs
```diff
@@ -166,7 +167,13 @@ pub(crate) async fn upgrade_to_v1(pool: &DbConnection) -> std::result::Result<()
                 .col(ColumnDef::new(Users::PasswordHash).blob())
                 .col(ColumnDef::new(Users::TotpSecret).string_len(64))
                 .col(ColumnDef::new(Users::MfaType).string_len(64))
-                .col(ColumnDef::new(Users::Uuid).string_len(36).not_null()),
+                .col(ColumnDef::new(Users::Uuid).string_len(36).not_null())
```
nitnelave, line 170: Oh, I was commenting on the wrong PR. Well then :)

Please don't change the v1, put all the changes in the v11 migration

crates/sql-backend-handler/src/sql_migrations.rs
```diff
@@ -1112,6 +1119,55 @@ async fn migrate_to_v10(transaction: DatabaseTransaction) -> Result<DatabaseTran
...
+                ),
+        ),
+    )
+    .one(&transaction)
+    .await?;
+
+    if let Some(check) = existing_attribute
+        && check.count > 0
+    {
+        warn!(
```
nitnelave, line 1149: This should be a hard error preventing upgrading.

nitnelave (approved):

nitnelave (changes requested):
There are still unaddressed comments

PR #1240, opened 2025-08-07
Replace JWT blacklist with user login_enabled validation

coolguy1771:
## Summary

  This PR replaces the JWT blacklist mechanism with real-time validation of the user's `login_enabled` status.

  ## Changes

  - **Removed JWT Blacklist**:
    - Removed the in-memory JWT blacklist that was incompatible with high-availability deployments

  - **Added User Status Validation**:
    - Added `is_user_login_enabled` method to check user status from the database
    - Modified `check_if_token_is_valid` to be async and query the user's status on every JWT validation
    - Returns "User account is disabled" error when a disabled user attempts to use a JWT

  - **Updated JWT Validation Flow**:
    - JWT validation now queries the database on every request to check if the user is enabled
    - Immediate access revocation when an account is disabled (no need to wait for JWT expiry)

  ## Testing

  - [x] Code compiles without errors
  - [x] JWT validation properly checks user status

  ## Related Issues

  Addresses part of #1193

  ## Dependencies
  
  Needs #1239 to be merged first

### DIFF
--- a/server/src/tcp_backend_handler.rs
+++ b/server/src/tcp_backend_handler.rs
@@ -17,7 +17,7 @@ pub trait TcpBackendHandler: Sync {
     async fn check_token(&self, refresh_token_hash: u64, user: &UserId) -> Result<bool>;
     async fn blacklist_jwts(&self, user: &UserId) -> Result<HashSet<u64>>;
     async fn delete_refresh_token(&self, refresh_token_hash: u64) -> Result<()>;
-    
+
     /// Check if a user's login is enabled
     async fn is_user_login_enabled(&self, user: &UserId) -> Result<bool>;
 


crates/sql-backend-handler/src/sql_migrations.rs
```diff
@@ -166,7 +167,13 @@ pub(crate) async fn upgrade_to_v1(pool: &DbConnection) -> std::result::Result<()
                 .col(ColumnDef::new(Users::PasswordHash).blob())
                 .col(ColumnDef::new(Users::TotpSecret).string_len(64))
                 .col(ColumnDef::new(Users::MfaType).string_len(64))
-                .col(ColumnDef::new(Users::Uuid).string_len(36).not_null()),
+                .col(ColumnDef::new(Users::Uuid).string_len(36).not_null())
+                .col(
```
nitnelave, line 171: Don't change an earlier migration, do all the work in the new one.

crates/sql-backend-handler/src/sql_migrations.rs
```diff
@@ -1112,6 +1119,55 @@ async fn migrate_to_v10(transaction: DatabaseTransaction) -> Result<DatabaseTran
...
+                ),
+        ),
+    )
+    .one(&transaction)
+    .await?;
+
+    if let Some(check) = existing_attribute
+        && check.count > 0
+    {
+        warn!(
```
nitnelave, line 1149: I think this should be a hard error. You can't upgrade until you rename/delete this attribute.

crates/sql-backend-handler/src/sql_migrations.rs
```diff
@@ -1112,6 +1119,55 @@ async fn migrate_to_v10(transaction: DatabaseTransaction) -> Result<DatabaseTran
...
+            r#"Found existing custom user attribute "login_enabled". 
+                This will not conflict with the new system field, but you may want to 
+                migrate your data and remove the custom attribute to avoid confusion."#
+        );
+    }
+
+    transaction
+        .execute(
```
nitnelave, line 1157: We'll also want an index on that field.

server/src/auth_service.rs
```diff
@@ -457,17 +457,24 @@ async fn opaque_register_start<Backend>(
...
     use actix_web::FromRequest;
     let inner_payload = &mut payload.into_inner();
-    let validation_result = BearerAuth::from_request(&request, inner_payload)
-        .await
-        .ok()
-        .and_then(|bearer| check_if_token_is_valid(&data, bearer.token()).ok())
-        .ok_or_else(|| {
-            TcpError::UnauthorizedError("Not authorized to change the user's password".to_string())
-        })?;
+    let validation_result = match BearerAuth::from_request(&request, inner_payload).await {
```
nitnelave, line 464: What is this change about? Is it some LLM over-eagerness?

server/src/auth_service.rs
```diff
@@ -606,12 +613,20 @@ pub(crate) fn check_if_token_is_valid<Backend: BackendHandler>(
...
-    if state.jwt_blacklist.read().unwrap().contains(&jwt_hash) {
-        return Err(ErrorUnauthorized("JWT was logged out"));
+
+    // Check if user is enabled
+    let user_id = UserId::new(&token.claims().user);
+    if !state
+        .get_tcp_handler()
+        .is_user_login_enabled(&user_id)
+        .await
+        .map_err(|_| ErrorUnauthorized("Could not verify user status"))?
```
nitnelave, line 623: Propagate the error message here, so we know _why_ we couldn't verify the user status

nitnelave: And maybe have a match for EntityNotFound with a custom error message.

nitnelave (changes requested):
Looks good! A few comments here and there, but that's the right approach.

Sorry for the slow review, life gets in the way :)

PR #1241, opened 2025-08-09
Infinite security patch

kazhuravlev:


nitnelave:
Erm... What is this PR?

kazhuravlev:
Haha) sorry for clickbait, but here i was try to remove words that looks like a "bad language" in Russian.

kazhuravlev:
@nitnelave I’m pretty sure you can find other words without reducing the quality of the documentation 😂

nitnelave:
I don't think it's worth changing, no one is going to get offended by this.

kazhuravlev:
I understand your position. The issue here is not about someone “getting offended,” but rather that using obscene language in code examples reduces the professional appearance of the project and may prevent it from being adopted in companies.

This is a simple, neutral fix (e.g., replacing it with `my-token`) that removes unnecessary risks and does not harm the project in any way.

PR #1242, opened 2025-08-10
bootstrap: fine grained cleanup

ibizaman:
This allows to have groups managed declaratively while users are managed through the UI.


scripts/bootstrap.sh
```diff
@@ -12,6 +12,9 @@ USER_CONFIGS_DIR="${USER_CONFIGS_DIR:-/bootstrap/user-configs}"
 GROUP_CONFIGS_DIR="${GROUP_CONFIGS_DIR:-/bootstrap/group-configs}"
 LLDAP_SET_PASSWORD_PATH="${LLDAP_SET_PASSWORD_PATH:-/app/lldap_set_password}"
 DO_CLEANUP="${DO_CLEANUP:-false}"
+DO_CLEANUP_USERS="${DO_CLEANUP_USERS:-false}"
```
nitnelave: Can you use DO_CLEANUP as the default value? That should simplify the checks below

nitnelave (approved):

scripts/bootstrap.sh
ibizaman: Great idea!

nitnelave (approved):

PR #1245, opened 2025-08-12
Improve bootstrap.sh and documentation for use with Kubernetes

jonas-resch:
- Pass `--password` flag to `lldap_set_password` (it doesn't work by passing the ENV variable anymore)
- Add a parameter `LLDAP_FORCE_SET_PASSWORD` to only sync a password once when creating the user. It won't be overwritten afterwards
- Improve documentation for use with Kubernetes (fix Job and add example yaml for config map)

### DIFF
--- a/example_configs/bootstrap/bootstrap.md
+++ b/example_configs/bootstrap/bootstrap.md
@@ -250,7 +250,7 @@ spec:
 
           env:
             - name: LLDAP_URL
-              value: "http://lldap:8080"
+              value: "http://lldap:17170"
 
             - name: LLDAP_ADMIN_USERNAME
               valueFrom: { secretKeyRef: { name: lldap-admin-user, key: username } }
@@ -260,6 +260,9 @@ spec:
 
             - name: DO_CLEANUP
               value: "true"
+
+            - name: LLDAP_FORCE_SET_PASSWORD
+              value: "false"
 
           volumeMounts:
             - name: bootstrap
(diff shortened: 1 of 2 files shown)


example_configs/bootstrap/bootstrap.md
```diff
@@ -312,3 +305,14 @@ spec:
                     - key: group-configs.json
                       path: group-configs.json
 ```
+### Kubernetes config map for bootstrap.sh script
+
+```yaml
+apiVersion: v1
+data:
+  bootstrap.sh: |
+    PASTE CONTENT OF bootstrap.sh HERE
```
nitnelave: Really? There's no way to refer to an existing script?

example_configs/bootstrap/bootstrap.md
```diff
@@ -261,6 +261,9 @@ spec:
             - name: DO_CLEANUP
               value: "true"
 
+            - name: LLDAP_FORCE_SET_PASSWORD
```
nitnelave: so, IIUC, this PR makes the default behavior be "Only set the password if the user doesn't exist" which makes the password field a "default password", and this flag enables forcing the setting.

What I don't understand is: why have a default password? If you're creating it for users, then you _shouldn't_ have a default password, they should reset it themselves before logging in for the first time.

example_configs/bootstrap/bootstrap.md
jore1699: I checked and the current instructions for using bootstrap.sh in kubernetes use an old version of LLDAP (v0.5.0). In that version the bootstrap.sh file was not included in the docker image. That's why it was mounted manually from a config map...
Will do some tests with a newer version and then update the .md file.

example_configs/bootstrap/bootstrap.md
jore1699: That's a good point, thanks. Will revert the script. The solution to my problem (the password being overwritten every time the bootstrap runs) is to not set a password in the first place.
But where/how should users reset their password? Through a "connected" app with support for pw reset via ldap?

Probably shouldn't tinker and make PRs half asleep...

nitnelave (approved):

PR #1247, opened 2025-08-17
app: Add site_header_label for very basic branding customization

joseph-long:
All this does is add a config option that changes the `<title>` and the header component to read something other than "LLDAP Administration" and "LLDAP" respectively.

I have tested it locally, but I'm not sure what a unit test should look like for this.

### DIFF
--- a/app/src/components/app.rs
+++ b/app/src/components/app.rs
@@ -44,6 +44,7 @@ pub fn app_container() -> Html {
 
 pub struct App {
     user_info: Option<(String, bool)>,
+    site_header_label: String,
     redirect_to: Option<AppRoute>,
     password_reset_enabled: Option<bool>,
 }
@@ -73,6 +74,7 @@ impl Component for App {
                             None
                         })
                 }),
+            site_header_label: "LLDAP".into(),
             redirect_to: Self::get_redirect_route(ctx),
             password_reset_enabled: None,
         };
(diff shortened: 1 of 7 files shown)


nitnelave:
Thank you for the PR, and it looks nice, but my question is: is it really necessary? I'd rather keep the software simple if possible, and that includes trimming non-essential "for the looks" functionality.

joseph-long:
That's a good point, but I think there's decent argument for this little change. If you have users (even for self-hosted stuff, could be friends or family) you'd like to help them practice good security hygiene by not typing credentials into random sites. This makes it easier to tell your LLDAP apart from someone else's.

Also, I predict that some day there will be people posting GitHub issues about getting their password reset on whatever random LLDAP deployment (if there haven't been already!)... A way to white-label LLDAP (which this PR isn't a complete solution for) could be a good defense.

nitnelave:
Hmmm... I don't buy the security aspect. The URL is the only authoritative part.

And I haven't felt the need to distance myself from specific LLDAP instances either :)

joseph-long:
If you would prefer it, I could try making a "no_branding" setting that just makes the name "Account Management" or something and removes the discord etc. links in the footer.

nitnelave:
Again, I don't see the point, unless you consider that "LLDAP" has such a bad branding that you want to dissociate yourself from the project!

joseph-long:
Not at all 😄 it's more I thought you might not want your branding on *my* project. (Nothing nefarious, just user auth for an astronomical instrument. But, in my experience, astronomers get confused when faced with novel acronyms—and I anticipate getting questions about what an LLDAP is.)

Feel free to close!

nitnelave:
Then let them click the link and learn what an "LLDAP" is, if they dare :p

PR #1248, opened 2025-08-20
server: DB Connection Pool configuration

thielj:
Alrighty, I tried my Rust hat with some help from AI. Go easy on me.

Automatic summary follows at the end.
 
Reference: [Discussions](https://github.com/lldap/lldap/discussions/910)

What still needs to be done is migrating database_url (or LLDAP_DATABASE_URL) from existing setups to the new db_options.url (LLDAP_DB_OPTIONS__URL). I tried to implement this but couldn't get my head around the `mut` stuff to promote the deprecated field to the new struct if the `url` field there is `None`. There's a `// TODO` comment somewhere this may fit. Your help here is very much appreciated.

The new configuration also affects the examples and in particular the default configuration file in docker that sets LLDAP_DATABASE_URL explicitly. I haven't touched these yet to keep the scope small.

I decided to increase max_connections default for SQLite. Async/await should benefit from this. If not, it shouldn't matter as the pool only creates one connection initially and everything else on demand. For non-SQLite databases, I defaulted this to min(5), max(10).

I've also defaulted the connection timeout to 3 seconds. This should help improve responsiveness if a remote DB isn't reachable.

Finally, I made the DB log level dependent on the verbose flag using either Debug or Info.

---  

This pull request refactors how database connection options are configured and used in the server. Instead of relying on a single `database_url` field, it introduces a new `DbOptions` struct to group all database-related settings, making configuration more flexible and future-proof. The changes affect CLI argument parsing, configuration management, and database connection setup.

**Configuration improvements:**

* Added a new `DbOpts` struct in `server/src/cli.rs` to handle database connection options via CLI, including URL, connection pool sizes, timeouts, and lifetimes. (`[server/src/cli.rsR215-R242](diffhunk://#diff-7626d972d62203e5dc68698a3ac2adcef8ea1642b095870c93945da03324cfb6R215-R242)`)
* Updated the main configuration struct (`Configuration` in `server/src/configuration.rs`) to use a new `DbOptions` field instead of a single `database_url`, with `database_url` now deprecated. (`[server/src/configuration.rsL115-R142](diffhunk://#diff-40bbcf477684ac3a8b76a2303a5f4e2e63214f4044f97cd1bef0eb4aad086606L115-R142)`)
* Implemented `ConfigOverrider` for `DbOpts`, allowing CLI/database environment overrides to be correctly applied to the configuration. (`[[1]](diffhunk://#diff-40bbcf477684ac3a8b76a2303a5f4e2e63214f4044f97cd1bef0eb4aad086606R507-R529)`, `[[2]](diffhunk://#diff-40bbcf477684ac3a8b76a2303a5f4e2e63214f4044f97cd1bef0eb4aad086606R479)`)
* Added a warning message when the deprecated `database_url` field is used, guiding users to switch to `db_options`. (`[server/src/configuration.rsR686-R690](diffhunk://#diff-40bbcf477684ac3a8b76a2303a5f4e2e63214f4044f97cd1bef0eb4aad086606R686-R690)`)

**Database connection logic updates:**

* Refactored database connection setup in `server/src/main.rs` to use all options from `DbOptions`, including pool limits and timeouts, and made logging level configurable based on verbosity. (`[server/src/main.rsL104-R145](diffhunk://#diff-01fd5002f7585d6f2b853a8c83e6c0a1acb2b16b6e6f8f581fb3365377d331bcL104-R145)`)
* Updated function signatures and usages throughout the codebase to use `DbOptions` instead of just `DatabaseUrl`, including in `setup_sql_tables` and related server setup functions. (`[[1]](diffhunk://#diff-01fd5002f7585d6f2b853a8c83e6c0a1acb2b16b6e6f8f581fb3365377d331bcL104-R145)`, `[[2]](diffhunk://#diff-01fd5002f7585d6f2b853a8c83e6c0a1acb2b16b6e6f8f581fb3365377d331bcL278-R292)`)

These changes make database configuration more robust and extensible, and prepare for future enhancements or backend changes.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -125,7 +125,7 @@ async fn setup_sql_tables(database_url: &DatabaseUrl) -> Result<DatabaseConnecti
 }
 
 #[instrument(skip_all)]
-async fn set_up_server(config: Configuration) -> Result<ServerBuilder> {
+async fn set_up_server(config: Configuration) -> Result<(ServerBuilder, DatabaseConnection)> {
     info!("Starting LLDAP version {}", env!("CARGO_PKG_VERSION"));
 
     let sql_pool = setup_sql_tables(&config.database_url).await?;
@@ -214,9 +214,9 @@ async fn set_up_server(config: Configuration) -> Result<ServerBuilder> {
         .await
         .context("while binding the TCP server")?;
     // Run every hour.
-    let scheduler = Scheduler::new("0 0 * * * * *", sql_pool);
+    let scheduler = Scheduler::new("0 0 * * * * *", sql_pool.clone());
     scheduler.start();
-    Ok(server_builder)
+    Ok((server_builder, sql_pool))
 }
 
 async fn run_server_command(opts: RunOpts) -> Result<()> {
@@ -225,9 +225,14 @@ async fn run_server_command(opts: RunOpts) -> Result<()> {
     let config = configuration::init(opts)?;
     logging::init(&config)?;
 
-    let server = set_up_server(config).await?.workers(1);
+    let (server, sql_pool) = set_up_server(config).await?;
+    let server = server.workers(1);
 
-    server.run().await.context("while starting the server")
+    let result = server.run().await.context("while starting the server");
+    if let Err(e) = sql_pool.close().await {
+        error!("Error closing database connection pool: {}", e);
+    }
+    result
 }
 
 async fn send_test_email_command(opts: TestEmailOpts) -> Result<()> {
@@ -275,8 +280,11 @@ async fn create_schema_command(opts: RunOpts) -> Result<()> {
     debug!("CLI: {:#?}", &opts);
     let config = configuration::init(opts)?;
     logging::init(&config)?;
-    setup_sql_tables(&config.database_url).await?;
+    let sql_pool = setup_sql_tables(&config.database_url).await?;
     info!("Schema created successfully.");
+    if let Err(e) = sql_pool.close().await {
+        error!("Error closing database connection pool: {}", e);
+    }
     Ok(())
 }
 


nitnelave:
I was hoping for a much smaller PR, just changing the hardcoded numbers in the main. I'm not convinced we need that much customization.

thielj:
I would think that most people don't need  to configure a DB at all - but once they replace SQLite and dig deeper, they might appreciate a familiar set of options in one place.

nitnelave:
Let's aim for minimalist and add things if we really need to. I'm thinking it won't ever come up again (hopefully!)

nitnelave (changes requested):
Let's keep the change minimal and hardcode the new values.

thielj:
Would you have a comprehensive data analysis of which hardcoded values are
best for the usage scenarios you envision?

nitnelave:
I thought that the ones we talked about would be a good start. We can see if we need to make it configurable later. I think it was 8 connections, no?

thielj:
You seem to know which values are best for everyone.

I simply don't.

If the DB server idles in the next rack, requirements are obviously very
different compared to busy servers 2000km away. Same for timeouts. For a
request coming in while the worker thread is hashing none of this matters
anyway.

🤷

nitnelave:
Hey, no need to take that tone! We're just working based on different priorities.

One of the main goals of LLDAP is to have an easy-to-install, easy-to-use LDAP server to replace openLDAP for self-hosters. Note that this doesn't mention performance or configurability as a goal (also note that it says "opinionated"). I try to accommodate these when I can, but still I try to reduce the amount of configuration points when they're not absolutely necessary. Most users of LLDAP have a co-located server (SQLite or PG), and a number of queries per day likely measured in double digits. Neither throughput nor latency are primary goals for them.

From our conversations, it seems like holding a pool of 8 connections should be more than enough to be able to respond with low latency while not using too many resources. It won't be tailored for a high performance environment, but that's okay.

thielj:
If you need another reason to make this configurable: I hit the DB user's connection limit today. 

What basically happened was that all LLDAP instances with this patch saw their pool connection count creep up to the allowed maximum over the last 10 days or so. No idea yet why that happened. Timeout and maximum lifetime were using the default values (300s, and 3600s).

thielj:
Digging a bit deeper into this: I have over 40 "dead" connections from LLDAP that were opened between 8h and 5 days ago. PostgreSql is typically configured without a connection timeout itself so doesn't clean this up automatically. Other DB servers will probably have quite long timeouts as well.

Do you have a graceful shutdown handler somewhere to call pool.close().await() ?

--

https://docs.rs/sqlx/latest/sqlx/struct.Pool.html

### Note: Drop Behavior

Due to a lack of async Drop, dropping the last Pool handle may not immediately clean up connections by itself. The connections will be dropped locally, which is sufficient for SQLite, but for client/server databases like MySQL and Postgres, that only closes the client side of the connection. The server will not know the connection is closed until potentially much later: this is usually dictated by the TCP keepalive timeout in the server settings.

Because the connection may not be cleaned up immediately on the server side, you may run into errors regarding connection limits if you are creating and dropping many pools in short order.

We recommend calling [.close().await](https://docs.rs/sqlx/latest/sqlx/struct.Pool.html#method.close) to gracefully close the pool and its connections when you are done using it. This will also wake any tasks that are waiting on an .acquire() call, so for long-lived applications it’s a good idea to call .close() during shutdown.

nitnelave:
That would be in `run_server_command` in main.rs

That sounds like a good idea to do regardless of whether we want to make the number of connections configurable. Would you mind making this another PR? It should be easy to review.

thielj:
My knowledge of Rust isn't good, but adding a statement there might not be
enough - unless "something" already handles a clean shutdown. The more
common scenario is to just exit() and let the OS do the cleanup. Just try
it with a print at the end of run_server_command and send the process a
"kill PID". If the print is executed: perfect.

If not, maybe have a look here:
https://chatgpt.com/share/68b8b78e-6a60-8013-a044-b4a8578eafec

And yes, this is a good idea, regardless of how many connections you have,
or if they're pooled or not.

nitnelave:
Oh, I know the signal is caught and orderly shutdown is interrupted.

Issue #1249, opened 2025-08-20
[BUG] Schema breaks query with `memberof` attribute


deuill:
**Describe the bug**

Seen with `ldap3` library in a script written in Python, but it seems that performing a search with a `memberof` attribute will fail on 0.6.2, but would succeed in 0.6.1.

**To Reproduce**

Sample script:

```python
from ldap3 import Server, Connection, AUTO_BIND_NO_TLS
from ldap3.utils.conv import escape_filter_chars

ldap_bind_dn = "admin,ou=people,dc=ldap,dc=local"
ldap_password = "admin"

ldap_base_dn = "ou=people,dc=ldap,dc=local"
ldap_search_filter = "(&(memberof=cn=test_user,ou=groups,dc=ldap,dc=local)(|(uid={username})(mail={username})))"


def main():
    server = Server("localhost", port=389)
    connection = Connection(
        server, user=ldap_bind_dn, password=ldap_password, auto_bind=AUTO_BIND_NO_TLS
    )
    username="test@test.com"
    connection.search(
        search_base=ldap_base_dn,
        search_filter=ldap_search_filter.format(username=escape_filter_chars(username)),
        search_scope=SUBTREE,
        attributes=["uid"],
        time_limit=15,
    )
```

This will fail with `ldap3.core.exceptions.LDAPAttributeError: invalid attribute memberOf` on the last call to `connection.search()`.

**Expected behavior**

Using `memberof` attribute should succeed in this case.

**Logs**

[lldap-python-ldap3-fail.log](https://github.com/user-attachments/files/21901255/lldap-python-ldap3-fail.log)

nitnelave:
The problem is that in the schema  returned by LLDAP, there is no mention of the memberOf attribute. We need to add that to the response.

nitnelave:
Should be fixed now!

PR #1254, opened 2025-08-20
Add memberOf attribute definition to LDAP schema

nitnelave:
From https://github.com/lldap/lldap/pull/1250 :

The memberOf attribute was missing from the LDAP schema definition returned by LLDAP, causing some LDAP clients (like Python's ldap3 library) to reject searches using the memberOf attribute with an LDAPAttributeError.

While the memberOf functionality worked correctly for searches and filters, the attribute was not declared in the schema's attributeTypes section, leading to client-side validation failures.

This change adds the memberOf attribute definition to the hardcoded attributes list in the LDAP schema response with:

-     OID: 1.2.840.113556.1.2.102 (Microsoft's standard OID for memberOf)
-     Distinguished Name syntax for group DN values
-     Multi-valued attribute (no SINGLE-VALUE restriction)
-     NO-USER-MODIFICATION flag (computed attribute)

The fix ensures LDAP clients can properly validate and use the memberOf attribute in search filters and attribute requests without schema validation errors.

Example that now works:

```python
from ldap3 import Server, Connection, SUBTREE, AUTO_BIND_NO_TLS
from ldap3.utils.conv import escape_filter_chars

server = Server("localhost", port=3890)
connection = Connection(server, user="uid=admin,ou=people,dc=example,dc=com", password="password", auto_bind=AUTO_BIND_NO_TLS)
connection.search(
    search_base="ou=people,dc=example,dc=com",
    search_filter="(&(memberof=cn=test_user,ou=groups,dc=example,dc=com)(|(uid=test@test.com)(mail=test@test.com)))",
    search_scope=SUBTREE,
    attributes=["uid"]
)
# No longer throws: ldap3.core.exceptions.LDAPAttributeError: invalid attribute memberOf
```

Fixes https://github.com/lldap/lldap/issues/1249.

### DIFF
--- a/crates/ldap/src/search.rs
+++ b/crates/ldap/src/search.rs
@@ -236,6 +236,7 @@ pub fn make_ldap_subschema_entry(schema: PublicSchema) -> LdapOp {
             vals: {
                 let hardcoded_attributes = [
                     b"( 0.9.2342.19200300.100.1.1 NAME ( 'uid' 'userid' 'user_id' ) DESC 'RFC4519: user identifier' EQUALITY caseIgnoreMatch SUBSTR caseIgnoreSubstringsMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{256} SINGLE-VALUE NO-USER-MODIFICATION )".to_vec(),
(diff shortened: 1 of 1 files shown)


T0byV:
Manually corrected the attribute to fully match the OpenLDAP attribute definition and changed the index from where the non-hardcoded attributes numbering starts, as it was counting all attributes even if they were not in the `10.x` range. @nitnelave could you have a very quick look at the PR again? (As I technically worked on the PR)

nitnelave:
Thanks! That looks good :)

PR #1271, opened 2025-09-04
server: DB Connection Pool configuration (merged)

thielj:
See https://github.com/lldap/lldap/pull/1248

### DIFF
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -88,19 +88,6 @@
 ## Alternatively, you can set it to "always" to reset every time the server starts.
 # force_ldap_user_pass_reset = false
 
-## Database URL.
-## This encodes the type of database (SQlite, MySQL, or PostgreSQL)
-## , the path, the user, password, and sometimes the mode (when
-## relevant).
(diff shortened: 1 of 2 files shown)


nitnelave:
Let's try to keep PRs small and separate, one PR per issue.

I'm not convinced that we need a whole new service for a background hasher thread. Bumping up the number of workers seems like a much easier fix for 99% of cases.

As for the shutdown logs, what is merged in master works for me: whether I send SIGTERM or SIGINT, there is a graceful shutdown and the pool is cleaned up.

thielj:
I created the PR this morning without the password_service code - which was only committed this afternoon. 

Not sure why this is referenced here.

nitnelave:
Just checking: are you not sure why the password service is included in the PR, or why I'm talking about it? Because the proposed changes for this PR currently include the password service

thielj:
As I said, the PR was created hours before the password_service code was even written. Seeing the other entries here, I suppose everything pushed to main in my own fork will be included here :(

@Copilot please explain how I can cherry-pick a PR.

thielj:
see new set of PRs on dedicated branches

PR #1279, opened 2025-09-05
feat: readonly backend

thielj:
PR 3/3 - this builds on top of PR2 and I had to include PR1 to not cause a merge conflict (two adjacent lines from each PR). PR1 is *NOT* required though.

### Changes:
- added `readonly` configuration option to `DbOptions`
- added parsing etc
- added `fn is_readonly()` to `Backend` trait
- implemented trait where necessary
- added `readonly` flag to `SqlBackendHandler`
- modified write operations to check readonly status
- the frontend only exposes /health if the backend is readonly

### Motivation:
Deploying 'headless' local LLDAP instances using a read-only replica - or for some use cases, maybe even sync'ed SQLite files. Users are typically directed to a single master instance for all account operations. 

### Configuration:
Add `readonly = true` to the `[db_options]` section to enable readonly mode.

### Testing:
- All existing functionality preserved
- Readonly mode blocks write operations
- /health is still available for docker health checks
- Readonly mode doesn't try to schemes, users, groups, etc
- Readonly mode doesn't run the cron cleanup job

### TBD:
- Re-enabling the GraphQL endpoint is an option. If tokens are obtained from the master instance, they should be available in the read-only replica. I don't do GraphQL though.
- Re-enabling the GUI in strict read-only mode is an option, but I don't think there's a use case for this.  
- Low level checks might be removed if write functionality is disabled i higher levels

### DIFF
--- a/crates/domain-handlers/src/handler.rs
+++ b/crates/domain-handlers/src/handler.rs
@@ -184,6 +184,7 @@ pub trait BackendHandler:
     + ReadSchemaBackendHandler
     + SchemaBackendHandler
 {
+    fn is_readonly(&self) -> bool;
 }
 
 #[cfg(test)]
--- a/crates/sql-backend-handler/src/sql_backend_handler.rs
+++ b/crates/sql-backend-handler/src/sql_backend_handler.rs
@@ -24,14 +24,14 @@ impl SqlBackendHandler {
     pub fn pool(&self) -> &DbConnection {
         &self.sql_pool
     }
-
-    pub fn is_readonly(&self) -> bool {
-        self.readonly
-    }
 }
 
 #[async_trait]
-impl BackendHandler for SqlBackendHandler {}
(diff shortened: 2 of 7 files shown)


thielj:
That would bring LLDAP to what I have started to use on the test server. No further updates planned at the moment.

nitnelave:
I don't want to offend you, @thielj , but I feel like you're not listening to me :)

We have had multiple discussions in various issues going back and forth over whether the features in the 3 PRs you listed should be implemented or not, and in all 3 cases I said that they were out of scope. I appreciate your enthusiasm and the work you want to do on the project, but you're pulling it in a direction I do not want to go to. I'm not going to rehash the arguments, but I hope you won't be surprised if I close the PRs.

thielj:
And I hope you realize that I didn't do these for you, but for those who might have a need for these patches.

PR #1282, opened 2025-09-08
Password policy

gplubeck:
BLUF: This PR add the ability to set password complexity requirements via a PasswordPolicyOptions struct.

There appear to be a number of requests for volunteer organizations or small businesses (https://github.com/lldap/lldap/issues/521, https://github.com/lldap/lldap/issues/783, etc) that want basic password complexity. This pull request would add the ability to to add pseudo complex password.

The struct, its default implementation, and the validation function for password have been placed in the frontend-options create.  Perhaps there is a better location for the struct definition and the validation function?

Additionally, the server configuration now loads the password policy options when started.  This means password policies can change as requirements change.

set-password now queries the /settings endpoint and receives back the password policy options and enforces them unless the --bypass-password-policy is given.  Moreover, the front end also queries the password policy and bubbles up all password requirement failures as seen in the screenshot.

password_rejection

## Summary by CodeRabbit

- New Features
  - Clients (web UI and CLI) fetch and apply server password policy and validate passwords before submission.

- Bug Fixes
  - Validation now aggregates violations and displays multiline, user-friendly error messages.
  - Policy-driven checks replace prior inline/minimum-only checks.

- Documentation
  - Docker template updated with a password_policy section and sensible defaults.

- Chores
  - Password policy propagated end-to-end from server configuration to clients and CLI.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/frontend-options/src/lib.rs
+++ b/crates/frontend-options/src/lib.rs
@@ -1,5 +1,13 @@
 use anyhow::{Result, bail};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
+
+fn deserialize_specials<'de, D>(deserializer: D) -> Result<Vec<char>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    Ok(s.chars().collect())
+}
 
 #[derive(Serialize, Deserialize)]
 pub struct Options {
@@ -15,7 +23,9 @@ pub struct PasswordPolicyOptions {
     pub min_lowercase: usize,
     pub min_digits: usize,
     pub min_special: usize,
-    pub allowed_specials: Vec<char>,
+
+    #[serde(deserialize_with = "deserialize_specials")]
+    pub special_characters: Vec<char>,
 }
 
 impl Default for PasswordPolicyOptions {
@@ -26,7 +36,7 @@ impl Default for PasswordPolicyOptions {
             min_lowercase: 0,
             min_digits: 0,
             min_special: 0,
-            allowed_specials: vec!['!', '@', '#', '$', '%', '^', '&', '*'],
+            special_characters: vec!['!', '@', '#', '$', '%', '^', '&', '*'],
         }
     }
 }
@@ -54,7 +64,7 @@ pub fn validate_password(password: &str, policy: &PasswordPolicyOptions) -> Resu
             lowercase += 1;
         } else if c.is_ascii_digit() {
             digits += 1;
-        } else if policy.allowed_specials.contains(&c) {
+        } else if policy.special_characters.contains(&c) {
             special += 1;
         }
     }
@@ -85,7 +95,7 @@ pub fn validate_password(password: &str, policy: &PasswordPolicyOptions) -> Resu
             "Password must contain at least {} special character(s) from: {}",
             policy.min_special,
             policy
-                .allowed_specials
+                .special_characters
                 .iter()
                 .map(|c| c.to_string())
                 .collect::<Vec<_>>()
--- a/lldap_config.docker_template.toml
+++ b/lldap_config.docker_template.toml
@@ -91,7 +91,7 @@
 
 ## Password Policy Options
 [password_policy]
-# Minimum length of password.  Recommended 8
+# Minimum length of password.  Recommended 12, default is 8.
 min_length = 8
 # Minimum number of uppercase characters required.
 # Default is 0.  This feature is not recommended as it is an anti pattern in security.
@@ -105,9 +105,9 @@ min_length = 8
 # Minimum number of special characters required
 # Default is 0.  This feature is not recommended as it is an anti pattern in security.
 min_special = 1
-# If min_sepcial is greater than zero, this vector of chars is what characters are considered special.
+# If min_special is greater than zero, this vector of chars is what characters are considered special.
 # Default special chars are ! @ # $ % ^ & *
-#allowed_specials= ['@', '*']
+#special_characters= "!@*"
 
 ## Database URL.
 ## This encodes the type of database (SQlite, MySQL, or PostgreSQL)


crates/frontend-options/Cargo.toml
```diff
@@ -8,5 +8,8 @@ homepage.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[dependencies]
+anyhow = "*"
+
```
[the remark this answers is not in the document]
gplubeck: I can do this if you would like.  I was keeping in line with how other crates in the repo were including anyhow.

set-password/src/main.rs
```diff
@@ -100,6 +101,18 @@ pub fn register_finish(
...
+    let url = append_to_url(base_url, "settings");
+    let client = reqwest::blocking::Client::new();
+    let resp = client
+        .get(url)
+        .bearer_auth(token)
+        .send()?
+        .error_for_status()?;
+    let options: Options = resp.json()?;
+    Ok(options)
+}
```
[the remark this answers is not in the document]
gplubeck: This seems like a good change.  I can implement.

app/src/components/change_password.rs
```diff
@@ -72,6 +63,8 @@ pub enum Msg {
     SubmitNewPassword,
     RegistrationStartResponse(Result<Box<registration::ServerRegistrationStartResponse>>),
     RegistrationFinishResponse(Result<()>),
+    PasswordPolicyResponse(lldap_frontend_options::PasswordPolicyOptions),
+    ShowError(String),
 }
```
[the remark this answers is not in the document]
gplubeck: This appears to be a hallucination from coderabbitai.  Only two l's in the actual code.  I can change to include the create at the top level; however, current state is consistent with file.

app/src/components/change_password.rs
```diff
@@ -11,6 +11,8 @@ use crate::{
 use anyhow::{Result, anyhow, bail};
 use gloo_console::error;
 use lldap_auth::*;
+use crate::infra::api;
```
nitnelave: Is that how the formatter organizes the imports? Or do you maybe need to run `cargo fmt --workspace`?

app/src/components/change_password.rs
```diff
@@ -72,6 +63,8 @@ pub enum Msg {
     SubmitNewPassword,
     RegistrationStartResponse(Result<Box<registration::ServerRegistrationStartResponse>>),
     RegistrationFinishResponse(Result<()>),
+    PasswordPolicyResponse(lldap_frontend_options::PasswordPolicyOptions),
+    ShowError(String),
```
nitnelave: The more common pattern is 
```suggestion
    PasswordPolicyResponse(Result<lldap_frontend_options::PasswordPolicyOptions>),
```

app/src/components/change_password.rs
```diff
@@ -195,13 +207,34 @@ impl Component for ChangePasswordForm {
     type Message = Msg;
     type Properties = Props;
 
-    fn create(_: &Context<Self>) -> Self {
-        ChangePasswordForm {
+    fn create(ctx: &Context<Self>) -> Self {
+        let this = Self {
```
nitnelave:
```suggestion
        let self = Self {
```

Nit, for consistency with the other instances of the pattern.

app/src/components/change_password.rs
```diff
@@ -211,61 +244,63 @@ impl Component for ChangePasswordForm {
         let is_admin = ctx.props().is_admin;
         let link = ctx.link();
         html! {
-          <>
-            <div class="mb-2 mt-2">
-              <h5 class="fw-bold">
+            <>
+                <div class="mb-2 mt-2">
```
nitnelave, line 235: Is the indentation okay? The changes look weird

app/src/components/change_password.rs
```diff
@@ -195,13 +207,34 @@ impl Component for ChangePasswordForm {
...
 
-    fn create(_: &Context<Self>) -> Self {
-        ChangePasswordForm {
+    fn create(ctx: &Context<Self>) -> Self {
+        let this = Self {
             common: CommonComponentParts::<Self>::create(),
             form: yew_form::Form::<FormModel>::new(FormModel::default()),
             opaque_data: OpaqueData::None,
+            password_policy: None,
+            error_message: None,
```
nitnelave: I think you don't need this field, `self.common.error` already has it.

app/src/components/change_password.rs
```diff
@@ -211,61 +244,63 @@ impl Component for ChangePasswordForm {
...
-                  required=true
-                  label="Current password"
-                  field_name="old_password"
-                  input_type="password"
-                  autocomplete="current-password"
-                  oninput={link.callback(|_| Msg::FormUpdate)} />
-              }} else { html! {} }}
-              <Field<FormModel>
+            {if !is_admin { html! {
+                                      <Field<FormModel>
```
nitnelave, line 253: Indentation looks weird.

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,102 @@
...
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(default)]
+pub struct PasswordPolicyOptions {
+    pub min_length: usize,
+    pub min_uppercase: usize,
+    pub min_lowercase: usize,
+    pub min_digits: usize,
+    pub min_special: usize,
+    pub allowed_specials: Vec<char>,
```
nitnelave: The name is misleading: it implies that any character that's not a digit, letter or in this list would be rejected (yes, I've seen that happen many times).

This is rather the list of characters that count as special. Maybe just `special_characters`

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,102 @@
...
+                .join(" ")
+        ));
+    }
+
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        // join all messages into one big error string, or handle Vec<String> upstream
+        bail!("{}", errors.join("\n"));
+    }
```
nitnelave:
```suggestion
    if !errors.is_empty() {
        // join all messages into one big error string, or handle Vec<String> upstream
        bail!("{}", errors.join("\n"));
    }
    Ok(())
```

lldap_config.docker_template.toml
```diff
@@ -88,6 +88,27 @@
 ## Alternatively, you can set it to "always" to reset every time the server starts.
 # force_ldap_user_pass_reset = false
 
+
+## Password Policy Options
+[password_policy]
+# Minimum length of password.  Recommended 8
```
nitnelave:
```suggestion
# Minimum length of password.  Recommended 12, default is 8.
```

lldap_config.docker_template.toml
```diff
@@ -88,6 +88,27 @@
...
+# Minimum number of lowercase characters required.
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+#min_lowercase = 2
+# Minimum number of digits required.
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+#min_digits = 2
+# Minimum number of special characters required
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+min_special = 1
+# If min_sepcial is greater than zero, this vector of chars is what characters are considered special.
```
nitnelave:
```suggestion
# If min_special is greater than zero, this vector of chars is what characters are considered special.
```

lldap_config.docker_template.toml
```diff
@@ -88,6 +88,27 @@
...
+# Default special chars are ! @ # $ % ^ & *
+#allowed_specials= ['@', '*']
```
nitnelave: Can we make it parse from a string? It's much easier to type `"%[(}+="` than `['%', '[', ...]`

set-password/src/main.rs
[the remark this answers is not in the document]
nitnelave: Yeah, that'd be nice.

nitnelave (changes requested):

app/src/components/change_password.rs
```diff
@@ -211,61 +244,63 @@ impl Component for ChangePasswordForm {
...
+                    if let Some(e) = &self.common.error {
+                        let err_str = e.to_string();
+                        let lines = err_str.lines();
+                        html! {
+                            <div class="alert alert-danger mt-3 mb-3">
+                            { for lines.map(|line| html! { <p>{ line }</p> }) }
+                            </div>
+                        }
+                    } else { html! {} }
                 }
```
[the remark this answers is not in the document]
gplubeck: Left over from old iterations.  Removed error_message from ChangePasswordForm struct in favor of self.common.error.

app/src/components/change_password.rs
gplubeck: Updated.

crates/frontend-options/Cargo.toml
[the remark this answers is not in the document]
gplubeck: Pinned to 1.

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,102 @@
...
+pub fn validate_password(password: &str, policy: &PasswordPolicyOptions) -> Result<()> {
+    let mut errors = Vec::new();
+
+    if password.len() < policy.min_length {
+        errors.push(format!(
+            "Password must be at least {} characters long.",
+            policy.min_length
+        ));
+    }
+
```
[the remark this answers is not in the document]
gplubeck: Updated for non single byte characters.

app/src/components/change_password.rs
gplubeck: You are correct.  Left over from old iterations.

gplubeck:
I think that is all the changes @nitnelave requested.  One thing I found during testing of this is that since I implemented these changes through the change password component, the rules do not apply to the initial account creation.  I am okay with that, but just wanted to bring it up in case someone does not like that workflow.

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,111 @@
...
+}
 
 #[derive(Serialize, Deserialize)]
 pub struct Options {
     pub password_reset_enabled: bool,
+    pub password_policy: PasswordPolicyOptions,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(default)]
```
nitnelave, line 19: Why the serde default?  Same in configuration.rs

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,111 @@
...
+pub struct PasswordPolicyOptions {
+    pub min_length: usize,
+    pub min_uppercase: usize,
+    pub min_lowercase: usize,
+    pub min_digits: usize,
+    pub min_special: usize,
+
+    #[serde(deserialize_with = "deserialize_specials")]
+    pub special_characters: Vec<char>,
```
nitnelave, line 28: I think it would be simpler but equivalent if we stored a plain string here.

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,111 @@
...
+                .special_characters
+                .iter()
+                .map(|c| c.to_string())
+                .collect::<Vec<_>>()
+                .join(" ")
+        ));
+    }
+
+    if !errors.is_empty() {
+        // join all messages into one big error string, or handle Vec<String> upstream
```
nitnelave, line 107: It seems you made the choice to join the errors :)

I think it's fine, we can change it later if needed.

We can remove the comment

app/src/components/change_password.rs
```diff
@@ -195,11 +200,26 @@ impl Component for ChangePasswordForm {
     type Message = Msg;
     type Properties = Props;
 
-    fn create(_: &Context<Self>) -> Self {
-        ChangePasswordForm {
+    fn create(ctx: &Context<Self>) -> Self {
+        let link = ctx.link().clone();
+        wasm_bindgen_futures::spawn_local(async move {
```
nitnelave, line 205: I think we should fetch the options once on app startup, and store them in App. Then we can pass them to this component as a Property

crates/frontend-options/src/lib.rs
```diff
@@ -1,6 +1,111 @@
...
+
+    let pass_len = password.chars().count();
+    if pass_len < policy.min_length {
+        errors.push(format!(
+            "Password must be at least {} characters long.",
+            policy.min_length
+        ));
+    }
+
+    let mut uppercase = 0;
```
nitnelave, line 55: I'd rather have each check be separate and encapsulated, even if we have to run 4 times through the password (oh, the performance!)

Something like:
```rust
{
  let uppercase = password.chars().filter(is_uppercase).count();
  if uppercase < policy { ... }
}
```
And give the user the number of uppercase characters in the password, don't make them count.

There are avenues for refactoring too, the checks are similar.

lldap_config.docker_template.toml
```diff
@@ -88,6 +88,27 @@
...
+#min_uppercase = 1
+# Minimum number of lowercase characters required.
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+#min_lowercase = 2
+# Minimum number of digits required.
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+#min_digits = 2
+# Minimum number of special characters required
+# Default is 0.  This feature is not recommended as it is an anti pattern in security.
+min_special = 1
```
nitnelave, line 107: This should be commented out

Issue #352, opened 2022-10-25
Support header sso (Forward Auth)


RandomExplosion:
Since one common use case for lldap is as a ldap server for a separate authentication server such as authelia it would be good if the web ui was compatible with login from those providers e.g. [through trusted header sso](https://www.authelia.com/integration/trusted-header-sso/introduction/)

Currently if one wants to put lldap behind 2fa you can still use forward auth with a reverse proxy to block access to the dashboard if you haven't signed in with 2fa but you still need to sign-in again with your password after that.

Basically you can either have username+password or 2x username and password + 2fa
This works fine but does feel a bit clunky.

As I understand it it shouldn't be *too* difficult to implement either since it takes the place of the current login screen, either way I haven't read through all of lldap's source to know for sure.

dvv:
@nitnelave 
I wonder if we could switch native auth off completely with some config option.
Then authenticating to lldap via authelia (using lldap per se) would be similar to securing vanilla whoami example.

nitnelave:
@dvv How would you get the current user? The proposed way (that I approve) is through trusted headers, but do you have another proposal?

dvv:
Exactly via trusted headers. I think if we remove native login and honor `Remote-User:` as the current legal user.

nitnelave:
The backend changes are going to be straightforward (just parse the header if it's there), it's the front-end changes that are going to be a bit more challenging, I think: When the frontend loads, it doesn't know which user is logged in. It would have to do a request to the server to check "am I already logged in?" since it can't rely on a cookie.

dvv:
Well, I believe non-http-only cookies should go away.
Probably there should be some "profile" API (RPC?) method that returns user info.
I didn't dive in client-side of lldap.

nitnelave:
The only non-http-only cookies right now are storing who's logged in and whether they're admin, but they don't store any credentials (it's to help the frontend). There's a small chance they get out of sync with the actual logged in state (if the session was somehow invalidated from the backend), but so far it's worked out pretty well.

Another way to do it would be indeed to have a "whoami" endpoint that returns the currently logged in user, or none.

fiinnnn:
Is anyone working on this currently? I'm happy to take a shot at implementing it if I have some time.

nitnelave:
No one's working on it at the moment, it's all yours!

tecosaur:
@fiinnnn did you take this for a spin? It would be quite cool to have this functionality :slightly_smiling_face:

[...]

PR #1285, opened 2025-09-13
auth: Add trusted header authentication with IP allowlist

Kumpelinus:
Add support for **trusted header authentication** that allows users to be authenticated via HTTP headers (e.g., from reverse proxies like Authelia, Authentik, or Traefik ForwardAuth).

This implementation provides a **tokenless authentication flow** that works alongside the existing JWT-based authentication:

* **Mutually exclusive authentication**: When a trusted header is present, it takes precedence over JWT authentication in GraphQL requests
* **IP allowlist security**: Only requests from configured trusted proxy IP addresses/networks are allowed to use header authentication
* **Configurable options**: Header name, logout URL, and trusted proxy networks are all configurable

**Configuration options:**
- `trusted_header_options.enabled` - Enable/disable the feature
- `trusted_header_options.header_name` - Header containing username (default: "Remote-User")  
- `trusted_header_options.logout_url` - Optional logout redirect URL
- `trusted_header_options.trusted_proxies` - IP addresses/CIDR networks allowed to send trusted headers (default: localhost only)

The existing username/password login remains available as fallback when trusted headers are not present.


Kumpelinus:
Fixes #352

nitnelave:
Just mentioning that I haven't forgotten about you, but I want to give this PR a proper review since it's quite security-sensitive, and I'm short on free time.

Can you describe the (intended) user flow when logging in through trusted header? What will the first http request look like, what will the response look like, in terms of headers/cookies? What about subsequent same-session requests? What about next-day requests (after JWT expiration)?

Kumpelinus:
Hi, sorry for the late reply. I am busy as well. So, no worries :)

> Just mentioning that I haven't forgotten about you, but I want to give this PR a proper review since it's quite security-sensitive, and I'm short on free time.
> 
> Can you describe the (intended) user flow when logging in through trusted header? What will the first http request look like, what will the response look like, in terms of headers/cookies? What about subsequent same-session requests? What about next-day requests (after JWT expiration)?

I made it so that the trusted header flow bypasses the jwt flow and thus does not really have a "session" with jwt token or cookies. I personally preferred not having to deal with the jwt cookies but I could change that.

bendlas:
I've been wanting this functionality in lldap for some time. I did rebase the PR on master and test drive it in a container with my self-hosting setup. I could get it to work, however, the current handling of the IP allow list is a security footgun:

1. Trusted header auth implies having to trust http headers, thus controlling them in a trusted proxy.
    While the need to control `Remote-User` is evident (and configurable) in the config option, the need to also control `X-Forwarded-For` and `X-Real-Ip` is undocumented und non-configurable.
2. The IP whitelist is meant to apply to the authentication proxy address (the one setting `Remote-User`), however, neither `X-Forwarded-For` nor `X-Real-Ip` are commonly used that way. Rather, they are meant for the original client IP, which is not supposed to be whitelisted.
3. An attacker who's able to spoof `Remote-User` will likely also be able to spoof any other HTTP header, so in this case, the benefit of a whitelist is actually negated.

In my setup, nginx is performing authentication against authelia. In order for this to work, I had to set `X-Forwarded-For` to `127.0.0.1` (in addition to setting `Remote-User`), which - as detailed above - is not documented anywhere and doesn't make sense either.

So in summary, I'll have to agree with the bot: Check only the remote ip from the connection, then the whitelist can protect against real misconfigurations (publicly accessible lldap instances).

bendlas:
> Can you describe the (intended) user flow when logging in through trusted header? What will the first http request look like, what will the response look like, in terms of headers/cookies? What about subsequent same-session requests? What about next-day requests (after JWT expiration)?

IMO the purpose of trusted header auth is to completely pass off authentication to a proxy, sitting in front of lldap. So lldap should be as stateless as possible, as the proxy will already handle user session, expiration, ... so I agree with the author about "no jwt".

The only thing I'd expect lldap to do is to check if the authenticated user matches its current state and otherwise clear out any additional session cookies, ui state, ... that it may keep.

nitnelave:
Again, I apologize for the delay in review. I have no free time these days (LLDAP or not), and this requires a deep review. I agree with @bendlas 's concerns on a conceptual level, but that's without having had a look at the implementation.

NextCloud has a "trusted-proxies" config array (even supporting IP ranges), how does that compare with the proposed solution?
Any config proposal should be footgun resilient: I'd rather do a bunch of expensive checks at startup and be overly restrictive on the config than allow accidentally dangerous configurations. Again, that's high-level, but what can we do to check that the configuration is correct/safe?
Some alternatives include saving the (direct) incoming IPs that we see and enforcing that there should be a single one (the proxy) if trusted headers are configured. But even better is if this IP is set in the config.

server/src/auth_service.rs
```diff
@@ -102,11 +105,40 @@ async fn get_refresh<Backend>(
     data: web::Data<AppState<Backend>>,
     request: HttpRequest,
 ) -> TcpResult<HttpResponse>
+where
```
nitnelave, line 108: Why do we need to update the refresh token route? We don't give out a refresh token, I don't expect it to even work.

server/src/auth_service.rs
```diff
@@ -131,10 +163,134 @@ where
...
+    // Try X-Real-IP header
+    if let Some(real_ip) = request.headers().get("x-real-ip") {
+        if let Ok(header_value) = real_ip.to_str() {
+            if let Ok(ip) = IpAddr::from_str(header_value.trim()) {
+                return Some(ip);
+            }
+        }
+    }
+
+    // Fallback to connection info
```
nitnelave, line 196: I think we should only use the connection info, as the rest can be spoofed.

server/src/auth_service.rs
```diff
@@ -131,10 +163,134 @@ where
...
+    if let Some(peer_addr) = request.peer_addr() {
+        return Some(peer_addr.ip());
+    }
+
+    None
+}
+
+fn is_ip_in_trusted_proxies(ip: IpAddr, trusted_proxies: &[String]) -> bool {
```
nitnelave, line 204: `ip` should be taken by reference.

server/src/configuration.rs
```diff
@@ -83,6 +83,25 @@ impl std::default::Default for LdapsOptions {
...
+#[builder(pattern = "owned")]
+pub struct TrustedHeaderOptions {
+    #[builder(default = "false")]
+    pub enabled: bool,
+    #[builder(default = r#"String::from("Remote-User")"#)]
+    pub header_name: String,
+    #[builder(default = "None")]
+    pub logout_url: Option<String>,
+    #[builder(default = r#"vec!["127.0.0.0/8".to_string(), "::1/128".to_string()]"#)]
+    pub trusted_proxies: Vec<String>,
```
nitnelave, line 96:
```suggestion
    pub trusted_proxies: Vec<IpNet>,
```

Parse early, give early feedback (invalid configuration)

server/src/auth_service.rs
```diff
@@ -131,10 +163,134 @@ where
...
+    }
+
+    let user_id = UserId::new(username);
+
+    // Check if the user exists in LLDAP
+    let user_exists = !data
+        .get_readonly_handler()
+        .list_users(Some(UserRequestFilter::UserId(user_id.clone())), false)
+        .await?
+        .is_empty();
```
nitnelave, line 257: We should directly query the groups as well, to avoid a separate round-trip to the DB.

server/src/auth_service.rs
```diff
@@ -623,6 +801,56 @@ pub(crate) fn check_if_token_is_valid<Backend: BackendHandler>(
...
+#[instrument(skip_all, level = "debug", err, ret)]
+pub(crate) async fn check_if_trusted_header_is_valid<Backend: BackendHandler>(
+    state: &AppState<Backend>,
+    request: &HttpRequest,
+) -> Result<ValidationResults, actix_web::Error> {
+    if !state.trusted_header_options.enabled {
+        return Err(ErrorUnauthorized(
+            "Trusted header authentication is disabled",
+        ));
+    }
```
nitnelave, line 813: We can remove this check, it's done by the caller.

server/src/auth_service.rs
```diff
@@ -623,6 +801,56 @@ pub(crate) fn check_if_token_is_valid<Backend: BackendHandler>(
...
+            client_ip
+        )));
+    }
+
+    let header_name = &state.trusted_header_options.header_name;
+    let username = request
+        .headers()
+        .get(header_name)
+        .and_then(|h| h.to_str().ok())
+        .ok_or_else(|| ErrorUnauthorized(format!("Missing trusted header: {}", header_name)))?;
```
nitnelave, line 831: This whole chunk of logic is duplicated. Is there no way to share the implementation?

nitnelave: All the way down to returning the user+groups.

server/src/configuration.rs
nitnelave: Does this support unmasked IPs without subnet? Or do we need to introduce our own type for that? This would be the place.

nitnelave (changes requested):
Overall, I think the approach is sound. However, we can improve the implementation. As it stands, there is a bit too much duplication, some things that should be done in different places, and some complexity to remove.

But it's the right direction! Looking forward to the next iteration.

cqjjjzr:
I rebased the PR to `main` and made changes according to the @nitnelave 's suggestions.

1. Don't trust XFF headers, and use only the socket peer IP address
2. Decreased code duplication
3. Ensure the logout redirection actually work
4. Have the user specify CIDRs instead of single addresses, and validate the CIDR at startup
5. Prioritize trusted-header over JWT.

However, I wonder is it necessary to add an option to disable the JWT token authentication. This ensures even a malicious actor obtains an admin JWT token, without the proper access to the proxy authentication system, they can't gain admin access to the LLDAP admin page.

How should I submit my code here? Should I PR to @Kumpelinus 's repo or create a new PR here?

Issue #1290, opened 2025-09-15
[BUG] (Docker) Logs not respecting the Timezone


vic1707:
**Describe the bug**
The docker logs don't respect the timezone passed as a direct file or as the TZ env.

**To Reproduce**

The timezone is successfully detected when using either a volume mount or the TZ env
```
# No particular settings
[bofzilla@localhost ~]$ podman run  --entrypoint date lldap/lldap
Mon Sep 15 13:38:50 UTC 2025

# ENV
[bofzilla@localhost ~]$ podman run -e TZ=Europe/Paris  --entrypoint date lldap/lldap
Mon Sep 15 15:39:02 CEST 2025

# With direct mount of TZ file
[bofzilla@localhost ~]$ podman run -v /usr/share/zoneinfo/Europe/Paris:/etc/localtime:ro  --entrypoint date lldap/lldap
Mon Sep 15 15:44:29 CEST 2025
```

But when running the lldap container itself, the logs don't respect the timezone
```
# ENV
[bofzilla@localhost ~]$ podman run -e TZ=Europe/Paris -e LLDAP_JWT_SECRET='ofwehchnqoiheugbpqwiubxweipqufgpghfpewuhp[fhq3' lldap/lldap
[entrypoint] Copying the default config to /data/lldap_config.toml
[entrypoint] Edit this file to configure LLDAP.
> Setup permissions..
> Starting lldap..

Loading configuration from /data/lldap_config.toml
WARNING: A key_seed was given, we will ignore the key_file and generate one from the seed! Set key_file to an empty string in the config to silence this message.

thread 'main' panicked at server/src/main.rs:53:10:
The LDAP admin password must be initialized. Either set the `ldap_user_pass` config value or the `LLDAP_LDAP_USER_PASS` environment variable. A minimum of 8 characters is recommended.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-09-15T13:45:45.485704342+00:00  INFO     set_up_server [ 544ms | 100.00% ]
2025-09-15T13:45:45.487293348+00:00  INFO     ┝━ ｉ [info]: Starting LLDAP version 0.6.2
2025-09-15T13:45:45.655803149+00:00  INFO     ┝━ ｉ [info]: Upgrading DB schema from version 1
2025-09-15T13:45:45.656153149+00:00  INFO     ┝━ ｉ [info]: Upgrading DB schema to version 2
[...]

PR #1296, opened 2025-09-16
Local Tz logs

vic1707:
Fixes: #1290

Following the discussion we had in #1290, here's a proposition to make lldap logs abide by desired timezone.
This PR introduces a new `LLDAP_LOCAL_TZ_LOG` env variable to make the logs use the defined timezone instead of default UTC.

Due to how `tracing_subscriber` works, enabling `LLDAP_LOCAL_TZ_LOG` gives the same kind of logs as when `LLDAP_RAW_LOG` is enabled for simplicity as the `with_timer` method cannot be used without a `tracing_subscriber::fmt::Layer<S>` inner object ([doc](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_timer)). But another layer can be made if that behavior isn't desired.

This changes uses the `chrono` feature from `tracing-subscriber` instead of `local-time` because the later only yielded `<unknown time>` for the timestamp (apparently due to a `time` crate issue, see https://github.com/tokio-rs/tracing/issues/2715 for reference.

---

Results:
New `LLDAP_LOCAL_TZ_LOG`
image

Unchanged `LLDAP_RAW_LOG`
image

Unchanged default behavior
image

### DIFF
--- a/app/src/components/user_details_form.rs
+++ b/app/src/components/user_details_form.rs
@@ -214,7 +214,7 @@ fn format_attribute_value(attr_type: &AttributeType, raw: &str) -> String {
         AttributeType::DateTime => {
             console!(format!("Parsing date: {}", raw));
             chrono::DateTime::parse_from_rfc3339(raw)
-                .map(|dt| dt.to_local_time_display().to_string())
+                .map(|dt| dt.to_local_time_display())
                 .unwrap_or_else(|_| "Invalid date".to_string())
         }
         AttributeType::JpegPhoto => "Unimplemented JPEG display".to_string(),
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -17,14 +17,20 @@ pub fn run_app() -> Result<(), JsValue> {
 }
 
 trait DateToLocalDisplay {
-    fn to_local_date_display(&self) -> impl core::fmt::Display;
-    fn to_local_time_display(&self) -> impl core::fmt::Display;
+    fn to_local_date_display(&self) -> String;
+    fn to_local_time_display(&self) -> String;
 }
 impl<Tz: chrono::TimeZone> DateToLocalDisplay for chrono::DateTime<Tz> {
-    fn to_local_date_display(&self) -> impl core::fmt::Display {
-        self.with_timezone(&chrono::offset::Local).date_naive()
+    fn to_local_date_display(&self) -> String {
+        self.with_timezone(&chrono::offset::Local)
+            .date_naive()
+            .format("%Y-%m-%d")
+            .to_string()
     }
-    fn to_local_time_display(&self) -> impl core::fmt::Display {
-        self.with_timezone(&chrono::offset::Local).naive_local()
+    fn to_local_time_display(&self) -> String {
+        self.with_timezone(&chrono::offset::Local)
+            .naive_local()
+            .format("%Y-%m-%d %H:%M:%S")
+            .to_string()
     }
 }


server/src/logging.rs
```diff
@@ -43,13 +45,20 @@ pub fn init(config: &Configuration) -> anyhow::Result<()> {
             "sqlx=warn,reqwest=warn,info"
         })
     });
-    let registry = tracing_subscriber::registry()
-        .with(env_filter)
-        .with(tracing_forest::ForestLayer::default());
-    if env::var("LLDAP_RAW_LOG").is_ok() {
+    let registry = tracing_subscriber::registry().with(env_filter);
+
+    let raw_logs = env::var("LLDAP_RAW_LOG").is_ok();
```
nitnelave: That was a temporary workaround that I haven't gotten around to cleaning up, but it should go away.

server/src/logging.rs
```diff
@@ -43,13 +45,20 @@ pub fn init(config: &Configuration) -> anyhow::Result<()> {
...
+    let local_tz = env::var("LLDAP_LOCAL_TZ_LOG").is_ok();
```
nitnelave: I'm not convinced by this approach: I think the logs should always default to the local timezone (and you can control that with standard environment variables) if we don't want to log in UTC.

server/src/logging.rs
vic1707: If you want to explain what kind of workaround this was I can try to fix it for you so this PRs contains the final clean code

server/src/logging.rs
vic1707: No problem, I assumed you'd want UTC by default 👍

server/src/logging.rs
nitnelave: Basically, assume it's always false and that's the cleanup :)

Implied being that whatever logging solution we end up with, I'd like it to work with tracing forest

vic1707: Oh, a real cleanup then 👍 

But it seems like `tracing_forest` doesn't support `.with_timer` as it is only available from the bare `Layer` struct of `tracing_subscriber` and no equivalent seems to exist.

I believe this is a blocker ?

nitnelave: Yes, that's a blocker for me. 

We can raise an issue with the crate, they're really nice guys and reactive. 

Before we go much further though, I'd like to know how you plan on addressing the creation_date problem (timestamps in structs)

vic1707: didn't look that far yet as the blocker was really evident.

I'll check and see for myself asap, for now I can't seem to get the webui to show up after a `cargo run run` which is odd

nitnelave: Did you build the frontend? `./app/build.sh`

vic1707: found it a few seconds ago 👍

vic1707:
Regarding your concerns about the creation/update dates

image
image
image

everything looks fine to me, logs in my TZ but the db is still using UTC

frontend is showing UTC tho, wasn't that supposed to be automagically translated or did I misunderstood? 🤔 
image

nitnelave:
Huh, it's possible that the frontend displays in UTC, I didn't pay that close attention to it. Though I do believe I convert to local time! To be checked. 

For the logs, try turning on verbose mode, you'll see that user listing returns all the attributes, including the creation time.

vic1707:
looks good on verbose mode too 👍 
image

Will take a look at the frontend tomorrow (if fix is needed and I find it, maybe I'll include it in the PR as its the same topic)
Will also see for an issue on forest (except if you prefer to do it)
But that'll have to wait tomorrow 😴

vic1707:
Found a fix for the frontend
image

image

nitnelave:
Great! While you're at it, if it's a single place and it's not too hard, can we truncate the times to the second, on the web UI?

vic1707:
I believe we can, but its not in a single spot, I have 4 (or 5 see below).

[create_user.rs#L166](https://github.com/lldap/lldap/blob/main/app/src/components/create_user.rs#L166)
[group_table.rs#L125](https://github.com/lldap/lldap/blob/main/app/src/components/group_table.rs#L125)
[user_details_form.rs#L201](https://github.com/lldap/lldap/blob/main/app/src/components/user_details_form.rs#L201)
[user_table.rs#L134](https://github.com/lldap/lldap/blob/main/app/src/components/user_table.rs#L134)

And the _technically_ fifth one could be inserted here [user_details_form.rs#L168](https://github.com/lldap/lldap/blob/main/app/src/components/user_details_form.rs#L168) to mimic what's done line 201 (consistency)

I'm wondering if an extension trait should be done to centralize implementations

vic1707:
One thing that can also be done (but maybe in another commit/PR) is to make nice names for these date fields, they're the only ones in snake_case

nitnelave:
If we can have a central place for formatting dates, or even better a type that prints nicely, that would be ideal. 

And yes, nice names are a nice-to-have, but in a separate PR

nitnelave:
sorry if I missed something, are you waiting for my review?

vic1707:
I think we're mostly waiting for https://github.com/QnnOkabayashi/tracing-forest/issues/57
As this pr isn't compatible with tracing forest ATM
I believe you can still review it and I can split it to only include the frontend part (so not the CLI logs) if you want

nitnelave:
Sorry for the delayed response. @vic1707 I'm not sure having only the frontend migrated to the new timestamps is such a good idea, it can make for harder debugging if the backend doesn't match. But I could be convinced otherwise.

vic1707:
No need to convince you, I completely agree, either the whole feature lands or none of it, hopefully tracing-forest gets updated so I can get back to work here

PR #1316, opened 2025-10-05
Create action_button function for reusable link generation

webysther:
I think this fix the bug as alternative solution until upgrade the framework.


nitnelave:
Ah, different keys, that might do the trick! However, I think it'd be easier to have a local lambda in each place rather than a separate function.

webysther:
> Ah, different keys, that might do the trick! However, I think it'd be easier to have a local lambda in each place rather than a separate function.

Be my guest, do whatever you think is best, my code is just a patch I applied to check if works.

nitnelave:
I re-implemented it with lambda (but kept you as the commit author, since you had the crucial insight)

PR #1325, opened 2025-10-11
server: allow specifying the healthcheck addresses

### LINKED ISSUE #700: Bug: Listening to an address other than `127.0.0.1/32` or `0.0.0.0/32` breaks docker health checks
Forgive me if this isn't possible/within the scope of LLDAP. I tried to look through the code briefly to see whether I could figure it out on my own, but I'm not familiar enough with rust to find it easily.

#### The issue:
When binding the LDAP IP address to a specific IPv4 address (`LLDAP_LDAP_HOST = "192.168.2.2"`, which is technically `192.168.2.2/32` - unless LLDAP does something weird), docker health checks fail.

#### Why it happens:
It's simple in nature: The docker health check attempts to reach the LDAP api on `127.0.0.1/32`, but gets refused - LDAP is listening on `192.168.2.2/32`. Thus, it's request is refused.

#### The solution:
Unless it is a case of user error, I see three ways to resolve this:
- Allow listening on two different IPv4/6 addresses 
    - I find this to be unlikely, because I'm not actually sure if the Linux kernel allows this
-  Always listen on `127.0.0.1/32` when dockerized
    - Not necessarily ideal, since it adds complexity to development - namely fragmentation
- Use the configured `LLDAP_LDAP_HOST` address for docker healthchecks
    - This seems to be the sanest option imho, though I'm not sure if there are security/bandwidth/other concerns

Again, forgive me if I'm incorrect - I wasn't able to find any information in the issues, readme, or code regarding this.

sdwilsh:
This change adds two new optional configuration options:
- `ldap_healthcheck_host` to pair with `ldap_host`
- `http_healthcheck_host` to pair with `http_host`

These new options will allow someone to specify a specific address for `ldap_host` and `http_host` and still have functional health checks, as health checks were previously hard-coded to check against `localhost` prior to this change.  These new settings default to `localhost` to preserve the existing behavior.

Fixes #700

### DIFF
--- a/server/src/configuration.rs
+++ b/server/src/configuration.rs
@@ -1,7 +1,7 @@
 use crate::{
     cli::{
-        GeneralConfigOpts, LdapsOpts, RunOpts, SmtpEncryption, SmtpOpts, TestEmailOpts,
-        TrueFalseAlways,
+        GeneralConfigOpts, HealthcheckOpts, LdapsOpts, RunOpts, SmtpEncryption, SmtpOpts,
+        TestEmailOpts, TrueFalseAlways,
     },
     database_string::DatabaseUrl,
 };
@@ -83,6 +83,21 @@ impl std::default::Default for LdapsOptions {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, derive_builder::Builder)]
+#[builder(pattern = "owned")]
+pub struct HealthcheckOptions {
+    #[builder(default = r#"String::from("localhost")"#)]
+    pub http_host: String,
+    #[builder(default = r#"String::from("localhost")"#)]
+    pub ldap_host: String,
+}
+
+impl std::default::Default for HealthcheckOptions {
+    fn default() -> Self {
+        HealthcheckOptionsBuilder::default().build().unwrap()
+    }
+}
+
 #[derive(Clone, Deserialize, Serialize, derive_more::Debug)]
 #[debug(r#""{_0}""#)]
 pub struct HttpUrl(pub Url);
@@ -138,6 +153,8 @@ pub struct Configuration {
     #[serde(skip)]
     #[builder(field(private), default = "None")]
     server_setup: Option<ServerSetupConfig>,
+    #[builder(default)]
+    pub healthcheck_options: HealthcheckOptions,
 }
 
 impl std::default::Default for Configuration {
@@ -520,6 +537,18 @@ impl ConfigOverrider for SmtpOpts {
             .inspect(|&enable_password_reset| {
                 config.smtp_options.enable_password_reset = enable_password_reset;
             });
+    }
+}
+
+impl ConfigOverrider for HealthcheckOpts {
+    fn override_config(&self, config: &mut Configuration) {
+        self.healthcheck_http_host
+            .as_ref()
+            .inspect(|host| config.healthcheck_options.http_host.clone_from(host));
+
+        self.healthcheck_ldap_host
+            .as_ref()
+            .inspect(|host| config.healthcheck_options.ldap_host.clone_from(host));
     }
 }
 
--- a/server/src/healthcheck.rs
+++ b/server/src/healthcheck.rs
@@ -70,8 +70,8 @@ where
 }
 
 #[instrument(level = "info", err)]
-pub async fn check_ldap(port: u16) -> Result<()> {
-    check_ldap_endpoint(TcpStream::connect(format!("localhost:{port}")).await?).await
+pub async fn check_ldap(host: &str, port: u16) -> Result<()> {
+    check_ldap_endpoint(TcpStream::connect((host, port)).await?).await
 }
 
 fn get_root_certificates() -> rustls::RootCertStore {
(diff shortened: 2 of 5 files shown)


sdwilsh:
I can also update `lldap_config.docker_template.toml`, but wasn't sure if we wanted this additional complexity documented there or not, so I held off on doing so for now.

server/src/configuration.rs
```diff
@@ -92,10 +92,14 @@ pub struct HttpUrl(pub Url);
 pub struct Configuration {
     #[builder(default = r#"String::from("0.0.0.0")"#)]
     pub ldap_host: String,
+    #[builder(default = r#"String::from("localhost")"#)]
+    pub ldap_healthcheck_host: String,
```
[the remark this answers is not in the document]
sdwilsh: I had avoided doing this because `ldap_host` and `http_host` were not done this way here (which is maybe a bug in itself?).  Happy to address, but would need guidance from someone with more context.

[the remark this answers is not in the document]
nitnelave: Yes, let's add the new arguments. I'll have a look whether ldap_host missing is a bug or not.

Make sure to rebase, I _just_ updated that part of the code to use `inspect` instead of `if let Some`

server/src/healthcheck.rs
```diff
@@ -70,8 +70,11 @@ where
 }
 
 #[instrument(level = "info", err)]
-pub async fn check_ldap(port: u16) -> Result<()> {
-    check_ldap_endpoint(TcpStream::connect(format!("localhost:{port}")).await?).await
+pub async fn check_ldap(address: &str, port: u16) -> Result<()> {
+    check_ldap_endpoint(
+        TcpStream::connect(format!("{}:{}", get_ipv6_url_safe_address(address), port)).await?,
```
nitnelave: `connect` can take a `(&str, u16)` as argument, that seems like a better match. Then you don't have to muck around with IP parsing, V4 vs V6 and all that

nitnelave:
> I can also update `lldap_config.docker_template.toml`, but wasn't sure if we wanted this additional complexity documented there or not, so I held off on doing so for now.

Ah, I just saw your comment. Yes, we want docs there!

server/src/healthcheck.rs
sdwilsh: I was led astray by testing `"localhost".to_socket_addrs()` (which fails).  `("localhost", 80).to_socket_addrs()` actually parses out to two addresses (IPv4 and IPv6 cases), which is fine for our uses.  Unfortunately, `reqwest::get` doesn't have smarts like this, so we still need to keep `get_ipv6_url_safe_address` around.

server/src/configuration.rs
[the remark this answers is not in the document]
sdwilsh: @nitnelave, I think maybe you have not pushed those changes to `main` yet, as that is still showing the `if let Some` syntax.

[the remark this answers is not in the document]
nitnelave: My bad, just did now

nitnelave: On second thought, would it make sense to have a separate healthcheck section for the config? Like we have for LDAPS and so on.

That would make the top-level configuration simpler, and group the healthcheck options for clearer access. And that way if we have other options in the future, we can add them there (configurable timeout, error code or whatever)

server/src/healthcheck.rs
```diff
@@ -151,10 +149,42 @@ pub async fn check_ldaps(ldaps_options: &LdapsOptions) -> Result<()> {
...
 
 #[instrument(level = "info", err)]
-pub async fn check_api(port: u16) -> Result<()> {
-    reqwest::get(format!("http://localhost:{port}/health"))
-        .await?
-        .error_for_status()?;
+pub async fn check_api(address: &str, port: u16) -> Result<()> {
+    reqwest::get(format!(
+        "http://{}:{}/health",
+        get_ipv6_url_safe_address(address),
```
nitnelave: I'd rather document that than make a needlessly complicated parsing method. An IPv6 hostname should be in square brackets.

server/src/configuration.rs
sdwilsh: I did consider doing this initially, but I didn't for a few reasons:
- It wasn't clear why some stuff was pulled out into groups vs not
- if someone changes `http_host` and `ldap_host`, they almost certainly need to update the health check variants as well (especially in docker since the health check is declared in the image file; kubernetes doesn't care about those though so you gotta do it yourself anyway)
- it seemed odd to have something outside of the global section that would impact the health check for `ldaps`, which has its own section

That's a lot of words for something I don't actually have a strong opinion on, however.  If you want it in a separate section, I can make that change.

server/src/healthcheck.rs
sdwilsh: I didn't even consider making it a documentation-only thing...and I should have.  Thanks for pointing that out.

server/src/configuration.rs
nitnelave: I would say that this is not the case, that everyone who sets http_host wants to set the healthcheck host. Actually it's even counter-productive: a (docker) health check is whether LLDAP responds to http/LDAP requests, not whether it is reachable from a given URL. That would make docker think that LLDAP is not up when your reverse proxy breaks, for instance. You usually want localhost, which is why it was initially not configurable.

But I can imagine running the healthcheck as a non-docker health check, rather like a system health check, in which case I agree that it should be configurable. However, I expect it to be a rather fringe use: there are much better uptime monitors out there.

I think it makes sense to make it a separate section. Whether the healthcheck part should be in the LDAPS section or the opposite is arguable, but I think it fits better in the healthcheck section.

sdwilsh: In my case, and of @TheRealGramdalf per their comment in the issue, setting the health check is more because we set the `http_host` and `ldap_host` urls to a specific IP address instead of binding to all addresses.  Once you do that, `localhost` doesn't work for a health check because nothing is listening on that address, and it's my motivation for fixing this.  That doesn't run into the issue you mention, because it's still the local host when you specify an IP it has (it it was a non-local IP, the software would never start because it couldn't bind to that IP and port).

I hadn't considered using the `lldap` binary to run a health check outside, but it totally makes sense to be separate then.

PR #1330, opened 2025-10-16
initial gogs.md documentation

thchha:
I setup Gogs and immediately configured it to make use of lldap.
This PR just documents my configuration.

Please let me know if I should further elaborate on the provided values or any other shortcomings.
After accepting the PR the documentation is entirely yours but may get corrected by me once the filter is working.

> commit message:

Gogs appears to be in maintenance mode.
So my current configuration can be extended in the future.

### DIFF
--- a/example_configs/gogs.md
+++ b/example_configs/gogs.md
@@ -1,0 +1,46 @@
+# Gogs LDAP configuration
+
+Gogs can make use of LDAP and therefore lldap.
+
+The following configuration is adapted from the example configuration at [their repository](https://github.com/gogs/gogs/blob/main/conf/auth.d/ldap_bind_dn.conf.example).
+The example is a container configuration - the file should live within `conf/auth.d/some_name.conf`:
+
+```yaml
+$ cat /srv/git/gogs/conf/auth.d/ldap_bind_dn.conf
+id           = 101
+type         = ldap_bind_dn
+name         = LDAP BindDN
+is_activated = true
+is_default   = true
+
+[config]
+host               = ldap.example.com
+port               = 6360
+# 0 - Unencrypted, 1 - LDAPS, 2 - StartTLS
+security_protocol  = 1
+# You either need to install the LDAPS certificate into your trust store -
+# Or skip verification altogether - for a restricted container deployment a sane default.
+skip_verify        = true
+bind_dn            = uid=<binduser>,ou=people,dc=example,dc=com
+bind_password      = `yourPasswordInBackticks`
+user_base          = dc=example,dc=com
+attribute_username = uid
+attribute_name     = givenName
+attribute_surname  = sn
+attribute_mail     = mail
+attributes_in_bind = false
+# restricts on the `user_base`.
+filter             = (&(objectClass=person)(uid=%s))
+# The initial administrator has to enable admin privileges.
+# This is only possible for users who were logged in once.
+# This renders the following filter obsolete; Though its response is accepted by Gogs.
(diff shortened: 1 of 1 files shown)


example_configs/gogs.md
```diff
@@ -0,0 +1,44 @@
+# Gogs LDAP configuration
+
+Gogs can make use of LDAP and therefore lldap.
+
+The following configuration is adapted from the example configuration at [their repository](https://github.com/thchha/lldap/new/main/example_configs).
+The example is a container configuration - the file should just needs to live within `conf/auth.d/some_name.conf`:
+
+```
```
thchha: No effect so omitted.

example_configs/gogs.md
```diff
@@ -0,0 +1,44 @@
+# Gogs LDAP configuration
+
+Gogs can make use of LDAP and therefore lldap.
+
+The following configuration is adapted from the example configuration at [their repository](https://github.com/thchha/lldap/new/main/example_configs).
```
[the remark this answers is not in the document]
nitnelave: The bot seems to have a point, that's not a valid link

example_configs/gogs.md
nitnelave: Wouldn't it be toml? If you remove the `cat` line

example_configs/gogs.md
```diff
@@ -0,0 +1,44 @@
...
+name         = LDAP BindDN
+is_activated = true
+is_default   = true
+
+[config]
+host               = ldap.example.com
+port               = 6360
+# 0 - Unencrypted, 1 - LDAPS, 2 - StartTLS
+security_protocol  = 1
+skip_verify        = true
```
nitnelave, line 23: Is that for self-sifned certificates? Can you mention that in a comment?

nitnelave (approved):
Looks good! Other than a couple of nits and an invalid link, LGTM

example_configs/gogs.md
thchha: There are only these languages according to github's documentation: [github fks with my texts - here](https://github.com/github/docs/blob/main/data/code-languages.yml)

I will use yaml since this will at least differ between comments and key-value pairs/properties.

// edit:
I didn't integrate/test your suggestion correctly.
Maybe your toml rendering would have worked flawlessly.
Anyhow: Explictily stating the mount point - even though we are not within the official gogs repository - is a valuable information for the quick reader.
We are pointing him to a working solution - the symbolic links within the gogs container are kind of obscure at first/quick glance. But you may judge about it:

Should I proceed with yaml syntax highlighting and omit the line?

nitnelave: Have you tried with toml? I found a different source that lists a bunch more languages, including toml: https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll#syntax-highlighting

thchha: I tried.
I tried toml, shell, bash and then settled on yaml.

I just tried toml and systemd without the prepended $ cat line (`# $ cat rest`) but it decreases rendering.

// edit:

`toml` renders red background within the line at the start of syntax errors.
My picks will just render it one coloured.

So yaml appears to be the pick which at least makes use of comment highlighting.

nitnelave: Yaml it is, then. 

Thanks for trying!

thchha:
Ready to merge; Please revisit the commit message before merging, though.

I do not want to have any rights remaining on this contribution; Just so you know.

nitnelave (approved):

nitnelave:
Thanks!

PR #1334, opened 2025-10-22
server: Adding User Manager group (`lldap_user_manager`)

asalimonov:
Hello! Can you please consider this PR for adding a role for managing of users?

Use case: we want to grant access rights for several persons who should be responsible for onboarding of new employees. Usually they are HR managers or assistants. They shouldn't be able to modify of admins and other user managers, just manage regular accounts.

User Manager can:
- create new users
- delete users
- modify users
- change user passwords
- add a users to groups

User Manager cannot:
- add users to the 'lldap_admin' or 'lldap_user_manager' groups
- remove users from the 'lldap_admin' or 'lldap_user_manager' groups
- delete users who belongs to the 'lldap_admin' and 'lldap_user_manager' groups
- change passwords of a users who belongs to the 'lldap_admin' and 'lldap_user_manager' groups

Added several tests to verify behavior.
Built and tested docker image. Added several screenshots.
Would be very thankful for review and accept!
Screenshot 2025-10-22 at 09 07 47
Screenshot 2025-10-22 at 09 08 02
Screenshot 2025-10-22 at 09 10 28

### DIFF
--- a/app/src/components/app.rs
+++ b/app/src/components/app.rs
@@ -66,21 +66,17 @@ impl Component for App {
                     None
                 })
                 .and_then(|u| {
-                    get_cookie("is_admin")
-                        .and_then(|is_admin_opt| {
-                            get_cookie("is_user_manager")
-                                .map(|is_user_manager_opt| {
-                                    is_admin_opt.and_then(|is_admin_str| {
-                                        is_user_manager_opt.map(|is_user_manager_str| {
-                                            (u, is_admin_str == "true", is_user_manager_str == "true")
-                                        })
-                                    })
-                                })
-                        })
-                        .unwrap_or_else(|e| {
-                            error!(&e.to_string());
-                            None
-                        })
+                    let is_admin = get_cookie("is_admin")
+                        .ok()
+                        .flatten()
+                        .map(|s| s == "true")
+                        .unwrap_or(false);
+                    let is_user_manager = get_cookie("is_user_manager")
+                        .ok()
+                        .flatten()
+                        .map(|s| s == "true")
+                        .unwrap_or(false);
+                    Some((u, is_admin, is_user_manager))
                 }),
             redirect_to: Self::get_redirect_route(ctx),
             password_reset_enabled: None,
@@ -319,7 +315,7 @@ impl App {
     fn is_user_manager(&self) -> bool {
         match &self.user_info {
             None => false,
-            Some((_, _, is_user_manager)) => *is_user_manager,
+            Some((_, is_admin, is_user_manager)) => *is_admin || *is_user_manager,
         }
     }
 }
--- a/app/src/infra/api.rs
+++ b/app/src/infra/api.rs
@@ -82,10 +82,12 @@ fn set_cookies_from_jwt(response: login::ServerLoginResponse) -> Result<(String,
     let is_admin = jwt_claims.groups.contains("lldap_admin");
     let is_user_manager = jwt_claims.groups.contains("lldap_user_manager");
     set_cookie("user_id", &jwt_claims.user, &jwt_claims.exp)
-        .map(|_| set_cookie("is_admin", &is_admin.to_string(), &jwt_claims.exp))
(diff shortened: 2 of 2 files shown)


nitnelave:
It would have helped (and saved you time) to reach out with an issue, or on discord to discuss it in advance. Then we could have talked about how we can achieve this, whether it fits in the scope of the project and how to achieve it. 

I'm afraid I can't merge this PR as is (or accept this feature request). The added complexity/difficulty of maintenance is not worth it.

In terms of technical implementation, before adding a new boolean to the frontend I would have asked to properly implement role-based permissions, fixing the "is_admin" existing code.

What I can suggest instead is to wait for the plugin API (you can of course use this PR locally until then). Once you can write a plugin, it should be easy to add people to both admin and user - manager groups, and then block  unauthorized requests in the plugin.

asalimonov:
>I'm afraid I can't merge this PR as is (or accept this feature request). The added complexity/difficulty of maintenance is not worth it.

It's sad, we thought this change wouldn’t add much complexity.

> In terms of technical implementation, before adding a new boolean to the frontend I would have asked to properly implement role-based permissions, fixing the "is_admin" existing code.

I agree that the current implementation is required some refactoring toward granular RBAC permissions, not flags. However, that would be a major change. This PR aims to introduce minimal code and reuse existing tests with minimal impact as possible.

nitnelave:
It's not entirely about the complexity of the code (though for the frontend it is).

It is also about the complexity of the feature for the project. Consider the target user base: self-hosters with a home lab, at home or on a VPS. They know all their users, they don't want to think about complex permissions, or even learn about LDAP except to the extent required to make their integrations work. We already have problems with people who don't understand why a role with password manager permissions is not allowed to change the password of an admin (trivial escalation). If we introduce another role that can create/delete users except for other privileged roles, it'll create more confusion, for something that most people don't need. And should that role be able to change the passwords of other users? I'd say no, but others might disagree.
Adding to the complexity, the roles are not strict supersets of each other now, so you have to consider a user with both groups, what does that do?

In terms of implementation, the accessors could mostly just work at the request level, either blocking completely, adding a filter, or there would be the one edge case of the password manager not being able to change the admin password. Now almost all the queries for that new groups must check for other permissions (admin). And what's the interaction with password managers? User managers shouldn't be able to add the permission (escalation), but password managers shouldn't be able to change the passwords of user managers either (another escalation).

I could go on, but I'm sure you can now appreciate that it would add a significant documentation/edge case/implementation/support complexity to the project, for a corner-case feature.

However, as a plugin it's a great idea! It's opt-in complexity for the people who are interested in the feature.

I'd love to direct you to our plugin API, but development has been slow (we're trying to get most of the difficult questions right before having to break all the existing plugins, even though we'll probably at least partly fail. That and limited time)

asalimonov:
Thank you very much for the quick and detailed explanation! It’s really appreciated.
I completely agree with you about maintaining simplicity in the **product** for the target users. I also fully support strict code reviews and careful examination of execution flow to avoid issues avoid escalation in A&A systems. 

From my experience with large projects involving IAM systems, it’s almost impossible to keep everything simple inside an RBAC implementation — even though that should always be the main goal. Over time, “magic” IDs for certain service accounts or roles tend to appear, breaking inheritance rules because of distinctions between “management accounts” and “managed accounts.” Of course, LLDAP is not an IAM for large system, I just wanted to share my thoughts on this particular case.

>I could go on, but I'm sure you can now appreciate that it would add a significant documentation/edge case/implementation/support complexity to the project, for a corner-case feature.

I was actually expecting a request for documentation, and I'm ready to write it along with tests for edge cases. That would still be simpler than maintaining a fork or an external patch set.

I don't know anything about plugins in LLDAP, and I don’t have an opinion on it at the moment — but thank you for bringing it up. Thank you for sharing information, need some time to think about it.

asalimonov:
I decided to close this PR after some thought. The change is quite small, but it introduces specific logic that shouldn’t be part of a directory service. I plan to implement this logic in https://github.com/asalimonov/authelia-admin as the next step in integrating with LLDAP via GraphQL.
@nitnelave, thank you for sharing your thoughts and your time.

Issue #1356, opened 2025-11-21
Missing required star (mail) when creating an user


jisef:
Next to mail is no red star, but a mail is required to create a user

Is this wanted?
If not i could make a PR
<img width="1336" height="594" alt="Image" src="https://github.com/user-attachments/assets/b8b31017-561d-44c4-a735-24b737b2ade4" />

I did not use an template since this is not really a bug, more like a small enhancement

nitnelave:
Yes, there should be a star. A PR would be most welcome!

PR #1358, opened 2025-11-22
app: asterisk for mail in create_user.rs #1356

jisef:
[Issue](https://github.com/lldap/lldap/issues/1356)

- Made it possible to make attributes required (Red asterisk next to the label)

### DIFF
--- a/app/src/components/create_user.rs
+++ b/app/src/components/create_user.rs
@@ -304,7 +304,7 @@ impl Component for CreateUserForm {
 }
 
 fn get_custom_attribute_input(attribute_schema: &Attribute) -> Html {
-    let mail_is_required = &attribute_schema.name.as_str() == &"mail";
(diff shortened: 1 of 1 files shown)


nitnelave:
Quick comment on the last commit: it's not that I don't want the project to be accessible, but that would be the _only_ aria label in the entire project 😄 

I'd rather do it well or not at all than only do it in one place.

app/src/components/create_user.rs
```diff
@@ -303,19 +303,25 @@ impl Component for CreateUserForm {
     }
 }
 
+pub const REQUIRED_USER_ATTRIBUTES: &[&str] = &["mail"];
```
nitnelave: Let's inline the value of this, it's going to be just "mail", we don't need a constant.

PR #1379, opened 2026-01-07
docs(bootstrap): add `password_file` for user configs

OsiPog:
`password_file` was not documented, so I documented it


nitnelave:
Thanks!

PR #1387, opened 2026-01-24
Installing and Configuring LLDAP on TrueNAS

mreid-tt:
Add setup guide for LLDAP on TrueNAS


nitnelave:
LGTM! Can you add a section in https://github.com/lldap/lldap/blob/main/docs/install.md and a link to that section in the Readme?

PR #1392, opened 2026-01-31
feat(server): add native HTTPS support for the web interface

lyzstrik:
This commit enables the web server to run natively over HTTPS, allowing secure deployments without requiring a reverse proxy.

Key changes:
- Configuration: Added `[https_options]` to config and CLI arguments (port, cert/key paths, enabled flag).
- Server: Updated `tcp_server.rs` to support binding with TLS.
- Healthcheck: Added `check_https` which uses certificate pinning to verify identity.
- Main: Updated healthcheck runner to check HTTP and HTTPS endpoints concurrently.
- Docs: Updated `install.md` and `lldap_config.docker_template.toml` with HTTPS configuration examples.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2858,7 +2858,6 @@ dependencies = [
  "pretty_assertions",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
- "reqwest 0.11.27",
  "reqwest 0.12.28",
  "rustls 0.23.35",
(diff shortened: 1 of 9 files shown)


nitnelave:
This is something I've been purposefully avoiding. Setting up a reverse proxy in front is not very hard and not resource intensive, so I chose to avoid complicating the code of LLDAP.

PR #1398, opened 2026-02-18
example_config: fix HA for blank displayName

santigz:
Fix example_config for Home Assistant.
Use the user .id when the .displayName is not present.


nitnelave (approved):
Thanks!

PR #1423, opened 2026-03-30
feat(auth): upgrade opaque-ke from 0.7 to 4.0 (RFC 9807)

### LINKED ISSUE #111: Update Opaque-ke dependency
As part of their new API, the serializer doesn't write binary as base64 in JSON anymore. We should switch the json serializer to a json+binary-to-base64 one.

fredericrous:
Progressive migration from opaque-ke 0.7 to 4.0 (RFC 9807 / Option 3). Existing passwords remain valid and are automatically upgraded to the new format on next login — no forced password reset, no user-visible change, and **no admin action beyond deploying the new version** (the earlier `--force-update-private-key` step is gone, see "Automatic server key upgrade" below).

## How to review: commit-by-commit guide

The branch is 8 conventional commits ordered bottom-up (plus post-review fix commits kept separate so the deltas stay visible); every commit compiles and is bisectable. Each layer only depends on the ones before it:

| Commit | What to review |
|---|---|
| `8e779c9` feat(auth) | opaque-ke 4.0 cipher suite + the self-contained `lldap_auth::v07` compat layer (blind wrapper types; no other crate imports opaque-ke 0.7). Crypto only, no policy. |
| `997d030` feat(sql) | Migration policy: schema v12 `password_version` column, version-aware `bind()` with inline auto-upgrade, v0.7 web-login handshake, orion-sealed login state. |
| `a63f95c` feat(server) | The big one (~1.1k lines): v0.7 HTTP routes + 409 mapping, **automatic key upgrade** (detection, DB confirmation, sidecar, deferred atomic rotation) in `configuration.rs`/`main.rs`, WASM client fallback, set-password/migration-tool on the 4.0 client API. The commit message maps each bullet to its files. |
| `b1dc2a5` fix(auth) | Username-enumeration fix on the v0.7 endpoint (dummy handshake for both unknown and v4.0 users). Small and security-relevant — review in isolation. |
| `0820b1a` feat(migration-tool) | Same 409 fallback as the web client. |
| `3ec0d43` test(auth) | End-to-end upgrade test (real baseline binary → HEAD, file + seed key modes). |
| `e27aa36` docs | Migration guide + CHANGELOG. |
| `9ee3ab4` feat(app) | Admin visibility: admin-only `hasLegacyPassword` GraphQL field + "Password upgrade pending" badge and count in the user list (RFC's optional admin indicator). |
| `c22347f` fix(auth) | Post-review: CodeRabbit findings (generic v0.7 auth error, sidecar IO logging, no `unreachable!`, e2e retry). |
| `ec2c883` fix(auth) | Post-review: base64 wire encoding for OPAQUE messages in the JSON API (#111's original request; see Wire contract). |

## RFC Option 3 conformance

| Requirement | Implementation |
|---|---|
| No admin action beyond updating | The opaque-ke 0.7 → 4.0 key change is detected and applied automatically at startup; no flag, no extra restart. |
| Password version column | `password_version INT NOT NULL DEFAULT 0` added in migration v12 (0 = v0.7, 1 = v4.0). |
| Dual opaque-ke dependencies | `opaque-ke 0.7` (aliased `opaque-ke-v07`) coexists with `opaque-ke 4.0`. All v0.7 crypto is centralized in `lldap_auth::v07`, which exposes opaque wrapper types (`V07ServerSetup`, `V07ServerLoginState`, `V07ClientLoginState`) so neither `sql-backend-handler` nor `app` import opaque-ke v0.7 directly. `sql-backend-handler` owns migration policy (password version, upgrade timing, state encryption). `domain-model` has `DomainError::OpaqueV07Version`, and `server` exposes the v0.7 HTTP endpoints. |
| "Old OPAQUE version" error signal | `DomainError::OpaqueV07Version` mapped to **HTTP 409 Conflict** with body `{"error_code":"opaque_v07_version"}`. |
| Client fallback + silent re-registration | WASM client parses 409 `error_code`, falls back to `/auth/opaque/v07/login/{start,finish}` (opaque-ke 0.7 handshake), then immediately re-registers via the v4.0 registration flow. Re-registration failure is best-effort — login still succeeds, retry on next login. |
| LDAP bind transparent upgrade | `bind()` has the plaintext password; on v0.7 validation success it calls `upgrade_password()` inline (awaited, not background) to re-register with opaque-ke 4.0. |
| Simple HTTP login transparent upgrade | `simple_login` calls `bind()` under the hood — same upgrade path. |
| No visual change in web UI | Login and change-password components handle the fallback internally. |
| Startup warning for remaining v0.7 users | `server/src/main.rs` queries `password_version = 0` count at startup and emits a `warn!` with context about whether the v0.7 key is available. |
| Admin UI indicator for v0.7 users | **Implemented** — admin-only `hasLegacyPassword` GraphQL field (`null` for non-admins, gated server-side per-field) backing a "Password upgrade pending" badge + remaining-count banner in the user list. This tells admins when it's safe to eventually drop v0.7 support. |

## Wire contract

The 409 body uses a structured error code so clients don't need to string-match:

```json
{"error_code":"opaque_v07_version"}
```

Clients that receive this response should fall back to the v0.7 endpoints at `/auth/opaque/v07/login/start` and `/auth/opaque/v07/login/finish`. The `error_code` check is intentional — HTTP status alone is insufficient because 409 may be reused.

**Message encoding**: all OPAQUE message fields in the JSON API (v4.0 and v0.7 endpoints alike) carry the RFC 9807 protocol serialization as a **base64 string** — the same JSON shape as the pre-upgrade API, and the encoding #111 originally asked for. opaque-ke 4.0's derived serde impls would otherwise have leaked its internal struct layout (nested integer arrays) into the public API; a `base64_wire` serde adapter in `lldap_auth` prevents that and lets non-Rust clients build messages from raw protocol bytes.

## Automatic server key upgrade (no flag, no restart)

The server's `ServerSetup` key also changes format incompatibly between opaque-ke 0.7 and 4.0 (the RFC only covers password hashes). Rather than reusing `--force-update-private-key` — whose documented meaning is "invalidate every password" — the upgrade is detected positively and applied automatically:

- **Detection**: a key file that fails 4.0 deserialization is parsed as a v0.7 `ServerSetup`. A corrupted file parses as neither and aborts startup *without touching the file*.
- **Confirmation**: pre-4.0 servers stored `stable_hash(<v0.7 private key bytes>)` in the DB; the startup check recomputes that hash from the detected v0.7 key and only accepts the change if it matches — proving it's the key from the last successful startup. A valid-but-wrong v0.7 key (e.g. wrong file restored) is rejected.
- **Deferred, atomic rotation**: nothing is written at config-load time (`healthcheck`/`test-email` can never rotate the key). Once confirmed, the old key is saved to a `<keyfile>.v07` sidecar *before* the key file is atomically replaced (temp file + rename), so a sidecar write failure can't destroy the only copy. The sidecar is loaded on subsequent startups and auto-deleted once `count_v07_passwords() == 0`. A crash between rotation and the DB update self-heals on the next startup (the stored hash still matches the sidecar key).
- **Seed-based deployments**: the v0.7 key is re-derived byte-for-byte from the same `key_seed` in memory on every start — no file, no sidecar; the same DB-hash confirmation applies.
- `--force-update-private-key` keeps its existing, purely destructive meaning and is **not** part of this upgrade.

## Upgrade path

```bash
# 1. Stop lldap
# 2. Deploy the new binary
# 3. Start normally:
lldap run
#    → detects the v0.7 key, confirms it against the DB, rotates it
#      (file mode: writes <keyfile>.v07 sidecar; seed mode: nothing on disk)
#    → users with v0.7 passwords are upgraded on next login
#    → once all users are upgraded, the sidecar is auto-deleted
```

No flags, no forced password resets, no user-visible change.

## Tests

**Unit tests in `sql_opaque_handler.rs`:**
- `test_v07_bind_auto_upgrade` — v0.7 bind succeeds and upgrades `password_version` to 1
- `test_v07_bind_wrong_password` — wrong password stays at 0 (no upgrade on failure)
- `test_new_registration_sets_v1` — fresh registration writes `password_version = 1`
- `test_mixed_users_both_bind` — v0 and v1 users coexist correctly
- `test_opaque_login_start_returns_v07_version_error` — v4.0 login start on v0.7 user returns `OpaqueV07Version`
- `test_v07_opaque_login_full_flow` — full v0.7 handshake → re-register → v4.0 login → `password_version = 1`
- `test_v07_login_wrong_password_fails` — v0.7 endpoint rejects bad password
- `test_v07_login_rejected_for_v4_user` — v4 user on v0.7 endpoint is rejected
- `test_corrupted_bytes_rejected` — corrupted password bytes are rejected
- `test_new_format_roundtrip` — v4.0 ServerRegistration serialization roundtrip
- `test_v07_format_compat` — v0.7 password file format compatibility

**Key-upgrade tests in `server/src/configuration.rs`:**
- corrupted key file (parses as neither version) errors without touching the file
- valid v0.7 key file is detected with **no disk writes at config-load time**; persisting the rotation writes the sidecar + rotates atomically
- the sidecar is reloaded on subsequent startups and cleaned up once all users are upgraded

**GraphQL access-control tests:** `hasLegacyPassword` returns real values for admins and `null` for non-admin callers.

**End-to-end upgrade test** (`scripts/test_opaque_upgrade.sh`):

Builds two `lldap` binaries from this repo at two different commits — a baseline (the parent of the migration commit, pinned at `bb2ea7b`) and HEAD — then exercises the full upgrade sequence for **both** key-storage modes (file and seed) against the same SQLite DB:

1. Baseline creates a test user with a known password (opaque-ke 0.7 credential)
2. HEAD starts **directly** (no flag, no extra restart) and reports the automatic upgrade; file mode asserts the key file was rotated and the `<keyfile>.v07` sidecar holds the original key, seed mode asserts nothing was written to disk
3. Admin GraphQL query asserts `hasLegacyPassword = true` for the test user (the admin login itself exercises the simple-login upgrade path)
4. LDAP bind with the original password succeeds and auto-upgrades `password_version` 0 → 1; `hasLegacyPassword` flips to `false`
5. Second bind with the upgraded credential succeeds
6. Wrong password is rejected

This test caught a real bug: the v0.7 key material was lost across restarts, fixed by the sidecar file persistence in this PR.

```bash
# Run locally (requires: cargo, git, curl, jq, sqlite3, ldapsearch)
# First run builds both binaries (~10-15 min); subsequent runs use cargo cache (~1 min)
bash scripts/test_opaque_upgrade.sh
```

## Migration safety

DB migration v12 adds `password_version INT NOT NULL DEFAULT 0` to the `users` table. Applied via the existing migration framework (idempotent). Downgrade-safe: old binaries only SELECT the columns they know about; the new column is ignored. No data is touched in migration — passwords are upgraded lazily on login.

The v0.7 key sidecar (`<keyfile>.v07`) is created during the automatic rotation and auto-deleted when no v0.7 users remain. If the file is manually deleted before all users are upgraded, the affected users will need an admin password reset — the startup warning makes this explicit.

## Known limitations / follow-ups

- **`change_password` with a v0.7 password** — requires the user to log out and back in first (the web UI shows an explanatory error message). A follow-up could handle the re-auth flow inline.
- **WASM bundle size impact** — raw wasm `2.89 MB → 3.01 MB` (+4.1%), gzipped `921.6 KB → 974.0 KB` (+5.6%). The increase comes from opaque-ke 0.7 and its transitive deps added as WASM deps for the v0.7 handshake. These can be removed once the migration window closes.
- **LDAP bind upgrade is inline** — `upgrade_password()` is awaited inside `bind()`, adding Argon2id latency (~100ms) to the first successful login after migration. Subsequent logins use the v4.0 path with no overhead.
- **Deprecation phase** (future release, per the RFC): switch the startup warning from a count to listing the users, then eventually drop the v0.7 dependency — the `hasLegacyPassword` admin visibility added here is the prerequisite for that.

Fixes #111


nitnelave:
You might have noticed the teeny-tiny warning at the top? About breaking every single password? 

As much as updating dependencies is nice, I would very much like to avoid breaking users' flow. What's the motivation for upgrading opaque?

There is a document about a progressive migration of the passwords as the users log in (they re-register their password with the new protocol), that's something I'd be willing to merge. Note that it doesn't drop the old dependency yet, it'd take a while (and I'd probably put that off until 1.0, once I have more time to work on it and can merge the plugin API as well).

fredericrous:
Thanks for the feedback.
This PR is intentionally a WIP to explore and validate the upgrade to opaque-ke 4.0, not a proposal to merge a breaking change as-is.

My main driver is that I currently don’t have a usable TypeScript implementation of the older OPAQUE version, so I’d rather contribute here than build something that diverges. I only found the migration RFC after starting this work.

I’m not in a rush to merge, but I do want to move forward with the newer protocol on my side, so I’ll rework this toward a progressive migration approach instead of a hard break.

fredericrous:
@coderabbitai review

fredericrous:
@coderabbitai review

server/src/auth_service.rs
```diff
@@ -537,6 +583,134 @@ where
...
+        .map_err(|e| TcpError::BadRequest(format!("{e:#?}")))?
+        .into_inner();
+
+    // Decode base64 registration_start_request → RegistrationRequest
+    let raw_bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        .decode(&b64_request.registration_start_request)
+        .map_err(|e| TcpError::BadRequest(format!("Invalid base64: {e}")))?;
+    let registration_request =
+        opaque::server::registration::RegistrationRequest::deserialize(&raw_bytes)
+            .map_err(|e| TcpError::BadRequest(format!("Invalid registration request: {e}")))?;
```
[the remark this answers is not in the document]
fredericrous: . Two distinct endpoint families exist: registration_base64 (used by external clients like serenity-kit) 
    consistently uses URL_SAFE_NO_PAD on both encode (auth_service.rs:655, 685) and decode (623, 685). login_base64 legacy    
    endpoints in sql_opaque_handler.rs consistently use STANDARD on both sides. No endpoint mixes encodings. They're different
                                                                                                                              
    by design — URL_SAFE_NO_PAD is the JS-ecosystem default for OPAQUE wire format.

nitnelave:
I'm only going off of the PR description, I haven't read the code, but I have a couple of notes:
- what is Versioned envelope for in-flight legacy state? I'm not sure I understand the need. Isn't the endpoint enough to know which version of the protocol you're using?
- doesn't lldap_auth need to know the two opaque versions? In fact I was almost hoping we could limit them to only that crate and expose blind wrappers
- let's not use tokio:: spawn when re-registering. Keep everything simple, even if logging in is twice as long once per user. Same for the frontend.

fredericrous:
Thanks for the review. All three points addressed. lldap_auth::v07 now owns all v0.7 crypto and exposes opaque wrapper types and functions. Neither sql-backend-handler nor app import opaque-ke v0.7 directly anymore.
PR https://github.com/lldap/lldap/pull/1430 needs to be merged for all the new tests to pass

I tested manually this PR with container  https://github.com/fredericrous/lldap/pkgs/container/lldap/789746685?tag=opaque-ke-4 lgtm

fredericrous:
Heads-up: I force-pushed to tidy this up for review. Rebased onto latest main (no merge commits) and split into 8 logical conventional commits, readable top to bottom.

What it does: progressive, backward-compatible migration from opaque-ke 0.7 → 4.0 (RFC 9807). Existing passwords keep working and re-encode to v4.0 on next login. no reset required.

Worth noting:
- One operator step: first start needs --force-update-private-key=true once (both server_key file and key_seed setups). The startup error explains it and warns against --force-ldap-user-pass-reset.
- Security: the v0.7 login endpoint uses an indistinguishable dummy handshake so it doesn't leak account existence/version (no bypass). Residual: the v4.0 endpoint's 409 opaque_v07_version is an existence signal during the migration window but is needed for the client fallback. Happy to revisit.
- Tests: scripts/test_opaque_upgrade.sh builds the pre-migration baseline + HEAD and runs the real upgrade end-to-end for both file and seed key modes.

fmt/clippy clean, tests + e2e pass, codecov green. Happy to adjust to your conventions.

nitnelave:
Sorry, maybe I misunderstand, but won't force-update-private-key=true invalidate every password? If we were to do that we don't need to support both protocols at the same time... (In other words, we should not do that)

fredericrous:
You're right — using a flag documented as "invalidate every password" for a password-preserving migration was confusing. Even though the previous version didn't actually invalidate anything (the flag was only being reused to authorize the key-format rotation while keeping the old v0.7 key around for the fallback handshake), that's not how anyone would reasonably read it. Since the goal of the RFC is a zero-admin-action upgrade anyway, I removed the flag entirely.

The upgrade is now automatic: stop, deploy, start, done. On startup, the server:

* Detects whether the key file is an opaque-ke 0.7 key by actually parsing it. If the file is corrupted, it parses as neither version and startup aborts without touching anything.
* Verifies it's the same key used on the last successful startup by recomputing the pre-4.0 `private_key_hash` from the v0.7 key and comparing it with the value stored in the database. A valid but incorrect key (for example, restoring the wrong backup) is rejected.
* Only then performs the rotation. The old key is saved as a `<keyfile>.v07` sidecar before the key file is atomically replaced, so existing passwords continue to work until each user logs in and is transparently re-registered under the new key. For seed-based setups, the v0.7 key is derived from the same seed in memory, so nothing extra is written to disk. The rotation also only happens during normal server startup, so commands like `healthcheck` and `test-email` can never trigger it.

`--force-update-private-key` keeps its existing destructive behavior and isn't involved in the upgrade path.

I also force-pushed and cleaned up the history so the temporary flag approach is gone entirely. The PR description now includes a commit-by-commit guide, and the upgrade section has been updated to match the new flow.

The end-to-end test (`scripts/test_opaque_upgrade.sh`) now covers the zero-action upgrade path directly: start with a 0.7 binary, upgrade to HEAD with no flags, and verify the credentials continue to work and are automatically upgraded in both `server_key` and `key_seed` configurations.

While I was in there, I also implemented the optional admin-visibility item from the RFC: an admin-only `hasLegacyPassword` GraphQL field (server-side gated, `null` for non-admins), which powers a "Password upgrade pending" badge and remaining-user count in the user list. That gives admins a way to see when it's safe to eventually remove v0.7 support.

fredericrous:
@coderabbitai review

fredericrous:
Since your force-update note I don't want where this landed getting lost under the churn above: the automatic, flag-free upgrade I described earlier is the current state, and that's the part I'd value your read on.

Two follow-ups since, only one worth your eye: I restored the base64-in-JSON encoding you originally asked for in #111. opaque-ke 4.0's serde had quietly turned the /auth/opaque/* messages into raw integer arrays, changing the wire shape and breaking non-Rust clients; it's back to the pre-upgrade format. The rest was a routine CodeRabbit pass.

Does the automatic approach sit right with you, or is there anything you'd want changed before it's worth a deeper look?

nitnelave:
I think we'll need https://github.com/lldap/lldap/pull/1475 first to gracefully handle the server_key migration case where you have several instances starting at the same time with a (shared) server_key file.

Whoever gets to do the DB migration can do the server_key migration.

Actually, I have a better idea for the server key: re-use the existing server_key as a random initialization seed. It should be re-hashed just to be sure, but that would allow us to derive a new v4 key deterministically without having to mess around with new files.

app/src/components/login.rs
```diff
@@ -37,17 +38,68 @@ pub struct Props {
     pub password_reset_enabled: bool,
 }
 
+/// State carried into `Msg::AuthenticationStartResponse`. The username and
+/// password are kept around so we can fall back to the v0.7 login flow on
+/// HTTP 409 from the v4.0 endpoint.
+pub struct AuthStart {
+    pub state: opaque::client::login::ClientLogin,
+    pub username: String,
+    pub password: String,
```
nitnelave, line 47: Can we use a safer type for the password? I think there's secstr for that.

crates/sql-backend-handler/src/sql_opaque_handler.rs
```diff
@@ -38,59 +66,185 @@ fn passwords_match(
...
+    /// Upgrade a v0.7 password to the current format after successful
+    /// validation. The caller must have already verified the password
+    /// against the v0.7 `Validator` — this re-runs the full v4.0
+    /// registration flow and updates `password_version` in the DB.
+    async fn upgrade_password(&self, username: &UserId, password: &str) -> Result<()> {
+        info!(
+            r#"Upgrading password for "{}" from v0.7 to current format"#,
+            username
+        );
+        register_password(self, username.clone(), &SecUtf8::from(password)).await
```
nitnelave, line 190: legacy auto-upgrade can restore a reset password.
A v0.7 LDAP bind validates the old credential, then asynchronously performs an unconditional v4 registration. If an admin/user resets that password between validation and the write, the stale upgrade can overwrite the reset and restore access for someone holding the old password.
Fix: conditional/CAS update on the validated legacy password hash and  password_version = 0 ; a zero-row update means the password changed concurrently and must be left untouched.

nitnelave:
For the password reset while logged in with an old session, I think we can simplify: during the DB migration, invalidate all user sessions. Then it's impossible to get into that edge case.

nitnelave:
Note that it hits a caching problem: if the WASM stays cached (which we don't properly invalidate), the client will just fail to log in. I'll open an issue to track that.

nitnelave (changes requested):
Overall LGTM, if you address my few comments.

fredericrous:
Thanks for the review. All points are addressed and the branch is re-tested end to end (both key modes), plus rehearsed on a Postgres deployment.

- secstr (b484f92, 0f8dedf): round-trip password copies in the web client, plus BindRequest and the simple-login request, are now SecUtf8.
- CAS upgrade (ac9f59d): both the LDAP-bind and web re-registration write with WHERE password_hash = <validated> AND password_version = 0; a concurrent reset always wins. The web path gets a sealed upgradeToken from the v0.7 login finish for that. This is a per-row runtime race, unrelated to #1475.
- Server key (a49792c): your idea. The v0.7 key file is never rewritten; its bytes are hashed with a fixed label and used as the ChaCha20 seed like key_seed. No sidecar, no rotation. The seed uses the raw file bytes, not the parsed private key, since the hash of that key is already stored in the DB. A golden-vector test pins the derived bytes.
- Sessions (1447bc6): migrating across v12 blacklists all unexpired JWTs and deletes refresh tokens.

On #1475: with the key derived rather than written, the server-key migration no longer has a multi-instance file race, so it does not depend on #1475. The schema-migration race itself still belongs to #1475; the session invalidation above runs only in the process that applied the migration, so it works the same under #1475's create_schema gate.

Docs and the e2e script updated (581c8e7). The red Docker jobs are #1497, fixed in #1498.

Issue #1202, opened 2025-07-06
[BUG] Attributes with the same name can be created with different types


nitnelave:
When creating a user and group attribute, they can have different types. However, the LDAP spec requires each named attribute to have a single associated type.

**To Reproduce**
Create a "foo" attribute of type "string" for users.
Create a "foo" attribute of type "int" for groups.
Request the LDAP schema: the attribute will be listed twice with different names.

**Expected behavior**
Each attribute name should have a single associated type, across users and groups.

**Additional context**
https://ldap.com/attribute-types/

> An optional set of names that may be used to reference the attribute type as an alternative to the numeric OID. Each of these names must be unique across the set of all attribute types, although it is legal for an attribute type to have the same name as a different type of schema element (e.g., it is acceptable to have an attribute type and an object class that both have the same name).

**Proposed solution**
The LDAP schema generation should check for validity of attribute types across users and groups. In case of agreement, deduplicate. In case of conflict, return an error instead of generating the schema.

In addition, we should have a startup warning on conflicting attribute, and prevent users from creating them.

nitnelave:
@broeng @T0byV

PR #1426, opened 2026-04-03
fix: prevent attributes with conflicting types across users/groups

mvanhorn:
## Summary

Prevents creating user and group attributes with the same name but different types. LDAP requires each attribute name to have a single associated type (RFC 4512).

## Why this matters

Currently, creating a "foo" attribute as String for users and Integer for groups succeeds silently. When the LDAP schema is generated, the attribute appears twice with different types, which violates the LDAP spec and confuses clients.

## Changes

`crates/sql-backend-handler/src/sql_schema_backend_handler.rs`:

- `add_user_attribute`: Before inserting, queries `get_schema()` and checks if a group attribute with the same name has a different type. Returns `DomainError::InternalError` if types conflict.
- `add_group_attribute`: Same cross-check against existing user attributes.

When the types match, the attribute is created normally (deduplication in LDAP schema generation is a separate concern).

## Testing

- Creating "foo" as String for both users and groups: succeeds (same type)
- Creating "foo" as String for users, then Integer for groups: returns error
- Creating "foo" as Integer for groups, then Integer for users: succeeds (same type)
- Existing attributes without conflicts: no behavior change

Partial fix for #1202 (validation at creation time; LDAP schema deduplication and startup warnings are future work)

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/crates/sql-backend-handler/src/sql_schema_backend_handler.rs
+++ b/crates/sql-backend-handler/src/sql_schema_backend_handler.rs
@@ -29,6 +29,23 @@ impl ReadSchemaBackendHandler for SqlBackendHandler {
 #[async_trait]
 impl SchemaBackendHandler for SqlBackendHandler {
     async fn add_user_attribute(&self, request: CreateAttributeRequest) -> Result<()> {
+        // Check for conflicting group attribute with the same name but different type.
+        let schema = self.get_schema().await?;
+        if let Some(group_attr) = schema
+            .group_attributes
+            .attributes
+            .iter()
+            .find(|a| a.name == request.name)
+        {
+            if group_attr.attribute_type != request.attribute_type {
+                return Err(DomainError::InternalError(format!(
+                    "A group attribute '{}' already exists with type {:?}, \
+                     cannot create a user attribute with type {:?}. \
+                     LDAP requires each attribute name to have a single type.",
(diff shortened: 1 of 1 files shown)


mvanhorn:
Addressed in 1bdf200: wrapped both `add_user_attribute` and `add_group_attribute` in database transactions so the conflict check and insert are atomic. Uses the existing `get_schema_with_transaction` helper.

nitnelave (approved):

nitnelave:
Just a note: apparently on MySQL there are no cross-table locks, so even with a transaction you can still get TOCTOU. I think it's a remote enough possibility that we don't need to over-engineer it, but it's technically possible. It's also not really a security concern.

A proper design would involve a separate table for the schema and user/groups only holding foreign keys to the attributes, but that's more work than this is worth.

mvanhorn:
Really appreciate the merge, @nitnelave. Blocking same-name attributes with conflicting types across users and groups closes a sharp edge.

nitnelave:
Sorry about not merging earlier, I just had no time for LLDAP. It was a good PR, thanks!

mvanhorn:
Nice to have this in, thanks @nitnelave. Enforcing one type per attribute name lines up with what LDAP (RFC 4512) expects.

PR #1429, opened 2026-04-04
auth: Add trusted header authentication with IP allowlist

cqjjjzr:
This is a continuation of @Kumpelinus 's #1285 . I created a new PR to utilize the CI/auto-review on my PR. Let me know if it is favorable to close this and PR to @Kumpelinus 's repo.

1. Don't trust XFF headers, and use only the socket peer IP address
2. Decreased code duplication
3. Ensure the logout redirection actually work
4. Have the user specify CIDRs instead of single addresses, and validate the CIDR at startup
5. Prioritize trusted-header over JWT for GraphQL endpoints. ~The caveat is that some other endpoints  (OPAQUE-related) are still JWT-only.~ I also updated that.

### DIFF
--- a/crates/auth/src/lib.rs
+++ b/crates/auth/src/lib.rs
@@ -60,6 +60,23 @@ pub mod login {
         #[serde(rename = "refreshToken", skip_serializing_if = "Option::is_none")]
         pub refresh_token: Option<String>,
     }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    pub struct ServerTrustedHeaderResponse {
+        #[serde(rename = "userId")]
+        pub user_id: String,
+        #[serde(rename = "isAdmin")]
+        pub is_admin: bool,
+        #[serde(rename = "logoutUrl", skip_serializing_if = "Option::is_none")]
+        pub logout_url: Option<String>,
+    }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    #[serde(untagged)]
+    pub enum ServerAuthResponse {
+        Token(ServerLoginResponse),
+        TrustedHeader(ServerTrustedHeaderResponse),
+    }
 }
 
 /// The messages for the 3-step OPAQUE registration process.
--- a/server/src/auth_service.rs
+++ b/server/src/auth_service.rs
@@ -3,7 +3,7 @@ use crate::{
     tcp_server::{AppState, TcpError, TcpResult, error_to_http_response},
 };
 use actix_web::{
-    HttpRequest, HttpResponse,
+    FromRequest, HttpRequest, HttpResponse,
     cookie::{Cookie, SameSite},
     dev::{Service, ServiceRequest, ServiceResponse, Transform},
     error::{ErrorBadRequest, ErrorUnauthorized},
@@ -534,11 +534,9 @@ where
 {
     use actix_web::FromRequest;
     let inner_payload = &mut payload.into_inner();
-    let validation_result = BearerAuth::from_request(&request, inner_payload)
+    let validation_result = get_validation_results(&data, &request, inner_payload)
         .await
-        .ok()
-        .and_then(|bearer| check_if_token_is_valid(&data, bearer.token()).ok())
-        .ok_or_else(|| {
+        .map_err(|_| {
             TcpError::UnauthorizedError("Not authorized to change the user's password".to_string())
         })?;
     let registration_start_request =
@@ -660,6 +658,25 @@ where
         };
 
         Box::pin(self.service.call(req))
+    }
+}
+
+#[instrument(skip_all, level = "debug", err, ret)]
+pub(crate) async fn get_validation_results<Backend: BackendHandler>(
+    data: &web::Data<AppState<Backend>>,
+    request: &HttpRequest,
+    payload: &mut actix_http::Payload,
+) -> Result<ValidationResults, actix_web::Error> {
+    if data.trusted_header_options.enabled
+        && request
+            .headers()
+            .contains_key(&data.trusted_header_options.header_name)
+    {
+        check_if_trusted_header_is_valid(data, request).await
+    } else {
+        BearerAuth::from_request(request, payload)
+            .await
+            .map(|bearer| check_if_token_is_valid(data.get_ref(), bearer.token()))?
     }
 }
 
--- a/server/src/auth_service.rs
+++ b/server/src/auth_service.rs
@@ -105,8 +105,25 @@ async fn get_refresh<Backend>(
 where
     Backend: TcpBackendHandler + BackendHandler + 'static,
 {
+    if data.trusted_header_options.enabled
+        && request
+            .headers()
+            .contains_key(&data.trusted_header_options.header_name)
+    {
+        return get_trusted_header_refresh_response(&data, &request).await;
+    }
+    try_refresh_token(&data, &request).await
+}
+
+async fn try_refresh_token<Backend>(
+    data: &web::Data<AppState<Backend>>,
+    request: &HttpRequest,
+) -> TcpResult<HttpResponse>
+where
+    Backend: TcpBackendHandler + BackendHandler + 'static,
+{
     let jwt_key = &data.jwt_key;
-    let (refresh_token_hash, user) = get_refresh_token(request)?;
+    let (refresh_token_hash, user) = get_refresh_token(request.clone())?;
     let found = data
         .get_tcp_handler()
         .check_token(refresh_token_hash, &user)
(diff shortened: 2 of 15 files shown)


nitnelave:
@Kumpelinus WDYT? Are you planning to iterate on your PR, or should @cqjjjzr pick it up, with a co-author mention in the git message?

Kumpelinus:
> @Kumpelinus WDYT? Are you planning to iterate on your PR, or should @cqjjjzr pick it up, with a co-author mention in the git message?

Since I am quite a bit busy still and @cqjjjzr already looked into the code and thus I am happy with them taking over 👍

cqjjjzr:
@nitnelave Can we proceed with this? The latest AI review comment isn't valid IMHO because password change requires the `opaque/register/start` path to be available even when trusted header auth is enabled.

nitnelave:
`I/O error: no space left on maintainer`

Life is crazy busy, I'm having trouble finding time for LLDAP. Give me a few months.

cqjjjzr:
Totally understood - Real World (tm) affairs come first! Thank you for your work on LLDAP. Let me know if follow up changes/fixes are required for this. We have this patch deployed in our prod, so I expect major issue to be spotted quickly by us...

nitnelave:
Sorry, I'm reviewing anyway, but could you rebase?

EDIT: I force-pushed a rebase, I _think_ it's correct, but please take a look.

EDIT 2: I managed to botch the rebase, dropped some commits, and then started to complain about missing implementation in the review :D I re-did the rebase, it should be better now.

server/src/auth_service.rs
```diff
@@ -623,6 +716,29 @@ pub(crate) fn check_if_token_is_valid<Backend: BackendHandler>(
...
+) -> Result<ValidationResults, actix_web::Error> {
+    if !data.trusted_header_options.enabled {
+        return Err(ErrorUnauthorized(
+            "Trusted header authentication is disabled",
+        ));
+    }
+
+    let (user_id, groups) = validate_trusted_header(data, request)
+        .await
+        .map_err(|e| ErrorUnauthorized(e.to_string()))?;
```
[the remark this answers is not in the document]
nitnelave: I think this concern is genuine: we should try to return a 5xx when possible/relevant.

server/src/auth_service.rs
```diff
@@ -623,6 +710,55 @@ pub(crate) fn check_if_token_is_valid<Backend: BackendHandler>(
...
+#[instrument(skip_all, level = "debug", err, ret)]
+pub(crate) async fn check_if_trusted_header_is_valid<Backend: BackendHandler>(
+    data: &web::Data<AppState<Backend>>,
+    request: &HttpRequest,
+) -> Result<ValidationResults, actix_web::Error> {
+    if !data.trusted_header_options.enabled {
+        return Err(ErrorUnauthorized(
+            "Trusted header authentication is disabled",
+        ));
+    }
```
nitnelave, line 722: redundant check, remove.

server/src/configuration.rs
```diff
@@ -98,6 +99,25 @@ impl std::default::Default for HealthcheckOptions {
...
+#[builder(pattern = "owned")]
+pub struct TrustedHeaderOptions {
+    #[builder(default = "false")]
+    pub enabled: bool,
+    #[builder(default = r#"String::from("Remote-User")"#)]
+    pub header_name: String,
+    #[builder(default = "None")]
+    pub logout_url: Option<String>,
+    #[builder(default = r#"vec!["127.0.0.0/8".parse().unwrap(), "::1/128".parse().unwrap()]"#)]
+    pub trusted_proxies: Vec<IpNet>,
```
nitnelave, line 112: Should we allow for domains here? I'm thinking of a docker setup where the specific IP can change, but within docker (compose) the domain is stable.

It would make the implementation more complex, that's true, and you can set things up to have a stable-ish IP within docker, so it may not be worth it. Maybe we just need to update the docs a bit. Can you create a page for trusted header setup?

server/src/auth_service.rs
```diff
@@ -131,10 +148,84 @@ where
...
+        .ok_or_else(|| TcpError::UnauthorizedError("Could not determine client IP".to_string()))?;
+
+    if !data
+        .trusted_header_options
+        .trusted_proxies
+        .iter()
+        .any(|cidr| cidr.contains(&client_ip))
+    {
+        return Err(TcpError::UnauthorizedError(format!(
+            "Client IP {} not in trusted proxies",
```
nitnelave, line 179: Let's clarify this error. It should tell you how to fix it. Something like "When trusted headers are enabled, LLDAP should only be reachable from the proxy whose address is listed in trusted_proxies"

server/src/auth_service.rs
```diff
@@ -102,11 +102,28 @@ async fn get_refresh<Backend>(
     data: web::Data<AppState<Backend>>,
     request: HttpRequest,
 ) -> TcpResult<HttpResponse>
+where
+    Backend: TcpBackendHandler + BackendHandler + 'static,
+{
+    if data.trusted_header_options.enabled
+        && request
+            .headers()
+            .contains_key(&data.trusted_header_options.header_name)
```
nitnelave, line 111: We have to double-check in validate_trusted_header that the header is there. Instead, let's extract it here, check that against None, and if it's Some(..) pass it down the chain so we don't have to re-check it.

crates/auth/src/lib.rs
```diff
@@ -60,6 +60,23 @@ pub mod login {
...
+        pub is_admin: bool,
+        #[serde(rename = "logoutUrl", skip_serializing_if = "Option::is_none")]
+        pub logout_url: Option<String>,
+    }
+
+    #[derive(Serialize, Deserialize, Clone)]
+    #[serde(untagged)]
+    pub enum ServerAuthResponse {
+        Token(ServerLoginResponse),
+        TrustedHeader(ServerTrustedHeaderResponse),
```
nitnelave, line 78: We have to document that in the scripting docs and/or in the trusted header docs (the response shape). One can point to the other for dedupe.

nitnelave (changes requested):
I feel fairly confident about this implementation. One thing I'd like to see though is more hand-holding for the admin: better/clearer error messages and some docs.

Basically, place yourself in the shoes of an admin getting this error: what could have gone wrong? How do you fix it? Put as much context as possible in the error.

nitnelave:
Oh, and can we improve the test coverage as well? They slipped my mind, but there were a few more cases that I wanted to see covered in the tests (e.g. with both a trusted header and a token, valid or not)

cqjjjzr:
I'm back. Thank you for reviewing. I will look about the issues you mentioned soon as soon as I have time!

server/src/configuration.rs
cqjjjzr: I would rather not... Maintaining domain resolving (which has unbounded time) and caching may overcomplicate things. If true dynamic IP range is required, the user may use a local reverse proxy IMO.

cqjjjzr:
I made the code changes. doc changes are pending

server/src/configuration.rs
adepssimius: I would agree that allowing a domain here would be overkill for what trusted header auth is typically used for. I don't think I've seen any implementations that allow anything other than a range of IPs.

bendlas: If this were to be done, my first question would be: Would this be implemented in such a way that:

a) (one of the records for) the domain needs to resolve to the connecting IP?
b) the PTR record for the connecting IP needs to point back to the domain?
c) we're secure against attacks on DNS? are we implementing DNSSEC? requiring it?

Just the complexity inherent in this question would already dissuade me from attempting this as well. But if we wanted to do it: What use case do we have, where maintaining DNS entries for trusted proxies would be lower effort than maintaining trusted proxy config in lldap?

nitnelave: Yeah, I was asking to brainstorm, and on second thought (and with your inputs) it seems like we don't want that complexity. 

However, it would be nice to have in the docs some help on how to set up some common scenarios (docker, k8s?) so that people who don't know how to assign static IP (blocks) can get started.

adepssimius: Here's some documentation around how to safely deploy lldap with trusted proxy auth on k8s and docker. It recommends using a dedicated auth-net in docker so that only containers on that network are trusted as the RP, and a NetworkPolicy on k8s so that only the RP can reach the web UI port.

It also calls out the gotchas we'd otherwise get issues about:

- Docker DNS with multiple networks: if lldap sits on both the app network and auth-net, the plain lldap hostname resolves to both IPs, so the RP can randomly connect via the untrusted network and header auth silently falls back to the login page. The doc uses a network-scoped alias (lldap-auth) for the RP's upstream to force the connection onto the trusted network.
- Don't cut off the LDAP port: applying a NetworkPolicy default-denies everything else, so the example explicitly keeps 3890 open for applications; same reason the docker example keeps lldap on the shared network too. Connections from untrusted addresses aren't dangerous — they just get password login instead of header auth.
- k8s source IP: the RP should reach lldap via a ClusterIP Service (pod-to-pod preserves the pod source IP); a NodePort/LoadBalancer hop can SNAT it outside the trusted range, which fails closed but is confusing to debug.
- Header stripping: the RP must strip/overwrite the identity header on incoming requests.
It explains why trusting non-routable CIDRs is safe (the check is against the socket peer address, which can't be spoofed across the internet — so the threat model is purely internal, hence keeping the trusted range small).

I don't have a ready-made stack in docker to test this against, but I did test that NetworkPolicy in my cluster against the latest LLDAP build and everything stays functional, lldap interface access via RP and LDAP from anywhere in the cluster. Looking forward to disabling the double auth.

---

# Trusted header authentication

Trusted header authentication lets a reverse proxy that has already authenticated the user (Authelia, authentik, oauth2-proxy, ...) log that user into the LLDAP web UI by adding a header (default `Remote-User`) with the username. LLDAP trusts the header **only** if the connection comes from an IP listed in `trusted_proxies`. Username/password login remains available as a fallback.

## Configuration

```toml
[trusted_header_options]
enabled = true
#header_name = "Remote-User"
## Optional logout redirect (usually your SSO provider's logout endpoint).
#logout_url = "https://auth.example.com/logout"
## Networks allowed to send the trusted header, in CIDR notation
## (use /32 or /128 for a single IP). Default: localhost only.
trusted_proxies = ["172.28.0.0/24"]
```

Your proxy **must strip or overwrite the identity header on incoming requests**, otherwise anyone who can reach the proxy can impersonate any user by setting the header themselves.

Note that only the trusted header treatment depends on the source IP: LLDAP's other traffic is unaffected. Applications can still connect to the LDAP port from anywhere you allow, and a client reaching the web UI from outside `trusted_proxies` simply gets the normal password login instead of header auth.

## Why trusting private IP ranges is safe

LLDAP checks the **source IP of the TCP connection itself**, never `X-Forwarded-For` (which anyone can forge). A TCP handshake can't be completed from a spoofed source address, and private ranges like `10.0.0.0/8` or `192.168.0.0/16` aren't routable on the internet — so a connection sourced from a private address can only come from inside your network.

The risk is therefore internal: **anything inside the trusted range can log in as any user, including admin**, with a single header. Keep the range as small as practical — ideally a dedicated subnet shared only by the proxy and LLDAP. Don't trust all of `172.16.0.0/12` on a Docker host (every container) or a whole pod CIDR on Kubernetes (every pod) unless network policy restricts who can connect.

## Docker Compose

Individual container IPs change across restarts, so don't try to trust a single address. Instead, create a network with a fixed subnet for the proxy and LLDAP, and trust the whole subnet. A container can be attached to several networks at once, so your proxy can still serve everything else through your normal network — only its leg on `auth-net` matters here:

```yaml
networks:
  default: # the rest of your services live here
  auth-net:
    ipam:
      config:
        - subnet: 172.28.0.0/24

services:
  proxy:
    image: your-sso-proxy
    ports: ["443:443"]
    networks: [default, auth-net]

  lldap:
    image: lldap/lldap:stable
    networks:
      # on default too, so other services can reach the LDAP port (3890)
      default:
      auth-net:
        aliases: [lldap-auth]
    volumes:
      - ./lldap_config.toml:/data/lldap_config.toml

  someapp:
    image: some/app
    networks: [default]
```

Then `trusted_proxies = ["172.28.0.0/24"]`, and point the proxy's upstream for the web UI at `http://lldap-auth:17170`. The alias matters: since both containers are on both networks, the plain name `lldap` resolves to *both* of LLDAP's addresses, and if the proxy happens to connect over `default`, its source address falls outside the trusted range and header auth silently falls back to the login page. The `lldap-auth` alias only exists on `auth-net`, which forces the connection (and thus the proxy's source address) onto the trusted network.

Services on `default` can still do LDAP binds against LLDAP, and could even reach the web UI — from `default` addresses they only get password login. What matters is that nothing besides the proxy is attached to `auth-net`.

If instead everything shares one network and you trust its subnet, understand what that means: every container on it could impersonate any LLDAP user by sending one HTTP request with the header set. That may be acceptable for a small self-hosted stack where you trust all your own services, but the dedicated network above costs three lines and removes the question.

## Kubernetes

Pod IPs are ephemeral and your proxy is typically deployed by its own Helm chart, so you can't trust a single address. Instead, **trust the cluster's pod CIDR and pair it with a NetworkPolicy**: set `trusted_proxies` to the pod CIDR (e.g. `10.42.0.0/16` — check your CNI's configuration), and add a NetworkPolicy allowing ingress to LLDAP's port 17170 only from the proxy's pods:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: lldap-http-from-proxy-only
spec:
  podSelector:
    matchLabels:
      app: lldap
  policyTypes: [Ingress]
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: auth # proxy's namespace
          podSelector:
            matchLabels: {} # narrow to the proxy's pod labels if you can
      ports:
        - port: 17170
    # applying a policy denies everything not listed, so keep the
    # LDAP port open for the applications that authenticate against it
    - ports:
        - port: 3890
```

The CIDR check proves the connection came from inside the cluster; the policy ensures the only pods that can reach the web UI are the proxy's. This requires a CNI that enforces NetworkPolicies (most do; Flannel alone does not).

Have the proxy reach LLDAP via a ClusterIP Service, which preserves the pod source IP; a NodePort/LoadBalancer hop can SNAT it outside the trusted range (login is then refused — fails closed, but confusing).

PR #1436, opened 2026-04-10
fix: filter Kubernetes service-discovery env vars from config

fredericrous:
.


fredericrous:
issue in my deployment and wrong fix anyway. sorry, closing

PR #1443, opened 2026-04-28
fix(bootstrap): fix user_in_group() exit codes and improve error messages

XavierBeheydt:
## Description

The `user_in_group()` function used bare `return` statements when preconditions failed (group or user not found). A bare `return` defaults to exit code 0, so callers never received the expected failure signal and silently treated missing groups/users as a successful membership check, causing group assignments to be skipped.

## Changes

- `return` → `return 1` when the group does not exist
- `return` → `return 1` when the user does not exist
- Added `[WARNING]` prefix to the group-not-found message for consistency

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -192,12 +192,12 @@ user_in_group() {
 
   if ! group_exists "$group_name"; then
     printf '[WARNING] Group "%s" does not exist\n' "$group_name"
-    return
+    return 1
   fi
 
   if ! user_exists "$user_id"; then
-    printf 'User "%s" is not exists\n' "$user_id"
-    return
+    printf '[WARNING] User "%s" does not exist (cannot verify group membership)\n' "$user_id"
+    return 1
   fi
 
   if [[ "$(get_user_details "$user_id" | jq --raw-output --arg displayName "$group_name" '.data.user.groups | any(.[]; select(.displayName == $displayName))')" == 'true' ]]; then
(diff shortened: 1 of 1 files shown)


nitnelave:
I'm surprised we need the 1 second wait. Every read and write go to the same SQL DB, so there should be no propagation necessary. The only two edge cases I can think of is if the group creation and user creation are running in parallel (obviously not going to work since there's a dependency between the 2), or if you have a distributed DB with eventual consistency, and I don't know how you plugged that into LLDAP :D

In short, I don't think this is the right change.

XavierBeheydt:
Hey, you were right — the `sleep 1` wasn't needed and we removed it 😅

Looking back, the real bug was `user_in_group()` returning `0` (success) when the user didn't exist, which caused `add_user_to_group` to silently skip the assignment thinking the user was already in the group. The `sleep 1` masked this by introducing just enough timing variance to make tests feel like they worked sometimes.

Classic mistake: we reached for `sleep` out of habit from distributed systems (replica lag, LDAP replication...) without stopping to check that SQLite is synchronous and has no propagation delay. Lesson learned.

The fix is now properly in `user_in_group()` with explicit `return 1` on user-not-found, and the sleep is gone.

nitnelave:
I'm not against LLM written code, but at least write your own messages, please. I have enough conversations with LLMs at work, I don't need more on my spare time.

XavierBeheydt:
Sorry but my English is very poor. But I'm just want to send a fix with an issue with the bootstrap script.

nitnelave:
LGTM, but can you fix the PR title and commit description? It's not about a timing issue.

Issue #585, opened 2023-05-17
"Modify password" fails when the User ID contains a space


marlon0:
The description says it all. To recreate:
- Create a user with a User ID `user name`
- Use the admin interface to "modify password" for `user name`
- Get the error `Could not initiate password change`
- In the logs, see the error `🚨 [error]:  | error: Entity not found: user%20name`

steveb166:
just hit exactly this issue, any chance this can be fixed?

PR #1452, opened 2026-05-21
Fix: "Modify password" fails for users with spaces in their User ID

Xre0uS:
`yew-router 0.16` does not percent-decode URL path parameters. When a user ID contains a space (e.g. `user name`), the browser encodes the URL as `/user/user%20name/password`, and yew-router hands the component `user%20name`as-is, causing the request to fail.

Fix: `Percent-decode` the user_id route parameter in dispatch_route (using `percent_encoding::percent_decode_str`) before passing it to `UserDetails` and `ChangePasswordForm`. This ensures both components always receive the actual stored user ID regardless of URL encoding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2743,6 +2743,7 @@ dependencies = [
  "lldap_auth",
  "lldap_frontend_options",
  "lldap_validation",
+ "percent-encoding",
  "rand 0.8.6",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


nitnelave:
Fixes #585

nitnelave (approved):
Thanks!

nitnelave:
Seems like I need to update the MSRV to 1.91 for wasm-pack.

nitnelave:
Can you run cargo fmt on the code?

Xre0uS:
Done

PR #1456, opened 2026-06-08
fix(ldap): persist custom schema attributes when creating users via ldapadd

### LINKED ISSUE #1438: [BUG] Custom schema attributes are ignored when adding users with ldapadd
**Describe the bug**
In trying to populate my directory with dummy users, it appears that when importing users via ldapadd (I'm feeding it a LDIF file), custom attributes are being dropped in the user creation process. In the logs you can see the incoming ldap message has all the attributes, but they are dropped by the time it makes it to create_user

**To Reproduce**
1. create new attribute
2. create ldif file that populates that attribute
3. feed ldif file into lldap
4. notice that attribute is not populated

**Expected behavior**
attribute should be populated

**Logs**
```
lldap-redacted | 2026-04-18T01:30:50.828708511+00:00  DEBUG    ┝━ 🐛 [debug]:  | msg: LdapMsg { msgid: 2, op: AddRequest(LdapAddRequest { dn: "uid=award@redacted.net,ou=people,dc=redacted,dc=net", attributes: [LdapPartialAttribute { atype: "objectclass", vals: ["inetOrgPerson", "posixAccount", "mailAccount", "person"] }, LdapPartialAttribute { atype: "primaryemail", vals: ["award@redacted.net"] }, LdapPartialAttribute { atype: "entitlement", vals: ["Law"] }, LdapPartialAttribute { atype: "cn", vals: ["Adrian Ward"] }, LdapPartialAttribute { atype: "displayname", vals: ["Adrian Ward"] }, LdapPartialAttribute { atype: "first_name", vals: ["Adrian"] }, LdapPartialAttribute { atype: "givenname", vals: ["Adrian"] }, LdapPartialAttribute { atype: "sn", vals: ["Ward"] }, LdapPartialAttribute { atype: "mail", vals: ["award@redacted.net"] }, LdapPartialAttribute { atype: "uid", vals: ["award@redacted.net"] }] }), ctrl: [] }
lldap-redacted | 2026-04-18T01:30:50.828710165+00:00  DEBUG    ┝━ create_user_or_group [ 860µs | 0.92% / 91.30% ]
lldap-redacted | 2026-04-18T01:30:50.828711647+00:00  DEBUG    │  ┕━ create_user_or_group [ 852µs | 1.64% / 90.38% ]
lldap-redacted | 2026-04-18T01:30:50.828719949+00:00  DEBUG    │     ┕━ create_user [ 836µs | 26.86% / 88.74% ]
lldap-redacted | 2026-04-18T01:30:50.828749499+00:00  DEBUG    │        ┕━ create_user [ 583µs | 61.88% ] request: CreateUserRequest { user_id: "award@redacted.net", email: "award@redacted.net", display_name: Some("Adrian Ward"), attributes: [Attribute { name: AttributeName(CaseInsensitiveString("first_name")), value: String(Singleton("Adrian")) }, Attribute { name: AttributeName(CaseInsensitiveString("last_name")), value: String(Singleton("Ward")) }] } | user_id: "award@redacted.net"
lldap-redacted | 2026-04-18T01:30:50.832094905+00:00  DEBUG    ┕━ 🐛 [debug]:  | response: AddResponse(LdapResult { code: Success, matcheddn: "", message: "", referral: [] })
lldap-redacted | 2026-04-18T01:30:50.857214605+00:00  INFO     LDAP request [ 19.0µs | 100.00% ] session_id: 380d0cc0-46de-4d94-868d-9ac67b7c09d2
lldap-redacted | 2026-04-18T01:30:50.857231558+00:00  DEBUG    ┝━ 🐛 [debug]:  | msg: LdapMsg { msgid: 3, op: UnbindRequest, ctrl: [] }
lldap-redacted | 2026-04-18T01:30:50.857233088+00:00  DEBUG    ┕━ 🐛 [debug]: Unbind request for admin
lldap-redacted | 2026-04-18T01:30:50.857249849+00:00  INFO     ｉ [info]: LDAP session end: 380d0cc0-46de-4d94-868d-9ac67b7c09d2
```

**Additional context**
Sample LDIF file:
```
dn: uid=award@redacted.net,ou=people,dc=redacted,dc=net
changetype: add
objectclass: inetOrgPerson
objectclass: posixAccount
objectclass: mailAccount
objectclass: person
primaryemail: award@redacted.net
entitlement: Law
cn: Adrian Ward
displayname: Adrian Ward
first_name: Adrian
givenname: Adrian
sn: Ward
mail: award@redacted.net
uid: award@redacted.net
```

mvanhorn:
## Summary
Creating a user over LDAP (`ldapadd`) now persists custom schema attributes instead of silently dropping them.

## Why this matters
Custom attributes defined in the schema were ignored when a user was added via `ldapadd`: only built-in fields were written, so an `ldapadd` that set a custom attribute appeared to succeed but the value never persisted. This change parses incoming attributes, maps the ones that match a custom schema definition, type-checks them, and stores them with the new user.

## Changes
- Parse and decode incoming LDAP attribute values, separating built-in fields from custom schema attributes.
- Validate each custom attribute against its schema type; reject invalid typed values and ignore unknown attributes.
- Persist the validated custom attributes alongside the built-in user fields.

## Testing
Added tests covering a scalar custom attribute, a list-valued custom attribute, rejection of an invalid typed value, ignoring unknown attributes, and confirming built-in fields are not treated as custom.

Fixes #1438

AI was used for assistance.

### DIFF
--- a/crates/ldap/src/create.rs
+++ b/crates/ldap/src/create.rs
@@ -14,6 +14,7 @@ use lldap_domain::{
     requests::{CreateGroupRequest, CreateUserRequest},
     types::{Attribute, AttributeName, AttributeType, Email, GroupName, UserId},
 };
+use lldap_domain_handlers::handler::ReadSchemaBackendHandler;
 use std::collections::HashMap;
 use tracing::instrument;
 
@@ -44,28 +45,15 @@ async fn create_user(
     user_id: UserId,
     attributes: Vec<LdapAttribute>,
 ) -> LdapResult<Vec<LdapOp>> {
-    fn parse_attribute(mut attr: LdapPartialAttribute) -> LdapResult<(String, Vec<u8>)> {
-        if attr.vals.len() > 1 {
-            Err(LdapError {
-                code: LdapResultCode::ConstraintViolation,
-                message: format!("Expected a single value for attribute {}", attr.atype),
-            })
-        } else {
-            attr.atype.make_ascii_lowercase();
-            match attr.vals.pop() {
-                Some(val) => Ok((attr.atype, val)),
-                None => Err(LdapError {
-                    code: LdapResultCode::ConstraintViolation,
-                    message: format!("Missing value for attribute {}", attr.atype),
-                }),
-            }
-        }
+    fn parse_attribute(mut attr: LdapPartialAttribute) -> (String, Vec<Vec<u8>>) {
+        attr.atype.make_ascii_lowercase();
+        (attr.atype, attr.vals)
     }
-    let attributes: HashMap<String, Vec<u8>> = attributes
+    let attributes: HashMap<String, Vec<Vec<u8>>> = attributes
         .into_iter()
         .filter(|a| !a.atype.eq_ignore_ascii_case("objectclass"))
         .map(parse_attribute)
-        .collect::<LdapResult<_>>()?;
+        .collect();
     fn decode_attribute_value(val: &[u8]) -> LdapResult<String> {
         std::str::from_utf8(val)
             .map_err(|e| LdapError {
@@ -74,13 +62,43 @@ async fn create_user(
             })
             .map(str::to_owned)
     }
-    let get_attribute = |name| {
-        attributes
-            .get(name)
+    fn decode_attribute_values(vals: &[Vec<u8>]) -> LdapResult<Vec<String>> {
+        vals.iter()
             .map(Vec::as_slice)
             .map(decode_attribute_value)
+            .collect()
+    }
+    let get_attribute = |name| {
+        attributes.get(name).map(|vals| {
+            if vals.len() > 1 {
+                Err(LdapError {
+                    code: LdapResultCode::ConstraintViolation,
+                    message: format!("Expected a single value for attribute {name}"),
+                })
+            } else {
+                vals.first()
+                    .map(Vec::as_slice)
+                    .ok_or_else(|| LdapError {
+                        code: LdapResultCode::ConstraintViolation,
+                        message: format!("Missing value for attribute {name}"),
+                    })
+                    .and_then(decode_attribute_value)
+            }
+        })
     };
-    let make_encoded_attribute = |name: &str, typ: AttributeType, value: String| {
+    let make_encoded_attribute =
+        |name: &str, typ: AttributeType, is_list: bool, value: Vec<String>| {
+            Ok(Attribute {
+                name: AttributeName::from(name),
+                value: deserialize::deserialize_attribute_value(&value, typ, is_list).map_err(
+                    |e| LdapError {
+                        code: LdapResultCode::ConstraintViolation,
+                        message: format!("Invalid attribute value: {e}"),
+                    },
+                )?,
+            })
+        };
+    let make_single_encoded_attribute = |name: &str, typ: AttributeType, value: String| {
         Ok(Attribute {
             name: AttributeName::from(name),
             value: deserialize::deserialize_attribute_value(&[value], typ, false).map_err(|e| {
@@ -93,24 +111,55 @@ async fn create_user(
     };
     let mut new_user_attributes: Vec<Attribute> = Vec::new();
     if let Some(first_name) = get_attribute("givenname").transpose()? {
-        new_user_attributes.push(make_encoded_attribute(
+        new_user_attributes.push(make_single_encoded_attribute(
             "first_name",
             AttributeType::String,
             first_name,
         )?);
     }
     if let Some(last_name) = get_attribute("sn").transpose()? {
-        new_user_attributes.push(make_encoded_attribute(
+        new_user_attributes.push(make_single_encoded_attribute(
             "last_name",
             AttributeType::String,
             last_name,
         )?);
     }
     if let Some(avatar) = get_attribute("avatar").transpose()? {
-        new_user_attributes.push(make_encoded_attribute(
+        new_user_attributes.push(make_single_encoded_attribute(
             "avatar",
             AttributeType::JpegPhoto,
             avatar,
+        )?);
+    }
+    let schema = ReadSchemaBackendHandler::get_schema(backend_handler)
+        .await
+        .map_err(|e| LdapError {
+            code: LdapResultCode::OperationsError,
+            message: format!("Unable to get schema: {e:#}"),
+        })?;
+    let consumed_attribute_names = ["givenname", "sn", "avatar", "cn", "mail", "email"];
+    for attribute_schema in schema
+        .user_attributes
+        .attributes
+        .iter()
+        .filter(|attribute_schema| !attribute_schema.is_hardcoded)
+    {
+        let Some((_, values)) = attributes.iter().find(|(attribute_name, _)| {
+            !consumed_attribute_names
+                .iter()
+                .any(|consumed| attribute_name.eq_ignore_ascii_case(consumed))
+                && attribute_schema
+                    .name
+                    .as_str()
+                    .eq_ignore_ascii_case(attribute_name)
+        }) else {
+            continue;
+        };
+        new_user_attributes.push(make_encoded_attribute(
+            attribute_schema.name.as_str(),
+            attribute_schema.attribute_type,
+            attribute_schema.is_list,
+            decode_attribute_values(values)?,
         )?);
     }
     backend_handler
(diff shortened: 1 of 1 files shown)


crates/ldap/src/create.rs
```diff
@@ -74,13 +62,43 @@ async fn create_user(
...
+                name: AttributeName::from(name),
+                value: deserialize::deserialize_attribute_value(&value, typ, is_list).map_err(
+                    |e| LdapError {
+                        code: LdapResultCode::ConstraintViolation,
+                        message: format!("Invalid attribute value: {e}"),
+                    },
+                )?,
+            })
+        };
+    let make_single_encoded_attribute = |name: &str, typ: AttributeType, value: String| {
```
nitnelave: Can we express that in terms of `make_encoded_attribute` (and then even inline that in the callers)? I think the only thing blocking that is that `value` is a `Vec<String>` but it could be a `&[String]`

crates/ldap/src/create.rs
```diff
@@ -74,13 +62,43 @@ async fn create_user(
...
-            .get(name)
+    fn decode_attribute_values(vals: &[Vec<u8>]) -> LdapResult<Vec<String>> {
+        vals.iter()
             .map(Vec::as_slice)
             .map(decode_attribute_value)
+            .collect()
+    }
+    let get_attribute = |name| {
+        attributes.get(name).map(|vals| {
+            if vals.len() > 1 {
```
nitnelave: I would expect a correlation with the schema to see if it's a list attribute here, no? Or are you relying on the fact that every callsite for `get_attribute` is with a single-value hardcoded attribute?

crates/ldap/src/create.rs
```diff
@@ -93,26 +111,57 @@ async fn create_user(
...
             avatar,
         )?);
     }
+    let schema = ReadSchemaBackendHandler::get_schema(backend_handler)
+        .await
+        .map_err(|e| LdapError {
+            code: LdapResultCode::OperationsError,
+            message: format!("Unable to get schema: {e:#}"),
+        })?;
+    let consumed_attribute_names = ["givenname", "sn", "avatar", "cn", "mail", "email"];
```
nitnelave: I think we don't need to hardcode givenname, sn and so on. They are part of the schema and will be deduced the same way as the other attributes. This is just a generalization of the approach.

crates/ldap/src/create.rs
```diff
@@ -44,28 +45,15 @@ async fn create_user(
...
-            attr.atype.make_ascii_lowercase();
-            match attr.vals.pop() {
-                Some(val) => Ok((attr.atype, val)),
-                None => Err(LdapError {
-                    code: LdapResultCode::ConstraintViolation,
-                    message: format!("Missing value for attribute {}", attr.atype),
-                }),
-            }
-        }
+    fn parse_attribute(mut attr: LdapPartialAttribute) -> (String, Vec<Vec<u8>>) {
```
nitnelave:
```suggestion
    fn attribute_to_bytes(mut attr: LdapPartialAttribute) -> (String, Vec<Vec<u8>>) {
```

This doesn't really do any parsing.

crates/ldap/src/create.rs
```diff
@@ -44,28 +45,15 @@ async fn create_user(
...
+        attr.atype.make_ascii_lowercase();
+        (attr.atype, attr.vals)
     }
-    let attributes: HashMap<String, Vec<u8>> = attributes
+    let attributes: HashMap<String, Vec<Vec<u8>>> = attributes
```
nitnelave:
```suggestion
    let raw_ldap_attributes: HashMap<String, Vec<Vec<u8>>> = attributes
```

WDYT? This expresses that it's just the unparsed, unvalidated bytes.

crates/ldap/src/create.rs
```diff
@@ -44,28 +45,15 @@ async fn create_user(
...
         .into_iter()
         .filter(|a| !a.atype.eq_ignore_ascii_case("objectclass"))
         .map(parse_attribute)
-        .collect::<LdapResult<_>>()?;
+        .collect();
```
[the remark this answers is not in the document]
nitnelave: Please address this concern. I'm not sure what the LDAP RFC has to say about that.

crates/ldap/src/create.rs
```diff
@@ -93,26 +111,57 @@ async fn create_user(
...
         )?);
     }
+    let schema = ReadSchemaBackendHandler::get_schema(backend_handler)
+        .await
+        .map_err(|e| LdapError {
+            code: LdapResultCode::OperationsError,
+            message: format!("Unable to get schema: {e:#}"),
+        })?;
+    let consumed_attribute_names = ["givenname", "sn", "avatar", "cn", "mail", "email"];
+    for attribute_schema in schema
```
nitnelave: I would write the loop the other way: iterate over the attributes, and make sure you do something for each of them.
- They are present in the schema, add them (unless they have the wrong arity, in which case the whole request is bad)
- They are not in the schema, error with undefinedAttributeType.

crates/ldap/src/create.rs
```diff
@@ -196,6 +275,227 @@ mod tests {
...
+                &mock,
+                UserId::new("bob"),
+                vec![LdapPartialAttribute {
+                    atype: "unknown".to_owned(),
+                    vals: vec![b"ignored".to_vec(), b"also ignored".to_vec()],
+                }],
+            )
+            .await,
+            Ok(vec![make_add_response(
+                LdapResultCode::Success,
```
nitnelave: That should be an error: UndefinedAttributeType

crates/ldap/src/create.rs
```diff
@@ -196,6 +275,227 @@ mod tests {
...
+                String::new()
+            )])
+        );
+    }
+
+    #[tokio::test]
+    async fn test_create_user_builtin_attributes_are_not_custom_attributes() {
```
nitnelave: On the test name: There is no functional difference between custom attributes and the avatar/sn/givenname attributes. Only mail and uid/cn are special.

nitnelave (changes requested):
The approach is overall correct, I just have a few questions about the specific implementation. I think with another round or two we should easily get there.

mvanhorn:
Reworked per your review:

- The attribute loop now iterates over the provided attributes and classifies each via `map_user_field` against the schema, rather than hardcoding `givenname`/`sn`/`avatar` — those flow through the schema path like any other attribute now.
- Unknown attributes error with `UndefinedAttributeType`.
- List-vs-single arity is validated from the schema.
- Renamed `parse_attribute` → `attribute_to_bytes` and the map to `raw_ldap_attributes` per your suggestions, and pulled the encoding into a proper `make_encoded_attribute`.
- Added a test covering schema-driven custom + list attributes.

`cargo test` is green locally (79 passing, including the new test).

crates/ldap/src/create.rs
```diff
@@ -44,85 +139,74 @@ async fn create_user(
...
+    let raw_ldap_attributes: HashMap<String, Vec<Vec<u8>>> =
+        attributes.into_iter().map(attribute_to_bytes).collect();
+    let mut new_user_attributes: Vec<Attribute> = Vec::new();
+    let mut mail = None;
+    let mut email = None;
+    let mut display_name = None;
+    for (ldap_attribute_name, raw_values) in raw_ldap_attributes {
+        if ldap_attribute_name.eq_ignore_ascii_case("objectclass") {
+            continue;
         }
```
nitnelave, line 171: Handle this as part of the return from map_user_field

crates/ldap/src/create.rs
```diff
@@ -44,85 +139,74 @@ async fn create_user(
...
-                LdapError {
-                    code: LdapResultCode::ConstraintViolation,
-                    message: format!("Invalid attribute value: {e}"),
+        let attribute_name = AttributeName::from(ldap_attribute_name.as_str());
+        match map_user_field(&attribute_name, &schema) {
+            UserFieldType::PrimaryField(UserColumn::Email) => {
+                let value =
+                    single_primary_field_value(&schema, &ldap_attribute_name, "mail", &raw_values)?;
+                if ldap_attribute_name == "mail" {
+                    mail = Some(value);
```
nitnelave: Isn't that already handled by "map_user_field"? You're matching on UserColumn::Email, that should be enough. What is mail vs email?

If you're just trying to match the previous semantics, I think now is a good time to tighten them: if you provide both mail and email, error.

crates/ldap/src/create.rs
```diff
@@ -44,85 +139,74 @@ async fn create_user(
...
-        )?);
+            }
+            UserFieldType::PrimaryField(UserColumn::DisplayName) => {
+                let value = single_primary_field_value(
+                    &schema,
+                    &ldap_attribute_name,
+                    "display_name",
+                    &raw_values,
+                )?;
+                if ldap_attribute_name == "cn" || display_name.is_none() {
```
nitnelave: Same here, drop the check.

crates/ldap/src/create.rs
```diff
@@ -44,85 +139,74 @@ async fn create_user(
...
+                    display_name = Some(value);
+                }
+            }
+            UserFieldType::PrimaryField(UserColumn::UserId) => {
+                single_primary_field_value(&schema, &ldap_attribute_name, "user_id", &raw_values)?;
```
nitnelave: We should check that this matches the provided user_id (passed to the function, extracted from the DN)

nitnelave (approved):
LGTM! Some nits on redundant checks, but otherwise, it's ready.

Sorry for the slow review

mvanhorn:
Nits addressed, and I took your suggestion to tighten the semantics while at it: providing both mail and email is now a constraint violation instead of silent precedence, same for multiple display names, and duplicate attributes in the request get a ProtocolError. The user_id from attributes is captured and checked against the DN-derived one.

nitnelave:
Thanks! Merged. Could I ask you to do the same thing with groups, for consistency?

mvanhorn:
Will do. create_group is right there in the same file and currently drops request.attributes on the floor: it builds CreateGroupRequest from the DN-derived name alone, so anything the client sends is silently ignored rather than persisted or rejected.

Same shape as the user path then: decode and validate the attributes, treat conflicting spellings of the same field as a constraint violation instead of picking one silently, ProtocolError on duplicates in the request, and check any group name carried in the attributes against the DN-derived one. I will open it as a separate PR.

PR #1458, opened 2026-06-16
Support LDAP equality filters on list user attributes

rowlc:
# Support LDAP equality filters on list-valued custom user attributes

## Summary

This change allows LDAP equality filters to match values inside list-valued custom user attributes.

The motivating case is Stalwart Mail Server recipient lookup against LLDAP with a filter like:

```ldap
(&(objectClass=person)(|(mail=?)(mailalias=?)))
```

`mailalias` is commonly configured as a custom list attribute. Before this change, LLDAP rejected equality filters on list-valued custom user attributes with `UnwillingToPerform`, so an LDAP client could not resolve a user by any of their alias addresses. That prevented Stalwart from treating alias recipients as local users when aliases were stored only in LLDAP.

After this change, an equality filter such as:

```ldap
(mailalias=alias@example.com)
```

matches a user when `alias@example.com` is one of the values in that user's `mailalias` list.

## What changed

- Added a new internal user request filter variant, `AttributeListContains`, to represent equality matching against a list-valued custom attribute.
- Updated LDAP user filter parsing so equality filters on list-valued custom user attributes are accepted instead of rejected.
- Preserved the existing case-insensitive behavior by building an OR filter with the supplied value and its ASCII-lowercased form.
- Updated the SQL backend to support `AttributeListContains` by:
  - using a broad SQL prefilter when a list-membership check is present, and
  - applying the exact list-membership predicate in memory after users and custom attributes are loaded.
- Added regression coverage for:
  - LDAP equality filters on list-valued custom user attributes, and
  - SQL-backed user listing with `AttributeListContains`, including OR-composed filters.

## Why the SQL backend filters in memory

Custom list attributes are stored as serialized attribute values, and the existing SQL attribute equality path compares whole serialized values. For a list-valued attribute that would only match if the entire serialized list equals the query value, which is not the LDAP equality semantics clients expect.

To avoid changing the schema or adding database-specific JSON/list queries, this patch uses SQL only as a safe prefilter and then evaluates the precise list-membership condition against the loaded `UserAndGroups` records. This keeps the behavior correct for nested `AND`, `OR`, and `NOT` filters while keeping the change localized.

## Validation

- Built a Docker image from this branch and deployed it against a live LLDAP/Stalwart setup.
- Verified direct LDAP searches such as `(mailalias=alias@example.com)` return the expected user when the alias is one of several `mailalias` values.
- Verified Stalwart's mailbox lookup filter `(&(objectClass=person)(|(mail=?)(mailalias=?)))` can resolve aliases that are not the first `mailalias` value.
- Added focused unit tests for the LDAP filter parser and SQL backend list-membership filtering.

Targeted local test commands:

```bash
cargo fmt --check --all
cargo test -p lldap_ldap test_equality_filter_on_list_user_attribute
cargo test -p lldap_sql_backend_handler test_list_users_attribute_list_contains_filter
```

## Notes

This change is intentionally limited to equality matching for list-valued custom user attributes. It does not alter custom attribute storage, LDAP schema generation, or substring matching behavior.

### DIFF
--- a/crates/domain-handlers/src/handler.rs
+++ b/crates/domain-handlers/src/handler.rs
@@ -79,6 +79,7 @@ pub enum UserRequestFilter {
     UserIdSubString(SubStringFilter),
     Equality(UserColumn, String),
     AttributeEquality(AttributeName, AttributeValue),
+    AttributeListContains(AttributeName, AttributeValue),
     SubString(UserColumn, SubStringFilter),
     // Check if a user belongs to a group identified by name.
     MemberOf(GroupName),
--- a/crates/ldap/src/core/user.rs
+++ b/crates/ldap/src/core/user.rs
@@ -183,23 +183,21 @@ fn get_user_attribute_equality_filter(
     is_list: bool,
     value: &str,
 ) -> LdapResult<UserRequestFilter> {
-    if is_list {
-        return Err(LdapError {
-            code: LdapResultCode::UnwillingToPerform,
-            message: format!(
-                "Equality filter on list attribute \"{}\" is not supported",
-                field
-            ),
-        });
-    }
     let value_lc = value.to_ascii_lowercase();
-    let serialized_value = deserialize_attribute_value(&[value.to_owned()], typ, false);
-    let serialized_value_lc = deserialize_attribute_value(&[value_lc.to_owned()], typ, false);
+    let serialized_value = deserialize_attribute_value(&[value.to_owned()], typ, is_list);
+    let serialized_value_lc = deserialize_attribute_value(&[value_lc.to_owned()], typ, is_list);
     match (serialized_value, serialized_value_lc) {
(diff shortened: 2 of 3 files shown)


nitnelave:
I really wish you'd have come to me before, I would have pointed you to a proper design and a temporary workaround in the meantime. Instead, you spent time (and tokens, if I'm not mistaken) creating a PR that I will not accept.

rowlc:
Well that's a disappointing response. However you fix it, I'll look forward to the fix. In the mean time I'll use my fork that works, because the suggested workarounds so far are kind of hard to scale or manage. Thanks though.

nitnelave:
Just to clarify why I'm not merging this: it doesn't scale with the complexity of boolean expressions, you have to reinplement the entire tree evaluation in code if you want to cover all the cases

PR #1461, opened 2026-06-22
lldap-ha-auth: show meaningful error if display name is not set

ibizaman:
Without this, having a user without a display name set try to login
will result in Home Assistant rejecting the login with no error.
LLDAP will correctly return a user,
the command_line auth provider will succeed,
but Home Assitant will still return a 401.

The fix is to set a non empty display name for the user.
But there's no indication for that anywhere.
The logs show nothing obvious.
This is the only log line indicating any error happened:

```text
hass[1444]: 2026-06-22 19:02:56.454 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [133917151886080] Error handling message: Unauthorized (unauthorized)  from 127.0.0.1 (Mozilla/5.0 (X11; Linux x86_...
```

With this change, the command_line auth provider will exit with an error and print:

```text
hass[2220]: Empty display name not supporter.
hass[1434]: 2026-06-22 19:11:30.438 ERROR (MainThread) [homeassistant.auth.providers.command_line] User 'alice' failed to authenticate, command exited with code 1
hass[1434]: 2026-06-22 19:11:30.446 WARNING (MainThread) [homeassistant.components.http.ban] Login attempt or request with invalid authentication from localhost (127.0.0.1). Requested URL: '/auth/login_flow/...
```

### DIFF
--- a/example_configs/lldap-ha-auth.sh
+++ b/example_configs/lldap-ha-auth.sh
@@ -46,6 +46,10 @@ if [[ -z "$3" ]]; then
 	exit 1
 fi
 
+if [[ "$4" = "-v" ]]; then
+	set -x
+fi
+
 RESPONSE=$(curl -f -s -X POST -m "$TIMEOUT" -H "Content-type: application/json" -d '{"username":"'"$username"'","password":"'"$password"'"}' "$SERVER_URL/auth/simple/login")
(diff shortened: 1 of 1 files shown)


nitnelave:
Lgtm apart from the bot comments. Can you ping me when they're addressed?

PR #1465, opened 2026-07-07
LLDAP Design Update for PR Request

mitchelljfranklin:
-I Swapped Bootstrap from the 2021 version (5.1) to the current version (5.3). Same was done for the icon library.
-I Removed two old dependencies still referenced: one a third-party dark theme called "nightshade" and the other Font Awesome 4 (2016), since they wouldn't be used anymore.
-Changed the code to use the browser's built-in dark mode support (data-bs-theme). The choice is remembered in localStorage, and the theme is applied before the page even paints. The toggle is a sun/moon icon button vs the switch.
-Created a full set of design tokens in CSS for colors, spacing, corner rounding, shadows, fonts — all defined once in one place and referenced by name for ease of on going maintenance.
-Both light and dark palettes are defined side by side in the same file, making them easy to tweak.
-Navigation bar changed to a sticky header with the brand name and logo icon, centered the nav links, dropdown user menu with avatar.
-Login and password-reset screens centered card layout as well with a heading, input fields, and a spinner instead of an animated GIF.
-User and group list pages with a proper page title with the action button next to it (Create user / Create group), tables wrapped in a card with hover highlighting.
-Footer styled with the same border/colors as the rest of the app of course keeping the links for GitHub, Discord, and Twitter.

image


mitchelljfranklin:
I will wrap my head around the automated feedback and review for changes :)

mitchelljfranklin:
> I will wrap my head around the automated feedback and review for changes :)

Ok I think I have it and leaned some new things  on the way :), just validating on my end at the moment

nitnelave:
Sorry, maybe I missed something, are you still iterating on this PR or is it ready?

mitchelljfranklin:
> Sorry, maybe I missed something, are you still iterating on this PR or is it ready?

Oh thats my bad, I missed updating after I pushed - this should be ready to go :)

app/index_local.html
```diff
@@ -5,38 +5,52 @@
...
+        try { localStorage.setItem("theme", theme); } catch (e) {}
+      }
+    </script>
     <script src="/static/main.js" type="module" defer></script>
     <link
-      href="/static/bootstrap-nightshade.min.css"
-      rel="preload stylesheet"
-      integrity="sha384-CvItGYrXmque42UjYhp+bjRR8tgQz78Nlwk42gYsNzBc6y0DuXNtdUaRzr1cl2uK"
-      as="style" />
+      href="/static/bootstrap.min.css"
```
nitnelave, line 28: There should be a .txt file somewhere where we specify which files to pre-download and which version/checksum for the offline container, can you update this as well?

app/src/components/admin_settings.rs
nitnelave, line 1: I'm not sure I want all this complexity, especially in the front-end. Is customizability really a necessary feature? If we allow customizing (and that's a big if), it would be _at most_ through server-side config, and even better through a CSS file that the admin can put in the right place next to the binary (or a custom path)

nitnelave (changes requested):

app/src/components/login.rs
```diff
@@ -152,19 +152,20 @@ impl Component for LoginForm {
...
+                <div class="spinner-border text-primary" role="status">
+                  <span class="visually-hidden">{"Loading"}</span>
+                </div>
               </div>
             }
         } else {
             html! {
-              <form class="form center-block col-sm-4 col-offset-4">
+              <form class="auth-card">
+                <h1 class="form-title h4">{"Sign in to LLDAP"}</h1>
```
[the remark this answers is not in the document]
mitchelljfranklin: removed as part of code change, will update PR once happy after i complete testing

app/index_local.html
mitchelljfranklin: libraries.txt is updated

app/src/components/admin_settings.rs
mitchelljfranklin: Thats fair enough, I am more of a GUI side for ease of use however totally understand the view as well :).

I have stripped this PR down to just the refresh of bootstrap, icons etc. No config, admin or other just the visual refresh

mitchelljfranklin:
Updated PR, all additional admin settings, branding etc. has been removed. The PR now only has the refresh and style alignment changes

app/src/components/app.rs
```diff
@@ -200,26 +196,28 @@ impl App {
...
                         {"Create a user"}
                       </Link>
-                    }
-                };
-                html! {
-                  <div>
-                    { user_button("top-create-user") }
+                    </div>
                     <UserTable />
-                    { user_button("bottom-create-user") }
```
nitnelave, line 222: Having a button at the top _and_ at the bottom was an intentional design decision, for long lists of users. Let's not undo that.

Same for groups.

app/src/components/banner.rs
```diff
@@ -15,44 +15,49 @@ pub struct Props {
 
 #[function_component(Banner)]
 pub fn banner(props: &Props) -> Html {
+    let logo_html = html! {
+        <span class="app-brand-logo"><i class="bi-shield-lock-fill"></i></span>
```
nitnelave: I don't want to put a logo until we decide on a logo.

Having the block ready is nice, but I don't want it visible.

app/src/components/banner.rs
```diff
@@ -114,19 +122,49 @@ fn user_menu(props: &UserMenuProps) -> Html {
 
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(js_namespace = darkmode)]
-    fn toggleDarkMode(doSave: bool);
+    /// Flips the `data-bs-theme` attribute on `<html>` and persists the choice.
+    /// Defined as a global function in `index.html` / `index_local.html`.
+    fn toggleDarkMode();
 
-    #[wasm_bindgen]
```
nitnelave, line 120: We don't need to access them from js anymore? Cool.

Or rather, we don't need the annotation anymore? I see we still have it on the extern C

app/src/components/banner.rs
```diff
@@ -114,19 +122,49 @@ fn user_menu(props: &UserMenuProps) -> Html {
...
+    /// Reads whether the document is currently in dark mode.
     fn inDarkMode() -> bool;
 }
 
+/// Bootstrap Icons class shown when dark mode is active (offers to switch to light).
+const LIGHT_MODE_ICON: &str = "bi-sun-fill";
+/// Bootstrap Icons class shown when light mode is active (offers to switch to dark).
+const DARK_MODE_ICON: &str = "bi-moon-stars-fill";
+const SWITCH_TO_LIGHT_LABEL: &str = "Switch to light mode";
+const SWITCH_TO_DARK_LABEL: &str = "Switch to dark mode";
```
nitnelave: We can inline these constants.

app/static/style.css
```diff
@@ -1,32 +1,318 @@
...
-  font-family: 'Bebas Neue', cursive;
+/* ==========================================================================
+   LLDAP design system
+   All visual values are defined once as design tokens (CSS custom properties)
+   and referenced by name below. Avoid hard-coded colors / sizes in rules;
+   add or reuse a token instead.
+   ========================================================================== */
+
+:root {
+  /* Brand accent. Overridable at runtime (see setThemeAccent in index.html). */
```
nitnelave: _Not_ overridable?

app/static/style.css
```diff
@@ -1,32 +1,318 @@
...
+
+/* Uploaded or URL-sourced logo image shown next to the brand name.
+   Constrained to match the height of the shield icon so oversized logos
+   never break the header layout. */
+.app-brand-logo-image {
+  height: 2.25rem;
+  width: auto;
+  max-width: 10rem;
+  object-fit: contain;
+  border-radius: var(--lldap-radius-sm);
```
nitnelave: That seems overkill if the logo is not customizable

nitnelave (changes requested):
Looks great! A couple of minor comments, but LGTM otherwise

nitnelave:
Oh, and we'll need to update the readme screenshot with the new look once it's finalized

app/src/components/app.rs
mitchelljfranklin: Ah makes total sense, will re-apply code and add as part of commit back with the other changes

app/src/components/banner.rs
mitchelljfranklin: Sorry I missed this from the cleanup of the configuration, this has been cleaned up

app/static/style.css
mitchelljfranklin: Sorry I missed this from the cleanup of the configuration, this comment will be updated

app/static/style.css
mitchelljfranklin: Sorry I missed this from the cleanup of the CSS, this has been cleaned up

app/src/components/banner.rs
mitchelljfranklin: Moved to inline

app/src/components/banner.rs
mitchelljfranklin: Good catch - the #[wasm_bindgen] on the extern "C" block is still needed. I removed it from the individual functions because they're no longer called from JS, but they are still called from Rust: inDarkMode() is called on every render of the <DarkModeToggle> component, and toggleDarkMode() is called in the click handler. Without #[wasm_bindgen] on the extern "C" block, Rust/WASM can't reach across to the JS functions defined in index.html. So the annotation stays on the block, just not on the individual function declarations.

mitchelljfranklin:
> Oh, and we'll need to update the readme screenshot with the new look once it's finalized

New Screen shot loaded, just firming up Code Coverage errors

mitchelljfranklin:
Ready for your review when timing permits

PR #1468, opened 2026-07-19
Fix: panic in password modify when a regular user targets another user

kanywst:
A regular (non-admin) user can panic the LDAP handler by sending a Password Modify extended request that targets another user's DN. The read-permission check at `crates/ldap/src/password.rs:108` uses `.expect()`, so it panics instead of denying the request whenever the caller isn't allowed to read the target. The LDAP Modify path returns `InsufficientAccessRights` for the same case.

Fix: return `InsufficientAccessRights` instead, matching `modify.rs`, and add a regression test.


nitnelave (approved):
Thanks!

PR #1475, opened 2026-07-27
server: Require explicit schema migration for networked DBs by default

weberc2:
Hey @nitnelave, sorry for the slow turnaround on this, busy weekend. Here's the PR for the migration-safety change we talked through. Let me know if anything needs tweaking and I'll sort it out.

Implements the approach we settled on in #1467.

**Problem.** Migrations run on every startup as an unguarded check-then-apply with no lock. Two instances booting against the same networked database can both try to apply a pending migration — the loser hits duplicate DDL and crash-loops. Running several stateless instances against one Postgres/MySQL is a supported setup, so startup should be safe for it by default.

**Change.** Adds an `auto_migrate` option (env `LLDAP_AUTO_MIGRATE`, config `auto_migrate`). When a migration is pending on startup:

| `auto_migrate` | Behavior |
|---|---|
| unset (default) | SQLite migrates (single-writer, no race); PostgreSQL/MySQL refuse and point at `create_schema` |
| `true` | always migrate on startup |
| `false` | always require the explicit step |

`create_schema` is unchanged and always migrates — it's the explicit entry point. A database already at the latest version is always allowed, so a running deployment is unaffected. Downgrades keep their existing error.

**Heads-up — behavior change on upgrade.** Because the gate fires on any pending migration for a networked DB, a fresh Postgres/MySQL deployment and any single-instance upgrade across a schema bump now need `create_schema` run once (or `LLDAP_AUTO_MIGRATE=true`). This is the safe reading — two instances racing the initial, untransacted `upgrade_to_v1` is the worst case — but it's worth a line in the release notes. Happy to scope it to upgrades-only, or flip the networked default, if you'd rather.

**Tests.** Unit tests for `check_migration_allowed`: an up-to-date DB is allowed even with auto-migration off; a pending migration is refused with `false` and allowed with `true`/unset.

Developed with AI assistance; I've read and understood every line and am happy to walk through or adjust anything.

### DIFF
--- a/crates/sql-backend-handler/src/sql_tables.rs
+++ b/crates/sql-backend-handler/src/sql_tables.rs
@@ -54,6 +54,52 @@ pub async fn init_table(pool: &DbConnection) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Check whether the schema may be migrated automatically on startup, returning
+/// an error with instructions if not.
+///
+/// Migrations run on every startup as an unguarded check-then-apply, so two
+/// instances booting against the same networked database can race and one of
+/// them crashes on the duplicate DDL. To keep multiple instances safe by
+/// default, a networked database (PostgreSQL/MySQL) whose schema is out of date
+/// must be migrated explicitly with the `create_schema` subcommand, unless
+/// auto-migration is turned on. SQLite is single-writer, so it has no race and
+/// migrates automatically.
+///
+/// `auto_migrate`:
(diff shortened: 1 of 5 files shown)


nitnelave (approved):
Thanks, looks great!

nitnelave:
Looks like the integration tests need to be updated to have the auto-migrate: true

That means your feature is working!

nitnelave:
oh, and sorry for the slow review!

